PR #430, opened 2021-07-27
Fix typo in _loadPresetData()

### LINKED ISSUE #429: User simvar files don't load, and instead a second copy of the default simvars are added to the dropdown
**To reproduce**

1. Create a `msfs2020_simvars_user.cip` file and add custom simvars to it
2. Run Mobiflight and create a new SimConnect output

**Result**

There are two copies of the default simvars in the UI.

**Expected result**

One copy of the default simvars and one copy of the user simvars.

**More details**

The bug is in this line of code:

https://github.com/Mobiflight/MobiFlight-Connector/blob/5c9cb7b54ae79bc02ef016148d38022dd8332af2/UI/Panels/Config/SimConnectPanel.cs#L77

It should actually be this:

```
string[] lines = System.IO.File.ReadAllLines(file);
```

danecreekphotography:
Fixes #429 

Correct a typo in _loadPresetData() that caused duplicate copies of the default simvars to load instead of a copy of the user vars to load.

### DIFF
--- a/UI/Panels/Config/SimConnectPanel.cs
+++ b/UI/Panels/Config/SimConnectPanel.cs
@@ -74,7 +74,7 @@ namespace MobiFlight.UI.Panels.Config
 
         private void _loadPresetData(String file, String DefaultGroupKey, String Prefix)
         {
-            string[] lines = System.IO.File.ReadAllLines(PresetFile);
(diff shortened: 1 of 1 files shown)


Mobiflight (approved):
good job!

PR #477, opened 2021-09-06
Change check for whether multiple output pins are selected

### LINKED ISSUE #476: Output shift register settings don't always load properly
**Describe the bug**

**Steps To Reproduce**
1. Add an output shift register
2. Add an output config that uses it. For `Sim Variable` pick `MobiFlight Variable` and change the transform to `1`. In the `Display` settings check the `select multiple` option and then select multiple pins.
3. Save the settings
4. Edit the config you just created, don't switch to the the display tab, and hit ok
5. Edit the config you just created, go to the display tab

**Expected behavior**
The multiple output pins are selected

**Actual behavior**
The settings are reset to a single pin and Output 0 is selected

I realize the steps to reproduce seem a little involved but fundamentally what's happening is when you make a change on the Sim Variable tab it causes the output options for the shift register to get forgotten and reset back to a single pin.

danecreekphotography:
Fixes #476 

Replace the check for whether the panel was ever visible with a check for whether the checkbox is selected.

### DIFF
--- a/UI/Panels/Output/DisplayPinPanel.cs
+++ b/UI/Panels/Output/DisplayPinPanel.cs
@@ -141,8 +141,8 @@ namespace MobiFlight.UI.Panels
         virtual internal OutputConfigItem syncToConfig(OutputConfigItem config)
         {
             config.Pin.DisplayPin = displayPortComboBox.Text + displayPinComboBox.Text;
-
-            if (MultiPinSelectPanel.Visible)
+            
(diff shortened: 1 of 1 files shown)


MobiFlight-Admin:
There must have been a reason to do it this way.,.. i just don't recall.

MobiFlight-Admin:
I reckon it has something to do with the Arcaze Boards not supporting this feature and having MobiFlight Boards supporting this feature.

PR #479, opened 2021-09-07
Enable run button correctly after closing the settings dialog

### LINKED ISSUE #465: Run button is activated after enabling/disabling logging mode
**Describe the bug**
When logging mode is enabled or disabled, the Run button is activated, but MF is already in Run mode and is running.  Pressing the Run button, actually Stops MF, and another press of Run is necessary to reestablish the Run mode.

**Steps To Reproduce**
1. MF should be running normally with the sim.
2. Go to Settings and check Enable logging mode.
2. Exit Settings
3. Note Run button is activated, also note MF is running normally with the sim.
4. Press Run button... note MF stops.
5. Press Run button again... note MF now runs normally
6. Same happens when disabling logging mode

**Expected behavior**
After enabling or disabling logging mode, if MF was running already, Run button should stay deactivated.

**Actual behavior**
Run button is activated even though MF is already running normally.

**Details**
 - MobiFlight Version: 9.0.0
 - FlightSim: MSFS2020
 - AddOn: Irrelevant

**Additional context**
None

danecreekphotography:
Fixes #465 

Add an extra check to the places where fsuipc cache is opened/closed so the run button gets disabled if execManager is started, otherwise fall back to the prior logic to determine if the button should be enabled.

### DIFF
--- a/UI/MainForm.cs
+++ b/UI/MainForm.cs
@@ -601,31 +601,59 @@ namespace MobiFlight.UI
             runTestToolStripButton.Enabled = execManager.ModulesConnected();
         }
 
+        /// <summary>
+        /// Returns true if the run button should be enabled based on various MobiFlight states.
+        /// </summary>
+        private bool RunIsAvailable()
+        {
+            // If execution is already running then disable the run button.
+            if (execManager.IsStarted())
+            {
+                return false;
+            }
+            // If test mode is running then disable the run button.
+            else if (execManager.TestModeIsStarted())
+            {
+                return false;
+            }
+            // If there are no modules connected then running doesn't make any sense so disable the run button.
+            else if (!execManager.ModulesConnected())
+            {
+                return false;
+            }
+            // Connected sim? Enable the run button.
+            else if (execManager.SimConnected())
+            {
+                return true;
+            }
+            // If offline mode is enabled then light up the button.
+            else if (execManager.OfflineMode)
+            {
+                return true;
+            }
+
+            // None of the checks to enable the button passed so default to turning it off.
+            return false;
+        }
 
         /// <summary>
         /// updates the UI with appropriate icon states
         /// </summary>
         void fsuipcCache_Closed(object sender, EventArgs e)
-        {            
+        {
             if (sender.GetType() == typeof(SimConnectCache))
             {
                 simConnectToolStripMenuItem.Image = Properties.Resources.warning;
-            } else if (sender.GetType() == typeof(Fsuipc2Cache)) { 
+            }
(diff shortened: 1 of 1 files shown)


MobiFlight-Admin:
I think we have to add the execManager.isStarted() check to 
`runToolStripButton.Enabled = (execManager.OfflineMode || execManager.SimConnected()) && execManager.ModulesConnected() && !execManager.TestModeIsStarted();`

like
`runToolStripButton.Enabled = (execManager.OfflineMode || execManager.SimConnected()) && execManager.ModulesConnected() && (!execManager.isStarted() && !execManager.TestModeIsStarted());`

danecreekphotography:
That could work too but that comparison statement is already really confusing. It took me multiple debugging sessions to figure out each of the different states. I figured going with an if statement would make it much clearer what the logic is.


________________________________

danecreekphotography:
Honestly I probably should have refactored that code into a method since it gets used in two places...

MobiFlight-Admin:
I would like to keep it as a one liner and create a isRunAvailable() method. That's a good idea to have it in one place.

danecreekphotography:
Refactored to a new `RunIsAvailable()` method. Once it's in a dedicated method the shorthand single line test can get replaced with individual if statements that make it really clear what the tests are and what the outcome is. Makes stepping through the code to debug/understand much clearer.

Tested this with various combinations of connected/disconnected boards, the simulator, test button, and offline mode.

Lemme know what you think!

MobiFlight-Admin:
I was thinking if the combination of conditions are really captured in this sequence, it is really three things that we are looking at 1) we are running or not, and
2) we are in Offline mode or the Sim is connected, and
3) we have modules connected

I feel that this grouping is missing now. And with the if else branches the order of tests is also important. If one takes a look at it, it seems as if the run button should light up when Offline Mode is enabled.

danecreekphotography:
It helps to see you write it out with three numbered points, thanks! Even though you write "AND" at the end of each number, items 1 and 3 are automatically disqualifying things for showing the button correct? If the sim is running then it's immediately known the button should be off. Same for if no modules are connected: it's immediately known the button should be off. And same for test mode: if it's active then we know right away the button should be off.

That's captured in the if statement by having them first in the list of tests: check the things we know will immediately cause the button to be false and kick out of the method. Then look at the last remaining test are ways the button might be on.

You are correct, the order of the if statements does matter. If they get moved around by someone the logic will break. Hmm.

danecreekphotography:
@MobiFlight-Admin Switched back to the single line test. Verified the different conditions appear to work correctly.

PR #480, opened 2021-09-07
Fix the tab order

### LINKED ISSUE #462: Tab order for mobiflight variable in the input config dialog is wrong
**Steps To Reproduce**
1. Create a new input config
2. On the input tab choose any input
3. Set the action type to `MobiFlight - Variable`
4. Click in the `Name` text box
5. Press the `tab` key on your keyboard

**Expected behavior**
Insertion point moves to the `Value` text box.

**Actual behavior**
Focus moves to the `Type` dropdown.

danecreekphotography:
Fixes #462 

Update the tab order for the three fields on the variable input panel.


MobiFlight-Admin (approved):
Thanks!!!!

Issue #496, opened 2021-09-23
Allow Custom Install Path within the Mobiflight Installer Tool !


pizman:
**Is your feature request related to a problem? Please describe.**
MF Installer just "unpack" the MF Files in the Folder where we start the installer.exe  ( Same way as a self extracting EXE in the past.)
Lots of USers ( mostly at first days in MF) not know this.  And also advanced users make this mistake from time to time.
I will say 99% of all   MF Users had istalled by Mistake within the C:/users/downlaod folder one ( or more)  time in there life !

**Describe the solution you'd like**

Simply include a PATH field into the installer.
It should have a "hardcoded Recomendation" like  c:/mobiflight
And it should allow to rework the path..... best way with a GUI ( like Windows Explorer) .... If not just with a Path Text Line.

**Additional context**
@kkr0kk  already build this for testing.
BUT 
We got a Windows Security Warning on first try.

Whatever...: Thsi should be a point on our ToDo List !

kkr0kk:
PR is on, i add a gui for the first install

MobiFlight-Admin:
I will work on this later today. Thanks @kkr0kk

pizman:
Refresh !
In case DocMoebius is not 100% happy with the current Idea...
Compromise Solution:
We not need a Custom User Path Option with a UI..... We basically just need a PROTECTION to avoid a install/unpack in the c:/users/myname/download folder.....

So maybe we could build a check like "is installer.exe currently in download folder?" if Yes not allow installation.... Give a Error Popup, and tell the user that he need to move the file.

If this is still to much as expected... Then a simple Popup before starting the install process like " Attnetion... You will install MF in the Folder where you currently start the Installer.... Do you really want to continue ??   YES/NO.

**Overall... We jsut need to avoid the user install it by fault in the Download Folder !**

DocMoebiuz:
superseded by https://github.com/MobiFlight/MobiFlight-Connector/issues/953

PR #498, opened 2021-09-23
Installer : Add Setup options (issue #496)

kkr0kk:
Add a setup page when installer is started for the first time, user can choose directory for installation.
Some primary prerequises are checked before install.


kkr0kk:
#496

neilenns:
Fixes #496

MobiFlight-Installer/MobiFlight-Installer/UI/InstallMainForm.cs
```diff
@@ -0,0 +1,156 @@
...
+        public InstallMainForm()
+        {
+            InitializeComponent();
+            SelectedPath.Text = Directory.GetCurrentDirectory();
+            AsyncCheckGithub();
+            BoxTargetDirectory.Checked = MobiFlightUpdaterModel.VerifyCurrentFolderRight();
+            IfReadyForTakeOFF();
+        }
+
+        private void button1_Click(object sender, EventArgs e)
```
neilenns, line 33: The buttons (both `button1` and `button2`) should be given real names so it's clear what their function is when reading the code, and the event handlers should use those names. This button looks like it's for selecting install location? something like `installLocationButton` would work for example, and the handler would be `installLocationButton_click()`.

MobiFlight-Installer/MobiFlight-Installer/UI/InstallMainForm.cs
```diff
@@ -0,0 +1,156 @@
...
+            {
+                SelectedPath.Text = folderBrowserDialog1.SelectedPath;
+                PathToInstall = SelectedPath.Text;
+                MobiFlightUpdaterModel.SetInstallPath(PathToInstall);
+            }
+            BoxTargetDirectory.Checked = MobiFlightUpdaterModel.VerifyCurrentFolderRight();
+            IfReadyForTakeOFF();
+        }
+
+        private void button2_Click(object sender, EventArgs e)
```
neilenns, line 46: Same comment as above, should have a descriptive name instead of `button2`.

MobiFlight-Installer/MobiFlight-Installer/UI/InstallMainForm.cs
```diff
@@ -0,0 +1,156 @@
...
+                SetupTitle.ForeColor = System.Drawing.Color.Red;
+            }
+        }
+
+        private void label1_Click(object sender, EventArgs e)
+        {
+
+        }
+
+        private void SetupTitle_Click(object sender, EventArgs e)
```
neilenns, line 151: This doesn't do anything and should be removed.

MobiFlight-Installer/MobiFlight-Installer/UI/InstallMainForm.cs
```diff
@@ -0,0 +1,156 @@
...
+                SetupTitle.ForeColor = System.Drawing.Color.White;
+            }
+            else
+            {
+                SetupTitle.Text = "Take OFF aborted, something wrong";
+                SetupTitle.ForeColor = System.Drawing.Color.Red;
+            }
+        }
+
+        private void label1_Click(object sender, EventArgs e)
```
neilenns, line 146: This was in the original installer, but should be removed anyway since it doesn't do anything.

MobiFlight-Installer/MobiFlight-Installer/model/MobiFlightUpdaterModel.cs
```diff
@@ -91,8 +98,8 @@ public static bool VerifyCurrentFolderRight()
                 Log.Instance.log("InstallerHaveAdministratorRight : False", LogSeverity.Debug);
             try
             {
-                Directory.CreateDirectory(Directory.GetCurrentDirectory() + "\\test");
-                Directory.Delete(Directory.GetCurrentDirectory() + "\\test", true);
+                Directory.CreateDirectory(InstallPath + "\\test");
```
neilenns, line 101: This should use `Path.Join()` instead of string concatenation.

MobiFlight-Installer/MobiFlight-Installer/model/MobiFlightUpdaterModel.cs
```diff
@@ -91,8 +98,8 @@ public static bool VerifyCurrentFolderRight()
...
+                Directory.Delete(InstallPath + "\\test", true);
```
neilenns, line 102: This should use `Path.Join()` instead of string concatenation.

MobiFlight-Installer/MobiFlight-Installer/model/MobiFlightUpdaterModel.cs
```diff
@@ -150,15 +157,18 @@ public static void GoExtractToDirectory(string zipPath, string destinationDirect
 
                     if (file.Name == "MobiFlight-Installer.exe")
                     {
-                        if (InstallerIsNewer(file)) // NewInstaller have greater version than current
+                        if ((!File.Exists(InstallPath + "//" + file.Name)) || InstallerIsNewer(file))  // NewInstaller have greater version than current or installer not exist
```
neilenns, line 160: `Path.Join()` again, and elsewhere in this file. I realize some of this is older code but since you are touching it might as well clean it up to the more robust way of joining path parts together.

neilenns (changes requested):

kkr0kk:
I change naming, but i don't modify path concat, your porposal is not compatible with .NET 4.8, and all concat has been do like that, but yes that can be improve. That will be a lot of change with next lot of testing, i will wait first that idea was approve.

neilenns:
@kkr0kk where can I go to read about `Path.Join()` and changes to it in 4.8? I'm obviously out of the loop!

kkr0kk:
https://docs.microsoft.com/en-us/dotnet/api/system.io.path.join?view=net-5.0

neilenns (approved):
Generally looks right to me, but haven't tried it locally. GitHub wants me to take an action on this so marking approved. Doesn't mean anything though, Sebastien still has to review it :)

DocMoebiuz:
I am close to wrapping MobiFlight with a Standard Installer like NSIS or InnoSetup.
Thank you so much for all the effort that went into this. Please know that I appreciate that a lot and feel bad, that I didnt merge this pull request.

PR #502, opened 2021-09-25
Fixed On Press-On Release conflict for button events

### LINKED ISSUE #497: "Retrigger all switches" should trigger in two phases: First OFF and then ON.
We are facing an undefined state with a three position toggle switch for A320 VOR1 / OFF / VOR2 selector for the A320 EFIS. The switch is wired as two buttons into Mobiflight, with "OFF" state being a shared "OFF" between the two, due to them using the same physical lever. This problem exists for all such switches with an "OFF" state in the middle shared byt both side positions.

image

The initial state of the switch when MobiFlight starts is undefined, which is a problem. Because if the switch is on either side, it is "OFF" for the other. We could sync this with the retrigger switches function, but that is actually also undefined, because the order of syncing is not clearly defined.

**If "Retrigger Switches" synced OFF states first, and then ON states**, the OFF state of the "wrong" side would only briefly set the state during "retrigger", as the ON state of the other side would later override that.

We tried to wire the switch explicitely, but such switches with ON-ON-ON states are rare and hard to find, and even then it would make the wiring on the pcb more complicated. We should be able to fix this in software just fune.

ssewell:
This fixes #497 

Physical switches, such as an On-Off-On rocker switches using On Press events for the on positions (to explicitly set the switch position in-sim), and paired On Release events (to implicitly set the switch position to center in-sim) now set the expected position when using the "Retrigger All Switches" feature of MobiFlight firmware. This is accomplished by firing all Release events first, followed by all Press events.

The following scenarios were tested by purposefully de-synchronizing the physical rocker switch with the in-sim representation, then executing the "Retrigger All Switches" feature to evaluate the result in all possible combinations. For the sake of this example, the switch was placed in a horizontal orientation to match the in-sim representation of the A320 EFIS panel

image

Test resutls:
| Initial Physical Switch State | In Sim Switch State (before retrigger) | In Sim Switch State (after retrigger) |
| --- | --- | --- |
| Left | Left | Left |
| Left | Center | Left |
| Left | Right| Left |
| Center | Left | Center |
| Center | Center | Center |
| Center | Right | Center |
| Right | Left | Right |
| Right | Center | Right |
| Right | Right | Right |


ssewell:
Closing this to submit a cleaner pull request without superfluous commit history.

PR #503, opened 2021-09-26
Fixed visual truncation issue in Config References combo box

### LINKED ISSUE #495: Config References dropdown truncates the full config name
**Describe the bug**
If a lengthy (approximately 20 characters) Config names are used, then attempting to select one within the ConfigWizard -> Sim Variable (tab) -> Config References, is difficult as you have to enumerate each option manually to see the full name.

**Steps To Reproduce**
 1. Enter a number of config items with long names
 2. Edit any one, and attempt to select another one in a Config References dropdown

**Expected behavior**
The dropdown is sufficiently wide to display longer names.

**Actual behavior**
The dropdown width is short and prevents easy selection:
image

**Details**
 - MobiFlight Version: 9.0.0

**Additional context**
Config names are often prefixed to ensure that they are grouped logically when sorting by name. This is also driven by the fact if you sort by any column you lose the original config order (as entered) and have no provision to return to the original sorting.

This could be addressed by increasing the width of the dropdown shown within the Configure References.

ssewell:
Fixes #495 

Increases the width of the Config References combo box to prevent visual truncation shown here
image

After the change:
image

### DIFF
--- a/UI/Panels/Config/ConfigRefPanelItem.cs
+++ b/UI/Panels/Config/ConfigRefPanelItem.cs
@@ -27,6 +27,7 @@ namespace MobiFlight.UI.Panels.Config
             configRefComboBox.DataSource = dv;
             configRefComboBox.ValueMember = "guid";
             configRefComboBox.DisplayMember = "description";
+            configRefComboBox.Width = 395;
         }
 
         public void SetPlaceholder (String placeholder)


MobiFlight-Admin (commented):
Thanks for helping with this.
There is another option for us to fix this:
If you make the anchor of the combobox left top and right, then the combobox will automatically grow to the available width of the parent container. And of course the other two controls (as and placeholder) would have to be anchored to the right, not left.

I think that is even better than statically setting it to a width, because if we decide to enlarge the width of the wizard in the future, it wont adapt.

tigert:
or make the dialog resizeable and remember its size?

MobiFlight-Admin:
OMG.... what happened to this... how could I not see this PR, or did we loose track after my last comment. I have to check if this got actually fixed.

ssewell:
Not sure how to proceed with this. Setting the ComboBox "configRefComboBox" anchor to "Top, Left, Right" and the other controls as indicated has no effect. I confirmed in an external app that the anchoring works as expected with a similar set of controls, but does nothing in this context. Possibly related to the FlowLayoutPanel in which UI elements reside.

MobiFlight-Admin:
i have fixed this last night

MobiFlight-Admin:
Thanks for helping on this and sorry that i fixed it now myself :( I hope you don't mind!

ssewell:
> i have fixed this last night, i had to remove the flowlayout and also i noticed that the combo boxes had the wrong mode and some other details. It is included in the latest beta

Great! I'll review the diff so I can understand the issue in more detail. Thanks!

PR #511, opened 2021-09-30
Installer add offline installation mode

kkr0kk:
Add offline mode installation, with MobiflightConnector last release zip has Embedded ressource. If no ressource are available the offline mode is disable and hide.


DocMoebiuz:
I am close to wrapping MobiFlight with a Standard Installer like NSIS or InnoSetup.
Thank you so much for all the effort that went into this. Please know that I appreciate that a lot and feel bad, that I didnt merge this pull request.

PR #520, opened 2021-10-08
Save transform edits regardless of which tab is active when OK is pressed

### LINKED ISSUE #516: Output config wizard forgets changes in Transform when OK is pressed from another tab
**Describe the bug**
When changing an input config transform formula, going to another tab to make other changes and press OK causes MF to forget the change made in the Transform field.

**Steps To Reproduce**
1. Open input config
2. Change transform formula
3. Press the Compare tab to check
4. Press OK
5. Reopen same input config

**Expected behavior**
Input config wizard should save the Transform changes when OK is pressed, even if done from another tab.

**Actual behavior**
Changes made to Transform are not saved when OK is pressed from another tab.

**Details**
 - MobiFlight Version: 9.0.0.4
 - FlightSim: irrelevant
 - AddOn: irrelevant

**Additional context**
Workaround is to return to the Sim Variable tab and press ok, then the change is saved.

neilenns:
Fixes #516 

Instead of testing against the visibility of the control to determine whether the transform string should be saved use the state of the transform checkbox.

### DIFF
--- a/UI/Panels/Config/FsuipcConfigPanel.cs
+++ b/UI/Panels/Config/FsuipcConfigPanel.cs
@@ -306,7 +306,7 @@ namespace MobiFlight.UI.Panels.Config
             }
 
             // TODO: refactor this conditional stuff.
-            if (fsuipcMultiplyTextBox.Visible)
+            if (config.Transform.Active)
                 config.Transform.Expression = fsuipcMultiplyTextBox.Text;
             if (SubStringFromTextBox.Text!="")
                 config.Transform.SubStrStart = Byte.Parse(SubStringFromTextBox.Text);


neilenns:
@MobiFlight-Admin How about just getting rid of the test entirely and always saving the field value when hitting ok?

MobiFlight-Admin:
> Somewhat related to #558 ?
No. It was a different cause.

MobiFlight-Admin:
@neilenns I don't see a reason why  

`config.Transform.Expression = fsuipcMultiplyTextBox.Text;` 

should not simply happen all the time. if a user edits the text, and then decides to not activate, we still want to keep it.

Issue #527, opened 2021-10-21
With HID support, Mobiflight should be able to run without an Arduino connected


tigert:
Currently there is a requirement to detect at least one Arduino running MobiFlight firmware to use the "Run" functionality, as there was nothing MobiFlight could do without any connected buttons or leds etc.

With HID joystick support this is no longer the case, as you can do cockpit logic and configure inputs using joystick axis and buttons only.

Thus the check should be altered to check that there is at least one input or output device available I guess.

epaga:
The core check seems to be in RunIsAvailable ... removing ModulesConnected seems to be feasible at this point since with HID joystick support alone, so much is already possible without a connected Arduino!

```

        /// <summary>
        /// Returns true if the run button should be enabled based on various MobiFlight states.
        /// </summary>
        private bool RunIsAvailable()
        {
            return (execManager.OfflineMode || execManager.SimConnected()) && execManager.ModulesConnected() && !execManager.IsStarted() && !execManager.TestModeIsStarted();
        }

```

(similar spots in timer_Stopped and testModeTimer_Stop)

PR #528, opened 2021-10-21
Remove ModulesConnected condition: HID input can work without Arduino

epaga:
See issue #527 

Not an expert in this project, but this change seems to work for me for simple usage at least, built it and ran it without an Arduino device connected without any problems.


epaga:
awesome, looks great and makes sense. thanks @MobiFlight-Admin

PR #563, opened 2021-11-14
Make SimConnect the default for new output configs

### LINKED ISSUE #562: Make SimConnect (MSFS2020) the default output variable type
The vast majority of people are flying MSFS2020 now, but when they try and make an output it defaults to FSUIPC Offset which shouldn't be used with FS2020.

Make the default MSFS2020.

neilenns:
Fixes #562

### DIFF
--- a/MobiFlight/OutputConfigItem.cs
+++ b/MobiFlight/OutputConfigItem.cs
@@ -45,7 +45,7 @@ namespace MobiFlight
 
         public OutputConfigItem()
         {
-            SourceType = SourceType.FSUIPC;
+            SourceType = SourceType.SIMCONNECT;
             FSUIPC = new FsuipcOffset();
             SimConnectValue = new SimConnectValue();
             MobiFlightVariable = new MobiFlightVariable();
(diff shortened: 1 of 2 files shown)


MobiFlight-Admin:
Hi Neil - thanks and yeah, that's an easy fix :D Let us also change the order in the Config Wizard. So that the FSUIPC Offset goes last.

Additional info: When I implemented the support for the MSFS SimVars, I refactored a lot of code and when I had to decide whether FSUIPC should be default or MSFS2020 could be default, I was not 100% if I had used the type consistently in (very) old configs.

I have checked now again in the git history and I believe it should definitely be a backward compatible change.

neilenns:
@MobiFlight-Admin Updated the position of the radio buttons. WPF changes cause a lot of churn, good grief.

image

MobiFlight-Admin:
@neilenns - thanks mate! Sorry to be annoying... let's move FSUIPC all the way over when we are at it :)

MobiFlight-Admin:
I am thinking of really disabling the FSUIPC support as a global flag (someone has brought it up, i think it was Jaime). It would make the interface then specific to MSFS2020 and would not show FSUIPC at all. But that would be beyond this task for sure.

For now, let's move FSUIPC to the right.

neilenns:
@MobiFlight-Admin Done!

image

MobiFlight-Admin:
Woohoo!

PR #580, opened 2021-11-21
Add support for dynamically loading boards from XML

neilenns:
1. Creates a new `BoardManager.cs` and `Board.cs` file to handle all board options dynamically.
2. Use those board definitions to detect board type.
3. Use the detected board type to handle conditional logic throughout the code.

A couple of the files are best viewed in "split" rather than "unified" view, I will add comments where that is suggested.

Note that unit tests fail, I have to go in and update the unit tests still.


MobiFlight/Board.cs
```diff
@@ -0,0 +1,130 @@
...
+        public int MaxAnalogInputs = 0;
+
+        /// <summary>
+        /// True if multiple attempts should be made when connecting to the board.
+        /// </summary>
+        public Boolean ExtraConnectionRetry;
+        /// <summary>
+        /// True if the board supports loading firmware via MobiFlight.
+        /// </summary>
+        public Boolean CanInstallFirmware;
```
neilenns, line 75: This property is new and gets used elsewhere to skip firmware checks if `false`. Makes it really easy to have a board definition file for non-Arduino boards where the board should be detected and used by MobiFlight but flashing isn't supported.

MobiFlight/BoardDefinitions.cs
```diff
@@ -0,0 +1,74 @@
+﻿using System;
```
neilenns, line 1: This is a simple class that loads the board definitions from disk and then provides methods for looking up matching board definitions by different properties.

MobiFlight/MobiFlightCache.cs
```diff
@@ -111,10 +104,10 @@ public IEnumerable<MobiFlightModule> GetModules()
             return Modules.Values;
         }
 
-        private static List<Tuple<string, string>> getArduinoPorts()
+        private static Dictionary<string, Board> getSupportedPorts()
```
neilenns, line 107: These changes are best viewed in split mode.

MobiFlight/MobiFlightCache.cs
```diff
@@ -128,91 +121,47 @@ public IEnumerable<MobiFlightModule> GetModules()
...
                     foreach (String regSubDevice in regUSB.OpenSubKey(regDevice).GetSubKeyNames())
                     {
+                        Board board;
                         String FriendlyName = regUSB.OpenSubKey(regDevice).OpenSubKey(regSubDevice).GetValue("FriendlyName") as String;
                         if (FriendlyName == null) continue;
 
-                        // determine type based on names
-                        if (FriendlyName.Contains("Mega 2560"))
+                        // Check to see if any board configurations exist
+                        board = BoardDefinitions.GetBoardByFriendlyName(FriendlyName) ?? BoardDefinitions.GetBoardByHardwareId(regDevice);
```
neilenns, line 136: All of the previous checks by vid/pid and friendly name condense down to this single line. Cool :)

MobiFlight/MobiFlightCache.cs
```diff
@@ -128,91 +121,47 @@ public IEnumerable<MobiFlightModule> GetModules()
...
+                        result.Add(portName, board);
+                        Log.Instance.log($"Found potentially compatible module ({FriendlyName}): {regDevice}@{portName}", LogSeverity.Debug);
                     }
-
-                    if (message != null)
-                        Log.Instance.log(message, LogSeverity.Debug);
                 }
-
-                result = result.Distinct().ToList();
-                result.Sort((Tuple<string, string> item1, Tuple<string, string> item2) =>
```
neilenns, line 211: One thing I could never figure out: is there a reason why this list was getting sorted?

MobiFlight/MobiFlightCache.cs
```diff
@@ -228,15 +177,15 @@ private async  Task<List<MobiFlightModuleInfo>> LookupAllConnectedArduinoModules
             _lookingUpModules = true;
             
             List<Task<MobiFlightModuleInfo>> tasks = new List<Task<MobiFlightModuleInfo>>();
-            List<Tuple<string, string>> arduinoPorts = getArduinoPorts();
+            var supportedPorts = getSupportedPorts();
```
neilenns, line 180: The changes to this function are best viewed in split mode.

MobiFlight/MobiFlightCache.cs
```diff
@@ -253,23 +202,14 @@ private async  Task<List<MobiFlightModuleInfo>> LookupAllConnectedArduinoModules
 
                 tasks.Add(Task.Run(() =>
                 {
-                    MobiFlightModule tmp = new MobiFlightModule(new MobiFlightModuleConfig() { ComPort = portName, Type = ArduinoType });
+                    MobiFlightModule tmp = new MobiFlightModule(portName, board);
                     ModuleConnecting?.Invoke(this, "Scanning Arduinos", progressValue);
                     tmp.Connect();
                     MobiFlightModuleInfo devInfo = tmp.GetInfo() as MobiFlightModuleInfo;
```
neilenns, line 208: This is where the detected board will get assigned to the module. If there's no MobiFlight firmware installed it will fall back to the board detected in the USB scan step. If there is a MobiFlight firmware installed it will get set to the one that matches by the board type reported by the firmware.

MobiFlight/MobiFlightCache.cs
```diff
@@ -303,10 +243,7 @@ public async Task<bool> connectAsync(bool force=false)
             {
                 if (!devInfo.HasMfFirmware()) continue;
 
-                MobiFlightModule m = new MobiFlightModule(new MobiFlightModuleConfig() {
-                    ComPort = devInfo.Port,
-                    Type = devInfo.Type
-                });
+                MobiFlightModule m = new MobiFlightModule(devInfo.Port, devInfo.Board);
```
neilenns, line 246: And this is the magic :) Assigning the board to the module here means that it's available on the module throughout the rest of the MobiFlight codebase.

UI/MainForm.cs
```diff
@@ -94,6 +94,9 @@ public MainForm()
             // finally set up logging (based on settings)
             InitializeLogging();
 
+            // Initialize the board configurations
+            BoardDefinitions.Load();
```
neilenns, line 98: Board definitions get loaded immediately after logging is set up and before any other work is done.

neilenns (commented):
Added comments to help understand some of the changes.

MobiFlight/Board.cs
MobiFlight-Admin: Cool idea.

neilenns:
Ok @MobiFlight-Admin this is updated based on your feedback.

1. XML is dead, long live JSON. Not only are the board definition files in JSON now I also made a JSON schema so when editing the board definition files in Visual Studio Code you get full IntelliSense support.
2. Restructured all the properties into groups for clarity.
3. Made the unit tests pass again.

MobiFlight-Admin (approved):
great work! you are so fast  but still with great coding standards!

PR #588, opened 2021-11-25
Support for MinimumFirmwareVersion

neilenns:
* Added `MinimumFirmwareVersion` to the board definition files
* Changed the version properties on the `Info` class to be true `Version` objects
* Add a JSON serializer to convert the string in the board file to real `Version` objects
* Clean up a few places in the code that can now use the version properties directly
* Add additional logic to `BoardDefinitions.GetBoardByMobiFlightType` that takes into account the minimum firmware version
* Add some additional validation to the JSON schema to enforce version format

### DIFF
--- a/Boards/arduino_mega.board.json
+++ b/Boards/arduino_mega.board.json
@@ -34,6 +34,7 @@
     "DelayAfterFirmwareUpdate": 0,
     "LatestFirmwareVersion": "1.12.1",
     "FriendlyName": "Arduino Mega 2560",
+    "MinimumFirmwareVersion": "0.0.1",
     "MobiFlightType": "MobiFlight Mega"
   },
   "ModuleLimits": {
--- a/Boards/arduino_micro.board.json
+++ b/Boards/arduino_micro.board.json
@@ -24,6 +24,7 @@
     "CanInstallFirmware": true,
     "FriendlyName": "Arduino Micro Pro",
     "LatestFirmwareVersion": "1.12.1",
+    "MinimumFirmwareVersion":  "0.0.1",
     "MobiFlightType": "MobiFlight Micro"
   },
   "ModuleLimits": {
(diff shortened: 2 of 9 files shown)


neilenns:
I need to think about this a bit more tomorrow and see if `CurrentFirmwareVersion` can do the same thing without an additional new property.

MobiFlight-Admin:
Hi @neilenns - oh wow, that was quick! i think that's a clever approach. Did you think of any naming convention for the files yet, in case there are 2 for, e.g., the Mega?

The version number of the firmware would make sense.
arduino_mega.1.12.1.board.json

MobiFlight-Admin (approved):

neilenns:
Yeah, I was thinking just `arduino_mega_1_12_1.board.json` etc.

I still haven't figured out what to do about `CurrentFirmwareVersion` though. You have to update it in every file every time for this to all work which is fine when there's only one file for each board, but will become a real drag when there's more than one file for each board.

neilenns:
@MobiFlight-Admin What do you think about a separate `firmware_versions.json` file that is just a mapping of `MobiFlightType` to `CurrentFirmwareVersion`?

That could also live in `BoardManager.cs` and would mean one place to update the firmware version. The board definition files could then have multiple copies for different minimum firmware versions but wouldn't need to repeat what the current firmware version is.

MobiFlight-Admin:
Do we actually need MinimumFwVersion? If we have multiple files and we can sort them by FwVersion, we should be able to derive which board definition actually fits the board.

I have a board. Version 1.0.0 and it supports 5 Buttons.
There is a board config for this board. MF.1.0.board.conf.
LatestFirmware is 1.0.0

Now we fix a bug, device definitions dont change but we increment the firmware. 1.0.1
We could update LatestFirmware 1.0.1 in the MF.1.0.board.conf

Then we make a change 1.1.0 - the device numbers change.
We would create a new MF.1.1.board with LatestFirmware 1.1

MobiFlight-Admin:
Maybe we should discuss this one more time interactively

neilenns:
We agreed this is not the right way to go :D

PR #595, opened 2021-11-28
Check for corrupt settings on launch

### LINKED ISSUE #589: Improve procedure to write settings file in Appdata
It has happened now twice that a PC crash results in a corrupt settings file in Appdata/Local/Mobiflight which causes Mobiflight to not start. I suspect the corruption is caused by mobiflight updating the file when the crash occurs. 

I suggest that the writing procedure could be improved by first writing the file to a temp file and then renaming the original file as .old and the .tmp as .cfg. This would prevent overwriting the original until the last minute and creates a backup that can be used if the corruption still occurs. 

Use the cfg from a previous version which can be months old. 

**Additional context**

neilenns:
Fixes #589 

It's apparently a known thing with the .NET settings framework that the user config file can get corrupted if the PC crashes while an app is running that uses settings.

Unfortunately, Flight Simulator 2020 likes to crash PCs so this isn't as rare for MobiFlight users as one would expect.

The fix is to manually test the settings file for validity before anything in the app references `Settings.Default` and if a corrupt file is detected delete it. This will cause the settings framework to use defaults.

### DIFF
--- a/Program.cs
+++ b/Program.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Windows.Forms;
 using System.Threading;
 using MobiFlight.UI;
+using System.Configuration;
 
 namespace MobiFlight
 {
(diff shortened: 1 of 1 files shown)


neilenns:
I tested this locally by manually chopping the end off my user.settings file and it correctly handled the exception, nuked the user settings file, and replaced it with a default.

DocMoebiuz (approved):
Thanks! Interesting.

PR #599, opened 2021-12-02
Fix errors in isPWM and isI2C pin assignments for the mega board config

### LINKED ISSUE #189: VIDs/PIDs configure in external config-file (std-config included in installiation)
User can add new Boards by itself.
CH340 Chip UNOs can set as UNO and not as MEGA by the user.

[Issue created by drome1800: 2017-04-04]
[Last updated on bitbucket: 2020-09-16]

[Comment created by DocMoebiuz: 2020-09-16]
I guess we are working on it right now… REBOOT!

[Comment created by ryucz: 2020-09-16]
@{557058:1c646c6b-f7d7-4548-b231-938050d2203a}  This should not be MINOR, this should  be MAJOR. Major above all. In next buid. In next hour. I just bought Arduino MEGA `VID_0403&PID_6001` it is mentioned many times in forums, even added in **MobiFlightModuleInfo.cs** @ 113, but i have tried 7.9 and 7.10 and none of them is looking for my board.

neilenns:
Fixes #189 

* Make the `isI2C` and `isPWM` pin assignments correct based on what was previously assigned in `MobiFlightModuleInfo.cs` before the move to a separate board configuration file

I also reviewed the micro and uno board configs, they were fine.

### DIFF
--- a/Boards/arduino_mega.board.json
+++ b/Boards/arduino_mega.board.json
@@ -51,74 +51,74 @@
   "Pins": [
     {
       "isAnalog": false,
-      "isI2C": true,
-      "isPWM": false,
+      "isI2C": false,
+      "isPWM": true,
       "Pin": 2
     },
     {
       "isAnalog": false,
-      "isI2C": true,
-      "isPWM": false,
+      "isI2C": false,
+      "isPWM": true,
       "Pin": 3
     },
     {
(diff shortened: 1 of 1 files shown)


MobiFlight-Admin (approved):
So basically i2c and pwm was the other way around

PR #612, opened 2021-12-11
Guard against failed firmware updates

### LINKED ISSUE #611: Exception throws when GetInfo fails after attempting to to flash firmware
**Describe the bug**

If the firmware fails to flash on a bare Arduino the subsequent call to GetInfo fails, which later results in an exception getting thrown because the device serial number is null.

**Steps To Reproduce**
(This is a hack to reproduce the issue)

1. Move the vid/pid for a nano from the arduino_mega.board.json file to the arduino_uno.board.json
2. Plug in the nano, run MobiFlight, and attempt to flash it with the Uno firmware

**Expected behavior**
Flash works and MobiFlight runs

**Actual behavior**
Flash fails and MobiFlight throws an exception

This is totally an unsupported scenario but it does expose an error in the post-flash logic that should be corrected.

neilenns:
Fixes #611

* Move when `OnAfterFirmwareUpdate` gets called so it only happens on successful firmware updates
* Add a check for null values to `RegisterModule` to avoid the crash in other situations where the serial number might be null.

### DIFF
--- a/MobiFlight/MobiFlightCache.cs
+++ b/MobiFlight/MobiFlightCache.cs
@@ -285,6 +285,14 @@ namespace MobiFlight
         {
             Log.Instance.log("MobiFlightCache.RegisterModule("+m.Name+":"+ m.Port +")", LogSeverity.Debug);
             
+            // Additional protection added for edge cases where this gets called after a failed firmware update, which resulted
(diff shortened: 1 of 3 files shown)


DocMoebiuz (approved):
thanks

PR #613, opened 2021-12-13
Add Active to list of elements that are tested for Equals()

### LINKED ISSUE #607: Change in Output Config interpolation setting does not activate Save button
**Describe the bug**
I disabled the interpolation setting in the Compare tab of an output config and upon exit noticed the Save button remains inactive.  Had to use Save As command to save.  Enabling the interpolation setting also does not activate the Save button.

**Steps To Reproduce**
1. Go to any output config
2. Click on Compare tab
3. Enable interpolation
4. Press OK

**Expected behavior**
Config was changed so Save command should be enabled upon exit of the config wizard.

**Actual behavior**
Save button remains inactive.

**Details**
 - MobiFlight Version: 9.1.0.5
 - FlightSim: irrelevant
 - AddOn: irrelevant

**Additional context**
none

neilenns:
Fixes #607 

The `Equals()` method on `Interpolation` wasn't including `Active` in the list of values that must be the same for two interpolation objects to be equals. Adding the test is sufficient to make the Save button work correctly.

### DIFF
--- a/MobiFlightUnitTests/InterpolationTests.cs
+++ b/MobiFlightUnitTests/InterpolationTests.cs
@@ -52,22 +52,24 @@ namespace MobiFlightUnitTests
 
             Assert.IsTrue(o1.Equals(o2));
 
+            // Verify that Active is used in the equals comparison
+            o1.Active = true;
+            Assert.IsFalse(o1.Equals(o2));
+
+            // Verify that a list of interpolations is used in the equals comparison
             float x1 = 0.1f; float y1 = 0.1f;
             float x3 = 0.5f; float y3 = 2.0f;
             float x2 = 1.0f; float y2 = 1.0f;
-            o1.Active = true;
             o1.Add(x1, y1);
             o1.Add(x2, y2);
             o1.Add(x3, y3);
-
             Assert.IsFalse(o1.Equals(o2));
 
-            o2.Active = true;
             o2.Add(x1, y1);
             o2.Add(x2, y2);
             o2.Add(x3, y3);
+            o2.Active = true;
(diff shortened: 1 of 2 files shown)


MobiFlight-Admin (approved):
Thanks!!!

MobiFlight-Admin:
@neilenns - you wanna add a test case too? Then it is perfect!

neilenns:
@MobiFlight-Admin Considering this is the entirety of the current InterpolationPanelTest.cs file... no? 😂

```C#
namespace MobiFlightUnitTests.Panels.Group
{
    class InterpolationPanelTests
    {
    }
}
```

MobiFlight-Admin:
You should look into InterpolationTests.cs , it is in the root folder which is not the right place... i wonder why it ended up there.

neilenns:
That unit test is already setting Active. The flaw in the tests is it sets a bunch of stuff all at once to verify equals is false and since active wasn't being tested at all and the others were it still passed.

I updated the unit test to independently check Active on its own.

MobiFlightUnitTests/InterpolationTests.cs
```diff
@@ -28,7 +28,7 @@ public void CloneTest()
 
             o.Active = true;
             o.Add(x1, y1);
-            o.Add(x2, y2);
+            o.Add(x2, y2);
```
MobiFlight-Admin, line 31: why are there so many lines marked as changed... strange.

MobiFlight-Admin (approved):
Wonderful!

neilenns:
The line endings were wrong and I let VS fix them

PR #622, opened 2021-12-27
Remove "v" from get_firmware script

### LINKED ISSUE #621: Don't assume "v" for version numbers
**Describe the bug**
When pulling specific version builds from the firmware source repo don't assume it leads with a "v". Require the explicit version tag be provided on the command line.

This is the companion issue to https://github.com/MobiFlight/MobiFlight-FirmwareSource/issues/125 on the firmware source side of things.

neilenns:
Fixes #621

Instead of assuming all tags start with "v" in the firmware source repo require the script get called with the actual version tag.

Change the command line parameter from "version" to "tag" to accurately reflect what it is.

### DIFF
--- a/Build/get_firmware.ps1
+++ b/Build/get_firmware.ps1
@@ -11,7 +11,7 @@ if ($version -eq "latest") {
 }
 else {
   Write-Output "Downloading firmware version $($version)"
-  $releaseDetails = Invoke-RestMethod -Uri "https://api.github.com/repos/MobiFlight/MobiFlight-FirmwareSource/releases/tags/v${version}"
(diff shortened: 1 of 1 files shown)


neilenns:
This is the second part of cleaning up how "v" is handled with firmware versions.

MobiFlight-Admin:
@neilenns I haven't used this script yet.. How does it work?

neilenns:
Oh. I made it for you to make it easy to get the latest firmware from the GitHub repo. Just run the PowerShell script frkma command line and it will download the latest firmware and put it in the connector's repo firmware folder

DocMoebiuz:
Ok. Sounds good! I would still have to add it to the project-file so that it will be copied over to the output folder - or we add it to the post build script… gotta think about it

Issue #625, opened 2021-12-29
Adding devices under circumstances causes an exception


GioCC:
**Describe the bug**
Adding devices to a module sometime causes an exception depending on which UI items happen to be selected.

**Steps To Reproduce**

***Scenario 1 (OK):***
- Open `(Settings/)Mobiflight` module dialog.
- DO NOT expand the device tree on a module, or alternatively: EXPAND the tree and SELECT a child (device) entry
- Right click on the `Modules` entry and add a device

Result:
- everything works as expected 

***Scenario 2 (BUG):***
- Open `(Settings/)Mobiflight` module dialog.
- expand module tree; DO NOT select a child entry
- Right click on the `Modules` entry and add a device

Result:
- `System.NullReferenceException` in `MobiFlightPanel.cs`, line 366 (`getVirtualModuleFromTree()` returns null)

See submitted PR #626 for fix.

**Details**
 - MobiFlight Version: 9.2

pizman:
Nice i found this Issue Report.
Still make a Screenshot some Mins ago and would like to Report the same bahavour.

Do you guys still find a way to reproduce ?  Or better do you know WHEN the Bug is happen 100%

Yesterday i got it by testing ~~  4 of 10 Trys....  But i not know what have changed in the trys ?
If it work, then it work all time.... Only a Restart ( Reconnect of Mega) could happen it again.
BUT
Today....  50 Trys.... No Bug !

GioCC:
Unfortunately I discovered a further bug in this fix - will post issue and fix shortly.

PR #626, opened 2021-12-29
Fix for issue #625: Adding devices under circumstances causes an exception

GioCC:
Pre-select a child item (at least one is always created) to prevent null reference


MobiFlight-Admin:
Thanks for this pull request. Is there an issue for this and what does it fix it exactly?

DocMoebiuz (commented):
Going over https://docs.microsoft.com/en-us/dotnet/api/system.windows.forms.control.select?view=windowsdesktop-6.0#System_Windows_Forms_Control_Select and reading what it does, I think it is a side-effect that we trigger the node to be selected.

In my opinion it would be even better to make sure we have a already one node selected when we expand the tree.

With the current proposed approach, if i have two modules, and 
* i first open the dialog
* Now the first node / module is preselected
* Now i expand the second module (node)
* Now I add a device

Expected result:
* New device is added under second module (node)

Actual result:
* New device is added under first selected module (node)

DocMoebiuz:
In line 367 after we are doing 

```
cfgItem = null;
tempModule = getVirtualModuleFromTree();
```
we should definitely add a 

`if (tempModule == null) return;`

because the getVirtualModuleFromTree() will return "null" - so I should have checked in the first place.

GioCC:
By "Module" you mean the parent node (i.e. the Arduino board)? In this case, first of all my wording is imprecise, because the issue does not arise when adding a module, but a... how is it called? Peripheral?
Anyway, I had not considered what happend with more that one board, I will look further into it. Thanks for pointing it out.

GioCC:
I added the forced node selection when the user clicks (usually, left clicks) on a node and - probably redundant - when a node is expanded.
I also encountered a case of failed board programming, when the configuration could not be loaded causing a further exception upon trying to display it; i made a single change, a check against null, in `MobiFlight/MobiFlightModule.cs`.

DocMoebiuz (commented):
I have to check on the Config == null check, because it doesn’t look right to me. The Config property is not implemented good, and when Config is null it should fetch it from the Audi o board on the fly again.

the other changes are all OK.

GioCC:
[THIS](https://github.com/MobiFlight/MobiFlight-Connector/commit/76b0509e9ecac649694fc4bf2f653cd3e26f1692) commit has wrecked the PR (and then some). I will probably have to do everything from scratch otherwise there's risk of leaving bugs around. Not very eager about that.

GioCC:
So what you suggest to do about the Config == null check? It probably made its way in during testing, but it's not strictly part of this PR... I could remove it from here and see to opening a new issue about it.

moebigit:
if it was by accident then i would remove it.

GioCC:
Done!

GioCC:
I lost sight of this for a little, is there anything else to fix? To me it should be OK.

DocMoebiuz (approved):
Thanks!

DocMoebiuz:
Fixed #625

GioCC:
Unfortunately I discovered that this fix introduces a further bug - will post issue and fix shortly.

Issue #638, opened 2022-01-03
Number of digits is lacking the 7 option for Display Modules


ssewell:
**Describe the bug**
The "Number of digits" dropdown in the ConfigWizard for Display Modules is lacking the 7 option.

**Steps To Reproduce**
1. Open the ConfigWizard to the Display panel
2. Select a module that contains a Display Module
3. Select Display Module in the "Use type of" dropdown
4. Observe the values shown in the "Number of digits" dropdown
![image](https://user-images.githubusercontent.com/2242776/147901400-19bed629-376c-4fb5-84e8-662673359499.png)

**Expected behavior**
Number of digits should allow for the selection of "7"

**Actual behavior**
Number of digits dropdown contains 3-6, 8

**Details**
 - MobiFlight Version: 9.2

pizman:
Hi Guys.
Whatever it sounds crazy i ask this ( after working 5+ Years with MF since the early days)
WHAT is the sense of this Setting..... I never used it till now and never understand the usecase.
Why should i "reduce" the GUI  vissible Diggits ?   I normaly choose e.g. 4 Diggit sof the 8 Available by checkmark.
Second question.... If i set e.g. "4" in that dropdown... Do i see the most left or the most right 4 diggits then ?
Or better explaned....   
If i like to use 4 Diggits with ConfigA and the other 4 Diggits with ConfigB ....  If i select "4" in that dropdown.... Would both configs show me the same 4 diggits ??

In easy words....  Why is this dropdown needed basically?

DocMoebiuz:
This is only needed if you have less than 8 digits driven by one max chip, e.g. with a custom adapter. The Scan Limit is set to the correct number of digits and it will make sure that they show with the same brightness.

ssewell:
This was fixed in #639

PR #639, opened 2022-01-03
Added the missing 7 option from "Number of Digits" dropdown

ssewell:
The "Number of digits dropdown for LED display modules was missing the 7 option.


ssewell:
Fixes #638

DocMoebiuz (approved):
Thanks!

PR #646, opened 2022-01-06
Add Genuino UNO

fedoracoreuser:
The Genuino UNO is an official (but older) Arduino UNO. I have one, and I managed to get it working by adding the proper PID value.
I tested this by rebuilding mobiflight and connecting it to MSFS 2020. I added a dual button setup for HDG inc and dec and it worked fine.
The test suite is unaffected by the change.


DocMoebiuz (approved):
Thanks!

Issue #523, opened 2021-10-10
Multi-Select Option in Bitmask GUI


pizman:
**Is your feature request related to a problem? Please describe.**
Its a little uncomfortable to work with the "BitMask Feature" in case mostly we need to only choose for one Bit , BUT by Default all Bits are marked in a "new" Config.

![grafik](https://user-images.githubusercontent.com/26087116/136696591-1e08a50e-29bf-40f5-b1a1-763834f9449f.png)

**Describe the solution you'd like**
I know we can also simply write the Hexadecimal Value for the needed Bit..... But many users ( specialy beginners) use the GUI with the Checkmarks!
So a "Clickspot" would be cool to DE-SELECT all Checkmarks by one click......   Maybe if possible also with a second function to SELECT all pins by press it again.

PR #652, opened 2022-01-07
Fix for issue #523: Multi-Select Option in Bitmask GUI

GioCC:
Added "Set all" and "Clear All" buttons to BytePanel


DocMoebiuz:
Hey cool! I will review this on my PC tomorrow! Thanks for helping out!

GioCC:
I added the "all set/clear" buttons on byte level rather than for the whole vector, it looks a little crammed but I think it's the most useful way.

DocMoebiuz (approved):
Thanks! - i like the border around every byte. And i think it is ok to have two buttons per group.

Issue #665, opened 2022-01-19
"Apply" Button in Outputs Precondition Tab is hidden


zorzini:
**Describe the bug**
The button "Apply" is almost completely hidden in the "Precondition" tab; just less than a millimeter to click it.

**Steps To Reproduce**
1. Go to 'Output Setting'
2. Click on 'Precondition Tab' - 'Configure a new Item'

**Expected behavior**
Apply button to press.

**Actual behavior**
Butto 'Apply' is hidden

**Details**
 - MobiFlight Version: 9.2.0.2
 - FlightSim: [e.g. MSFS2020]
 - AddOn: [e.g. Default Aircraft]

**Additional context**
It seems that the Frame container for Precondition Settings in the Form is too higher.

![Precondition tab](https://user-images.githubusercontent.com/65813913/150192799-84127f51-da49-41aa-ad6b-0d9a75c69c86.jpg)

DocMoebiuz:
No on my screen. Weird.

zorzini:
Hi i saw the mail and tested on laptop and is the same: no Apply button

![1642628226743_image](https://user-images.githubusercontent.com/65813913/150218957-5f6bb8a9-236d-4dac-a9c9-667f061f179b.png)

zorzini:
Hi i saw the mail and tested on laptop and is the same: no Apply button

![1642628226743_image](https://user-images.githubusercontent.com/65813913/150219348-585dfb96-17a6-4ba9-b21d-191da9a23a8d.png)

DocMoebiuz:
no worries... i got it.

PR #667, opened 2022-01-19
665 fix precondition

DocMoebiuz:
Finished the refactoring that was started for the Output Configs. Now also the input config wizard uses the central precondition panel. Now we can think of improving some of the things there.


DocMoebiuz:
Fixes #665

Issue #676, opened 2022-01-31
Missing argument in SetModule<...> log line


GioCC:
**Describe the bug**
In the log line for SetModule command, the <points> argument is missing.

**Details**
 - MobiFlight Version: 9.2.0

PR #677, opened 2022-01-31
Issue #676 Missing argument in SetModule<...> log line

GioCC:
Fix for issue #676


DocMoebiuz (approved):
Thanks! Good catch!!!

PR #698, opened 2022-02-20
Improved WASM Module with Multi Client Support and Connection Detection

DocMoebiuz:
- [x] Update the WASM Module to the latest multi-client module. Now multiple clients can simultaneously use the module, see https://github.com/MobiFlight/MobiFlight-WASM-Module/pull/3
- [x] Added a ping/pong logic to detect from the Connector side whether WASM Module is responding.
- [x] Only if "pong" response comes back from WASM Module then MF Connector shows connection as active
- [x] Renamed WASM module to use a "MobiFlightWasmModule" name which is unique and will not cause conflicts with other WASM modules
- [x] Remove old StandaloneModule.wasm file in MobiFlight module folder, just to make sure that this is really not loaded.


neilenns:
Nice!!!!!!!

PR #707, opened 2022-02-23
Add support for copying from the log window

### LINKED ISSUE #702: Allow copying content from the new logging window
The new logging output window is super fast (yay!) but doesn't support copying. It would be really handy for debugging purposes if people could copy the text from the output window so it can be pasted in places like Discord chat. Right now the only option is to take a screenshot which is a pain if you want to get a VID/PID out of the logs.

neilenns:
Fixes #702 

Quick and easy change to support copying log messages from the new logging window to the clipboard.


DocMoebiuz:
Thanks

PR #734, opened 2022-03-11
Use WMI to detect devices instead of the system registry

### LINKED ISSUE #733: Wrong board type used if duplicated entries in registry
**Describe the bug**
My Arduino Pro Micro could not be flashed because MF thought it was a Mega. The issue is that I have two valid module entries in the registry and they both point to COM3 which was the port that my Pro Micro was connected to.

**Steps To Reproduce**
1. Connect a Mega
2. Connect a Pro Micro
3. Disconnect the Mega
4. Open Registry Editor
5. Search for the Mega entry in HKEY_LOCAL_SYSTEM/CurrentControlSet/Enum/USB/{CORRECT VID PID}/{INSTANCE ID}/Description
6. Change the Port value to the same Port value that your ProMicro is using
7. Flash ProMicro with EEPROM clear sketch
8. Start MF

**Expected behavior**
ProMicro is detected as ProMicro Module based on the VID/PID.

**Actual behavior**
ProMicro is detected as Mega Module because MF finds two entries for the Port and takes the first one it finds.

neilenns:
Fixes #733 

Initial attempt at using WMI to detect devices instead of using the system registry. This works on my desktop and detects all my connected Arduinos, as well as a few other non-Arduino devices that are correctly skipped.

A few things to note:

1. The "FriendlyName" is no longer used to detect boards, only the VID/PID. In practice I don't believe this changes anything since I've never seen FriendlyName match at this stage of the detection, even using the old registry method.
2. The VID/PID string used to detect is slightly different than in the past but it doesn't matter given how the BoardDefinition class and board definition files were using regex for matching: the current tests are all based on "start of string" so things still work fine.
3. The log messages look slightly different now since the FriendlyName from the registry is no longer used.

Here's debug output example of this in action on my machine. note that it correctly skips the ACPI device.

```
Debug	3/11/2022 8:05:59 AM	Checking for compatible module: VID_1B4F&PID_9206&REV_0100&MI_00
Debug	3/11/2022 8:05:59 AM	Found potentially compatible module (Arduino Micro Pro): VID_1B4F&PID_9206&REV_0100&MI_00@COM22
Debug	3/11/2022 8:05:59 AM	Incompatible module skipped: VID/PID: ACPI\VEN_PNP&DEV_0501
Debug	3/11/2022 8:05:59 AM	Checking for compatible module: VID_2341&PID_0042&REV_0001
Debug	3/11/2022 8:05:59 AM	Found potentially compatible module (Arduino Mega 2560): VID_2341&PID_0042&REV_0001@COM3
Debug	3/11/2022 8:05:59 AM	Checking for compatible module: VID_1B4F&PID_9206&REV_0100&MI_00
Debug	3/11/2022 8:05:59 AM	Found potentially compatible module (Arduino Micro Pro): VID_1B4F&PID_9206&REV_0100&MI_00@COM15
Debug	3/11/2022 8:05:59 AM	Checking for compatible module: VID_1B4F&PID_9206&REV_0100&MI_00
Debug	3/11/2022 8:05:59 AM	Found potentially compatible module (Arduino Micro Pro): VID_1B4F&PID_9206&REV_0100&MI_00@COM17
Debug	3/11/2022 8:05:59 AM	Checking for compatible module: VID_1B4F&PID_9206&REV_0100&MI_00
Debug	3/11/2022 8:05:59 AM	Found potentially compatible module (Arduino Micro Pro): VID_1B4F&PID_9206&REV_0100&MI_00@COM7
```

### DIFF
--- a/MobiFlight/MobiFlightCache.cs
+++ b/MobiFlight/MobiFlightCache.cs
@@ -177,12 +177,14 @@ namespace MobiFlight
         {
             Log.Instance.log("MobiFlightCache.LookupAllConnectedArduinoModulesAsync: Start", LogSeverity.Debug);
             List<MobiFlightModuleInfo> result = new List<MobiFlightModuleInfo>();
+            string[] connectedPorts = SerialPort.GetPortNames();
 
             if (_lookingUpModules) return result;
             _lookingUpModules = true;
             
             List<Task<MobiFlightModuleInfo>> tasks = new List<Task<MobiFlightModuleInfo>>();
             var supportedPorts = getSupportedPorts();
+            List<string> connectingPorts = new List<string>();
             
             for (var i = 0; i != supportedPorts.Count; i++)
             {
@@ -190,6 +192,19 @@ namespace MobiFlight
                 String portName = port.Key;
                 Board board = port.Value;
                 int progressValue = (i * 25) / supportedPorts.Count;
+
+                if (!connectedPorts.Contains(portName))
+                {
+                    Log.Instance.log("MobiFlightCache.LookupAllConnectedArduinoModulesAsync: Port not connected ("+portName+")", LogSeverity.Debug);
+                    continue;
+                }
+                if (connectingPorts.Contains(portName))
+                {
+                    Log.Instance.log("MobiFlightCache.LookupAllConnectedArduinoModulesAsync: Port already connecting (" + portName + ")", LogSeverity.Debug);
+                    continue;
+                }
+
+                connectingPorts.Add(portName);
 
                 tasks.Add(Task.Run(() =>
                 {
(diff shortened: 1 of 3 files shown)


DocMoebiuz:
@neilenns I would merge this so that we can use it in a beta and get some feedback.

neilenns:
If you approve it I think you can still merge it even tho I have it marked as draft

neilenns:
Feel free!

DocMoebiuz:
i remember i had two occasions in the past where even in the device manager two connected modules had the same serial port

I think we should keep the check whether  we are already trying to connect - just in case to be able to log this case as an error but don't take a special action. That might allow us to find out if it works for the people during beta but with the log message it will be easier to pin down in case somebody reported a problem.

neilenns:
@DocMoebiuz Reverted the change to the connection test, should be ready for you now.

PR #781, opened 2022-05-02
Match FTDI-style VID/PIDs

### LINKED ISSUE #780: Boards with FTDI chips are not recongnized anymore
**Describe the bug**
With the new USB detection boards with FTDI chip gets not detected.

**Steps To Reproduce**
1. Connect a board with FTDI chip (e.g. Mega2560 clone)

**Expected behavior**
Board should get recognized.

**Actual behavior**
This board does only show up as unknown board.

**Details**
- latest version of MF

**Additional context**
See also attached log file.

As per @neilenns advice, with adding `^COMPORT&VID_0403&PID_6001` as a VID/PID entry to arduino_mega.board.json the board gets detected and changes/additions/deletions are possible.

neilenns:
Fixes #780 

Instead of doing a substring find to strip off `USB` and `FTDI` from the raw VID/PID use a regex to specifically identify `VID_` and pull out the rest of the string. That should work across FTDI devices as well, like the ones mentioned in the bug.

No real user behaviour changes however the log messages will now report random non-USB devices, e.g. like this:

```
Debug	5/2/2022 12:10:45 PM	Skipping device with no available VID/PID (BTHENUM\{00001101-0000-1000-8000-00805f9b34fb}_LOCALMFG&0000)
```

Having this logged seemed better than adding more checks for things that start with USB or FTDI or COMPORT just to avoid a log message.

### DIFF
--- a/MobiFlight/MobiFlightCache.cs
+++ b/MobiFlight/MobiFlightCache.cs
@@ -109,6 +109,7 @@ namespace MobiFlight
         {
             var portNameRegEx = "\\(.*\\)";
             var result = new Dictionary<string, Board>();
+            var regex = new Regex(@"(?<id>VID_\S*)"); // Pattern to match the VID/PID of the connected devices
 
             // Code from https://stackoverflow.com/questions/45165299/wmi-get-list-of-all-serial-com-ports-including-virtual-ports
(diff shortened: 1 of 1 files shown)


elral:
With these changes boards with FTDI are detected again without modifying the board.json file.
Reading/changing/deleting devices and uploading to the board works as expected.
See also my log.txt
[log.txt](https://github.com/MobiFlight/MobiFlight-Connector/files/8608930/log.txt)
.

PR #823, opened 2022-05-30
Check for max number of output shifters

### LINKED ISSUE #821: Max device limits are not enforced for output shifters
Looks like the maximum limits of devices like shift registers are not enforced.

Like for Pro Micro, there is defined MAX_OUTPUT_SHIFTERS 2 for Pro Micro and 4 for Mega in boards definition file on the firmware. But it appears I can just add as many as I like on MF Connector side, until I generally run out of pins.

The flashing seems to work too, but I dont have enough shift registers to prototype if they all actually work.

image

neilenns:
Fixes #821 

* Add a check for maximum number of output shifters when one is added.
* Expose the `MaxShifters` board.json property via the `Board` class. Somehow this got missed during original implementation!

image

### DIFF
--- a/MobiFlight/Board.cs
+++ b/MobiFlight/Board.cs
@@ -165,6 +165,11 @@ namespace MobiFlight
         public int MaxServos = 0;
 
         /// <summary>
+        /// Maximum number of output shifters supported by the board.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


neilenns:
Corrected the error message, uses the right device type now.

image

Issue #850, opened 2022-06-07
Shift Register Names are not displayed on the mappings screen


simontrhodes:
**Describe the bug**
On the mappings screen, the names of Shift Registers are not displayed in the Output column.

**Steps To Reproduce**

1. Add a Shift Register to an Arduino.
2. At an Output using the added Shift Register
3. The Shift Register name is not displayed on the grid.

**Expected behavior**
The name of the Shift Register should be displayed.

**Actual behavior**
The name is shown as -

**Details**
 - MobiFlight Version: 9.4.0 and above

**Additional context**

DocMoebiuz:
Did not auto close.
Please use fixes #myissuenumber in the PR next time. Thanks!

PR #851, opened 2022-06-07
Shift Register Names are not displayed on the mappings screen #850

simontrhodes:
Shift Register Names are not displayed on the mappings screen #850


neilenns:
Can you include a screenshot showing what it looks like after this fix? Does it match with how input shifters display?

neilenns:
Thanks! Makes sense

DocMoebiuz (approved):
Thanks!

GioCC:
Judging from the fix, shouldn't the same correction apply to multiplexers?

DocMoebiuz:
i thought i did already ... i will check

GioCC:
Scratch that... late night glitch on my part. Muxes are inputs, issue here is on the output config side.

DocMoebiuz:
fixed #850

Issue #857, opened 2022-06-17
User always have to select which Arduino board they are using, even if there is only one.


simontrhodes:
**Is your feature request related to a problem? Please describe.**
In the display tab of the output config wizard, you can select which Arduino board you want to select.
This field is named as the module.
If you only have one Arduino board, you still have to select it from the list.

**Describe the solution you'd like**
If there is only one module, it is pre-selected for you.

**Describe alternatives you've considered**
None

**Additional context**
Code already modified for the output tab.
I couldn't work out where the code is for the input tab.

DocMoebiuz:
fixed by https://github.com/MobiFlight/MobiFlight-Connector/pull/858

PR #858, opened 2022-06-17
User always have to select which Arduino board #857

simontrhodes:
User always have to select which Arduino board they are using, even if there is only one. #857


DocMoebiuz (approved):
Cool idea. I will merge it and see how the feedback is.

PR #870, opened 2022-07-14
Add assigned pin display to output shifter

### LINKED ISSUE #868: Output shifters
Output shift registers should display the pin number in the Output column:

image

Input shifters have this, which would also be consistent if output shifters used Shiftername:Pin notation;

image

neilenns:
Fixes #868 

Switch to using `ToString()` to get the display string for an output shifter, then put the logic for constructing the pin assignment display string in that method.

image

This is what it looks like with a single pin assignment and multiple pin assignments. It's not possible to have no pin assignments given how the UI is built, so no need to handle that special case.


DocMoebiuz (approved):
Unit Tests would have been cool. There you can also cover the different cases and reduce the comments in the method.

Issue #867, opened 2022-07-12
Mobiflight test mode in output shifter does not turn off last led when mode is turned off


JaimeLeon2:
**Describe the bug**
In recent tests with two output registers, Mobiflight test mode turned all 16 leds on in sequence correctly, but when test mode was turned off, the last LED being tested remains on.

**Steps To Reproduce**
1. Configure arduino board for output shift register with 8 leds (I actually did 16)
2. Create output configs for 16 leds
3. Press Test to initiate test mode
4. Note leds light up in sequence
5. Press Stop
6. Note last LED remains on.

**Expected behavior**
After turning off test mode, all LEDs should be off.

**Actual behavior**
Last LED remains on.

**Details**
 - MobiFlight Version: 9.4.0.4
 - FlightSim: irrelevant
 - AddOn: irrelevant

**Additional context**
Recent test mode fix apparently did not catch this minor bug.

neilenns:
This is a firmware bug can you open this in the MobiFlight-FirmwareSource repo @JaimeLeon2? Thanks!

DocMoebiuz:
fixed by #878

PR #878, opened 2022-08-05
ShiftRegister Pins > 8  turn off correctly when stopping test mode

DocMoebiuz:
The way the Stop method was written worked only for 8-bit shift register.
Now it will work for all pins.


DocMoebiuz:
fixes #867

PR #899, opened 2022-09-04
Proof of concept flashing Picos via copy to mounted USB drive.

neilenns:
This is just a proof of concept illustrating the possibility of flashing Picos directly from MobiFlight by copying the firmware over with a simple file copy to the removable disk.

It is based on the approach used by PlatformIO to copy to Picos without using picotool. Thanks to @elral for digging in to how PlatformIO is now doing this.

I'm amazed at how simple this was to get working. The code to force a COM reset is already there for Arduinos and gets reused as-is. Adding a new block to the board config file with USB flash settings is the trigger to run the new code flow. The new code is straightforward as well.

I still can't believe this actually works and took me less than an hour to write!


Boards/raspberrypi_pico.board.json
```diff
@@ -0,0 +1,199 @@
+{
+  "$schema": "./mfboard.schema.json",
+  "UsbDriveSettings": {
```
neilenns, line 3: The presence of `UsbDriveSettings` in the board config file will trigger that flash method instead of using AvrDude. Way back when I added board configs I planned for multiple flashing methods so this fit right in.

MobiFlight/MobiFlightFirmwareUpdater.cs
```diff
@@ -38,31 +39,108 @@ public static bool Update(MobiFlightModule module)
...
 
-            while (!SerialPort.GetPortNames().Contains(Port))
-            {
-                System.Threading.Thread.Sleep(100);
-            }
-
             if (module.Board.AvrDudeSettings != null)
             {
-                try {
+                while (!SerialPort.GetPortNames().Contains(Port))
```
neilenns, line 44: This serial port searching only applies to avrdude flashing so I moved it inside the avrdude flash block.

MobiFlight/MobiFlightFirmwareUpdater.cs
```diff
@@ -38,31 +39,108 @@ public static bool Update(MobiFlightModule module)
...
                     result = false;
                 }
 
                 if (module.Board.Connection.DelayAfterFirmwareUpdate > 0)
                 {
                     System.Threading.Thread.Sleep(module.Board.Connection.DelayAfterFirmwareUpdate);
                 }
-            } else
+            }
+            else if (module.Board.UsbDriveSettings != null)
```
neilenns, line 62: If the USB drive settings are present then try a USB flash instead. This is basically an exact copy of the AvrDude code and could probably be refactored a bit so the test is done in the try{} block and the delay code isn't duplicated.

MobiFlight/MobiFlightFirmwareUpdater.cs
```diff
@@ -38,31 +39,108 @@ public static bool Update(MobiFlightModule module)
...
+                }
+            }
+            else
             {
                 Log.Instance.log($"Firmware update requested for {module.Board.Info.MobiFlightType} ({module.Port}) however no update settings were specified in the board definition file. Module update skipped.", LogSeverity.Warn);
             }
             return result;
         }
 
+        public static void FlashViaUsbDrive(Board board)
```
neilenns, line 86: The meat of the flashing. It's super simple! Check out the code comments for details on what's going on.

neilenns:
Before this is ready for prime time there's some testing that has to be done:

1. Verify Arduino flashing still works.
2. Verify a USB stick called `RPI-RP2` without an `INFO-UF2.TXT` file present fails (as it should)
3. Verify flashing when more than one Pico is connected and all of them need updates

elral:
Great! I will test it also with more than one Pico.

Ralf

neilenns:
This needs to be updated to handle the new firmware reset flow.

neilenns:
Will make a new PR

PR #901, opened 2022-09-05
Add checks for in-use I2C pins and no I2C pins defined for the board.

### LINKED ISSUE #900: I2C devices can be added even if the I2C pins are used by other non-I2C modules
**Describe the bug**

**Steps To Reproduce**
1. Configure an encoder to use the I2C pins for the Arduino (e.g. pins 2 and 3 for a Micro)
2. Configure an LCD display

**Expected behavior**
An error when trying to add the LCD display because the I2C pins are in use by a non-I2C module.

**Actual behavior**
You can add the LCD display even though the pins are already used by the encoder.

**Additional context**
Discovered when working on Pico support. This causes the Pico to take too long to restart after uploading the config so MobiFlight thinks the upload failed. Arduinos seem to handle this a bit better and restart quickly, even though it is incompatible duplicate pin usage.

neilenns:
Fixes #900 

* Before adding an LCD display check for I2C pins being defined in the board definition file. If there aren't any then throw an error.
* When adding an LCD display check to see if any of the I2C pins are already assigned to non-I2C devices. If yes throw an error.

### DIFF
--- a/UI/Panels/Settings/MobiFlightPanel.cs
+++ b/UI/Panels/Settings/MobiFlightPanel.cs
@@ -565,8 +565,31 @@ namespace MobiFlight.UI.Panels.Settings
                             throw new MaximumDeviceNumberReachedMobiFlightException(MobiFlightLcdDisplay.TYPE, tempModule.Board.ModuleLimits.MaxLcdI2C);
                         }
 
+                        // Check and see if any I2CPins exist in the board definition file. If not then there's no way to add an LCD device
+                        // so throw an error.
+                        var I2CPins = tempModule.Board.Pins.FindAll(x => x.isI2C);
+
+                        if (!(I2CPins?.Any() ?? false))
+                        {
+                            throw new I2CPinsNotDefinedException(MobiFlightLcdDisplay.TYPE);
+                        }
+
+                        // Fix for issue #900. Check for any I2C pins that might already be in use by
+                        // other modules configured for the device.
+                        var pinsInUse = getVirtualModuleFromTree().GetPins(false, true).Where(x => x.Used);
+
+                        // Check and see if any I2C pins are in use. This only looks for the first I2C pin that's
+                        // in use since that's sufficient to throw an error and tell the user what to do. Trying to
+                        // write an error message that works for one or more in use I2C pins is way more trouble
+                        // than it's worth.
+                        var inUseI2CPin = I2CPins.Find(x => pinsInUse?.Contains(x) ?? false);
+
+                        if (inUseI2CPin != null)
+                        {
+                            throw new I2CPinInUseException(MobiFlightLcdDisplay.TYPE, inUseI2CPin);
+                        }
+
                         cfgItem = new MobiFlight.Config.LcdDisplay();
-                        // does not deal yet with these kind of pins because we use I2C
                         break;
 
                     case "ShiftRegisterToolStripMenuItem":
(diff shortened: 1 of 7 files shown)


UI/Panels/Settings/MobiFlightPanel.cs
```diff
@@ -565,8 +565,31 @@ private void addDeviceTypeToolStripMenuItem_Click(object sender, EventArgs e)
                             throw new MaximumDeviceNumberReachedMobiFlightException(MobiFlightLcdDisplay.TYPE, tempModule.Board.ModuleLimits.MaxLcdI2C);
                         }
 
+                        // Check and see if any I2CPins exist in the board definition file. If not then there's no way to add an LCD device
+                        // so throw an error.
+                        var I2CPins = tempModule.Board.Pins.FindAll(x => x.isI2C);
+
+                        if (!(I2CPins?.Any() ?? false))
```
neilenns, line 572: This is a basic test for any I2C pins being defined in the board definition file. Technically I guess this could check for exactly two pins being defined as I2C but that seems soooooo edge case.

UI/Panels/Settings/MobiFlightPanel.cs
```diff
@@ -565,8 +565,31 @@ private void addDeviceTypeToolStripMenuItem_Click(object sender, EventArgs e)
...
+
+                        // Fix for issue #900. Check for any I2C pins that might already be in use by
+                        // other modules configured for the device.
+                        var pinsInUse = getVirtualModuleFromTree().GetPins(false, true).Where(x => x.Used);
+
+                        // Check and see if any I2C pins are in use. This only looks for the first I2C pin that's
+                        // in use since that's sufficient to throw an error and tell the user what to do. Trying to
+                        // write an error message that works for one or more in use I2C pins is way more trouble
+                        // than it's worth.
+                        var inUseI2CPin = I2CPins.Find(x => pinsInUse?.Contains(x) ?? false);
```
neilenns, line 585: The `?.` and `??` handle the case where there are no pins in use on the board at all. in that situation `pinsInUse` is null.

neilenns:
@DocMoebiuz This needs German translations for the two new error messages and two new error message titles.

UI/Panels/Settings/MobiFlightPanel.cs
DocMoebiuz: Ugly syntax "??" - Contains(x) can be true or false if pinsInUse is not null... and if null, then it returns false. Later in 586 you test for non null... which a boolean always is.

Weird to read this part of the code.

UI/Panels/Settings/MobiFlightPanel.cs
DocMoebiuz: The naming of the variables could be improved to reduce the required lines of //comments

e.g I2CPins => AvailableI2Cpins

pinsInUse is clearly named.

inUseI2CPin => improve reading by naming it what the if test is doing: "IsAnyI2CPinAlreadyInUse"

UI/Panels/Settings/MobiFlightPanel.cs
DocMoebiuz: Test should simply be 

if (IsAnyI2CPinAlreadyInUse)

see naming proposal earlier.

DocMoebiuz (changes requested):

UI/Panels/Settings/MobiFlightPanel.cs
neilenns: Renamed `I2CPins` to `availableI2CPins`.

Renamed `inUseI2CPin` to `firstInUseI2CPin`. It doesn't wind up as a boolean, it winds up as the actual pin, so naming it `IsAnyI2CPinAlreadyInUse` doesn't make sense.

UI/Panels/Settings/MobiFlightPanel.cs
neilenns: It's not ugly, it's Linq ;)

`var firstInUseI2CPin = availableI2Cpins.Find(x => pinsInUse?.Contains(x) ?? false);` returns the first used pin or null if no I2C pins are in use, it's not a boolean return value. The part in the `Find()` clause is the test used to determine whether a given I2C pin is in the list of used pins.

The `??` works in conjunction with the `?.` to handle the case where no pins are currently in use on the board. Without that the `.Contains(x)` would throw a null reference exception.

The return result isn't a boolean, it's the first I2C pin that is in use by something else or null.

DocMoebiuz (approved):
Cool!

PR #915, opened 2022-09-25
Fixed output processing

Kavakuo:
Hi all,
I think the most recent PR (#909) introduced a bug regarding outputs. The refactoring of the `ExecuteConfig` function did not update the value passed into the `ExecuteDisplay` function. Therefore, this function was always called with an empty string.


MobiFlight-Admin:
Thanks for this PR. I will take a closer look tomorrow!

MobiFlight-Admin (approved):
Thanks. 
I will merge this. I will haev to think about 496, if that value should be parsed, if possible, but at the moment this feature is not enabled.

Issue #894, opened 2022-08-20
When adding 2nd multiplexer the base pins are not displayed correctly


DocMoebiuz:
**Describe the bug**
When adding 2nd multiplexer the base pins are not displayed correctly

**Steps To Reproduce**
1. Go to Mobiflight Settings
2. Click on a board
3. Add a first multiplexer
4. Upload the config
5. Close the Wizard
6. Open MobiFlight Board Settings again
7. Add another input multiplexer to the same board
8. Check the base settings

**Expected behavior**
Base settings show same pins for the multiplexer as first multiplexer

**Actual behavior**
Base settings show different pins for the multiplexer than first multiplexer

**Details**
 - MobiFlight Version: [e.g. 8.2.0]
 - FlightSim: [e.g. MSFS2020]
 - AddOn: [e.g. Default Aircraft]

**Additional context**
First Input Multiplexer
<img width="433" alt="image" src="https://user-images.githubusercontent.com/86157512/185745942-7eb13fb3-ca81-40b6-8265-89bd06c1d939.png">

Second Input Multiplexer
<img width="429" alt="image" src="https://user-images.githubusercontent.com/86157512/185745954-fdb78abb-ca4e-4115-9bfc-58ad249d349a.png">

Exception raised when accessing the Board Settings again.

ddt154:
Although its a bug, a work around is to manually modify the saved config file and then load it.  It'll then have the correct config and work properly.  You can then upload config to device and works fine.

PR #933, opened 2022-10-05
When adding 2nd multiplexer the base pins are not displayed correctly (#894)

GioCC:
Attribution of the multiplexer pins to the owner module was made according to attribute "Name", which apparently is not the name (and it's blank).
Changed to reference attribute "Text" instead.

### DIFF
--- a/UI/Panels/Device/MFMultiplexerDriverSubPanel.resx
+++ b/UI/Panels/Device/MFMultiplexerDriverSubPanel.resx
@@ -196,7 +196,7 @@
     <value>32</value>
   </data>
   <data name="labelMaster.Text" xml:space="preserve">
-    <value>Changing these values will affect every other device using multiplexers.</value>
+    <value>These values will be shared by all multiplexers of this module.</value>
   </data>
   <data name="&gt;&gt;labelMaster.Name" xml:space="preserve">
     <value>labelMaster</value>
--- a/UI/Panels/Settings/MobiFlightPanel.cs
+++ b/UI/Panels/Settings/MobiFlightPanel.cs
@@ -971,7 +971,7 @@ namespace MobiFlight.UI.Panels.Settings
         {
             MobiFlight.Config.MultiplexerDriver moduleMultiplexerDriver;
             
-            string moduleName = getModuleNode().Name;
+            string moduleName = getModuleNode().Text;
 
             if (!moduleMultiplexerDrivers.ContainsKey(moduleName)) {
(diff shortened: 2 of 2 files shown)


DocMoebiuz:
@GioCC thanks for looking into this. I am surprised that the name is not set. Maybe a "selectNode" update problem?

UI/Panels/Device/MFMultiplexerDriverSubPanel.resx
```diff
@@ -196,7 +196,7 @@
     <value>32</value>
   </data>
   <data name="labelMaster.Text" xml:space="preserve">
-    <value>Changing these values will affect every other device using multiplexers.</value>
+    <value>Changing these values will affect every other device in this module using multiplexers.</value>
```
DocMoebiuz, line 199: Do we want to say:
Changing these values will affect all other multiplexers of this module.

GioCC: > Do we want to say: Changing these values will affect all other multiplexers of this module.
I don't really see much of a difference meaning-wise, but I like the fact that it's shorter.
> it sounds almost like a warning. but it is not.
But it _is_ a warning. That's why it has been placed there.
> These values will be shared by all multiplexers of this module.
That's another equivalent way to put it. But it just seems like a lazy reminder, which doesn't alert the user that he might be doing something unintended - which defeats its purpose.

UI/Panels/Settings/MobiFlightPanel.cs
```diff
@@ -971,7 +971,7 @@ private MobiFlight.Config.MultiplexerDriver getModuleMultiplexerDriver()
         {
             MobiFlight.Config.MultiplexerDriver moduleMultiplexerDriver;
             
-            string moduleName = getModuleNode().Name;
+            string moduleName = getModuleNode().Text;
```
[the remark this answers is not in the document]
GioCC: Right, that explains it.

UI/Panels/Device/MFMultiplexerDriverSubPanel.resx
DocMoebiuz: i disagree. It's not supposed to be a warning, it's supposed to be a reminder, a hint. That's why i like the last option best.

These values will be shared by all multiplexers of this module.

DocMoebiuz (approved):
Thanks @GioCC for fixing this so quickly!!!

DocMoebiuz:
fixes #894

Issue #926, opened 2022-10-04
Add Multiplexer to a Pro Micro board fails


JaimeLeon2:
**Describe the bug**
Adding a Multiplexer to a Pro Micro board fails.  The device details page does not appear.

**Steps To Reproduce**
1. Connect a Pro Micro board and open Mobilfight
2. Go to Mobiflight Modules
3. Select the pro micro board and Add device
4. Select Multiplexer
5. Note the device is added to the board list, but no details page is presented.

![image](https://user-images.githubusercontent.com/86563553/193714344-a4aae8ad-a966-46bd-92e5-bd846d33fc8f.png)

**Expected behavior**
Details page is presented and device configuration proceeds normally

**Actual behavior**
No details page is shown by MFConector.

**Details**
 - MobiFlight Version: 9.5.0
 - FlightSim: irrelevant
 - AddOn: irrelevant

**Additional context**
The operation failed for two different pro micros connected.  I also tested adding a multiplexer to a mega board and to a Nano board and in both these cases, the details page was correctly shown.

![image](https://user-images.githubusercontent.com/86563553/193714358-018ad9bf-3547-4784-affe-9a74219ef457.png)

![image](https://user-images.githubusercontent.com/86563553/193714381-35fb3ca7-f508-4c77-8ed2-41ac191b4fff.png)

PR #934, opened 2022-10-05
Add Multiplexer to a Pro Micro board fails (#926)

GioCC:
Mux selector pins with value <=0 caused the object to be considered as uninitialized (initial values are set to negative numbers); should have correctly been "< 0".
Only affected ProMicro because on other Arduinos pin 0 is reserved and it's not available for assignment.


DocMoebiuz:
fixes #926

PR #958, opened 2022-10-10
Store VID/PID with MobiFlightModule and MobiFlightModuleInfo

DocMoebiuz:


MobiFlight/MobiFlightModule.cs
```diff
@@ -121,6 +120,8 @@ public enum Command
         String _comPort = "COM3";
         public String Port { get { return _comPort; } }
         public String Name { get; set; }
+        public string HardwareId { get; set; }
```
neilenns, line 123: Does this ever get set anywhere though? I don't think it does.

Issue #883, opened 2022-08-14
Spacemouse compact left/right button click causes .net exception


windingdk:
**Describe the bug**
I use this device (https://3dconnexion.com/uk/product/spacemouse-compact/) to control views. When I click on left/right mouse button I get an exception (see attached txt file)
[bug_spacemouse_exception.txt](https://github.com/MobiFlight/MobiFlight-Connector/files/9333620/bug_spacemouse_exception.txt)

Issue is 100% reproducible.

I have installed VS and tried to debug code. In UpdateButtons method in joystick.cs the Buttons list is empty (DeviceId = Buttons[i].Label).
```c#
private void UpdateButtons(JoystickState newState)
        {
            for (int i = 0; i != newState.Buttons.Length; i++)
            {
                if (!StateExists() || state.Buttons.Length < i || state.Buttons[i] != newState.Buttons[i])
                {
                    if (newState.Buttons[i] || (state != null))

                            OnButtonPressed?.Invoke(this, new InputEventArgs()
                            {
                                DeviceId = Buttons[i].Label,
                                Serial = SerialPrefix + joystick.Information.InstanceGuid.ToString(),
                                Type = DeviceType.Button,
                                Value = newState.Buttons[i] ? 0 : 1
                            });
                }
            }
        }
```

I have checkedout main/ 582465a322be48d5505773c4cdaf768da50dcb08

I initialy found the bug in version 9.4.0 download here https://www.mobiflight.com/en/download.html

Further details: I have single stepped in EnumerateDevices to get more details
<img width="719" alt="bug_mobiflight" src="https://user-images.githubusercontent.com/14841425/184544479-6d8b9441-9263-4227-8abc-1c6f8508df6c.png">

Method never finds IsAxis, IsButton, IsPOV. It finds a single "DeviceObjectTypeFlags.RelativeAxis".

```c#
private void EnumerateDevices()
        {
            foreach (DeviceObjectInstance device in this.joystick.GetObjects())
            {

                this.joystick.GetObjectInfoById(device.ObjectId);


                int offset = device.Offset;
                int usage = device.Usage;
                ObjectAspect aspect = device.Aspect;
                String name = device.Name;

                bool IsAxis = (device.ObjectId.Flags & DeviceObjectTypeFlags.AbsoluteAxis) > 0;
                bool IsButton = (device.ObjectId.Flags & DeviceObjectTypeFlags.Button) > 0;
[...]

PR #967, opened 2022-10-13
Joystick: Skipping Update Buttons if no buttons are registered.

DocMoebiuz:
Also changed the if condition from `!=` to `<`

Does not resolve the problem for the user as described in his issue #883 
But it does prevent an exception when pressing a button.


DocMoebiuz:
related #967

PR #987, opened 2022-10-27
Fix column name when updating input preconditions

### LINKED ISSUE #986: System.ArgumentException in ExecutionManager.cs:1461 (UpdateInputPreconditions())
**Describe the bug**
An argument exception gets thrown every time line 1461 is executed.

```
        private void UpdateInputPreconditions()
        {
            inputsDataGridView.SuspendLayout();
            foreach (DataGridViewRow gridViewRow in inputsDataGridView.Rows)
            {
                try
                {
                    if (gridViewRow.DataBoundItem == null) continue;
                    if (!(bool)gridViewRow.Cells["active"].Value) continue;
```

It's the last line above that throws the exception.

**Steps To Reproduce**
1. Create an input configuration
2. Hit run with a debugger attached

**Expected behavior**
No exception

**Actual behavior**
Exception on line 1461: System.ArgumentException 'Column name active cannot be found. Parameter name: columnName'.

**Details**
Version 9.5.0.3 as built from `main`.

neilenns:
Resolves #986 

The code that checks to see if the input line is active is using the wrong column name. It should be `inputActive` instead of `active`.

### DIFF
--- a/MobiFlight/ExecutionManager.cs
+++ b/MobiFlight/ExecutionManager.cs
@@ -1458,7 +1458,7 @@ namespace MobiFlight
                 try
                 {
                     if (gridViewRow.DataBoundItem == null) continue;
-                    if (!(bool)gridViewRow.Cells["active"].Value) continue;
+                    if (!(bool)gridViewRow.Cells["inputActive"].Value) continue;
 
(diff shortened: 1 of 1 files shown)


DocMoebiuz:
i am almost tempted to rename the column to active :D would even make more sense.

Issue #705, opened 2022-02-23
Re-categorize all log messages


neilenns:
Right now the log messages are assigned to different categories in a pretty haphazard way. It's understandable given how the code has grown over time with lots of contributors, but at this point unless you're running in `debug` log level the logs aren't particularly useful.

@DocMoebiuz suggested in #703 that we rationalize the current log entries. So... let's do that.

Here's a spreadsheet with all the current log messages in the code base, their current log level, and a proposed new log level: https://1drv.ms/x/s!AnZYhMlrZ6jRn-sAfklQr8YzzLmp1g?e=u2gE3g. Here's how I grouped items:

1. If it was an exception it went in `Error`
2. If it was basic information about normal operation, such as detecting boards, flashing firmware successfully, or sending/receiving messages to connected devices, it went into `Info`.
3. If it was error cases, random other bits of info, or things that generally aren't needed during normal operation they went in `Debug`.
4. I didn't put anything in `Warn`.

This is just a first pass proposal and we can certainly move things between the categories. My goal was to make it so running with `Info` log level would be sufficient for most uses: checking that Arduinos/Joysticks got detected, that buttons/encoders/Joystick inputs are getting picked up, and that output data is getting sent to the connected devices.

neilenns:
A good way to look at these is to filter by each of the different log levels in the `New Level` column. E.g. set the filter to just `Info` and you can get a nice overview of what the log window would look like with that level selected.

I also suggest when making this change that I go through and update the way the logs are formatted to use C# interpolation (`$""`) instead of the older style string concatenation that's used.

tigert:
I agree that "Warn" might be a bit redundant - I dont know what would fit there - maybe if we had some kind of deprecation scheme on hubhop when deleting some presets and someone is using them, those could go there. But we don't do this right now.

But other than that, things kinda are normal operation (info) or something goes wrong (error).

DocMoebiuz:
@neilenns  is the document still available?

neilenns:
Yes, did the link work?

DocMoebiuz:
Not for me

neilenns:
Try this one @DocMoebiuz https://1drv.ms/x/s!AnZYhMlrZ6jRn9MP6VZ4xyBIIxYnmQ?e=WVDt3A

neilenns:
[...]

PR #994, opened 2022-11-01
Update log levels

neilenns:
* Update the logged level for all messages based on the spreadsheet in #705.
* For any exception that's logged via a log message make sure the object is called `ex`.
* Add a period to the end of all messages, other than those that log an exception message directly.
* Automatically add the calling class and method name to `Error` logs
* Random little bits of message re-writing for consistency and clarity


DocMoebiuz (approved):
Nice!

PR #999, opened 2022-11-02
Add log when XML parsing fails

### LINKED ISSUE #998: Log XML parsing errors when they happen
If there's an error in the config XML an error dialog is shown to the user but the underlying syntax error that caused the problem isn't logged. This makes it challenging to figure out what is wrong: you have to get the settings file, open it in something like Visual Studio, and look for syntax error highlighting.

Log the exception message in the log like we do for other exceptions.

Not do this and just rely on debugging instead, which is lame.

neilenns:
Fixes #998 

Add a log error when XML parsing fails with the reason for the failure. Looks like this:

```
Error	11/02/2022 15:06:45	Unable to load configuration file: '', hexadecimal value 0x0B, is an invalid character. Line 1, position 524.
```

The line and position in the exception isn't accurate for some reason but the actual error is and made it quick to find out what was wrong.

### DIFF
--- a/UI/MainForm.cs
+++ b/UI/MainForm.cs
@@ -1153,6 +1153,7 @@ namespace MobiFlight.UI
             }
             catch (Exception ex)
             {
+                Log.Instance.log($"Unable to load configuration file: {ex.Message}", LogSeverity.Error);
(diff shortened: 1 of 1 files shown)


DocMoebiuz (approved):
Good idea!

Issue #925, opened 2022-10-02
Module focus is not changed when clicking + in another module 


JaimeLeon2:
**Describe the bug**
In MF Modules, when clicking on the + icon of another board to open the configured devices list, the focus is maintained on the first board on the list. 

**Steps To Reproduce**
1. Open MF Modules (several boards connected) 
2. Note focus is on first board 
3. Click on + of last board 
4. Note the device list opens, but the focus remained on the first board. 

**Expected behavior**
Focus changes to the board selected and displays the board information. 

**Actual behavior**
Focus remains on the first board of the list, even though the device list was opened on the selected board 

**Details**
 - MobiFlight Version: 9.5.0
 - FlightSim: irrelevant 
 - AddOn: irrelevant 

**Additional context**

DocMoebiuz:
@JaimeLeon2 I cannot reproduce the problem with the latest Beta and also not with 9.5.0
Can you please check the instructions for recreating the issue:

![925-cannot-reproduce](https://user-images.githubusercontent.com/86157512/195716674-c53052df-cf9f-458c-862c-18e341b4fe74.gif)

JaimeLeon2:
I tried to post a video but the file was too big.  The issue happens consistently in my PC with 7 boards with the procedure I described.  Maybe you could try with five or more boards connected.  

Further testing throws even weirder results.  The first board listed is mega but my two pro micros are fourth and sixth on the list.  Well, when I open the Modules window, the first board is selected, if I click any of the pro micros, the focus is correctly changed to them, but not to any other board on the list.  The device list opens in all cases.
Maybe we can so a screen share so I can show you.

First when the Modules window is opened
![image](https://user-images.githubusercontent.com/86563553/195746997-6cab7663-dbb7-468d-a84f-4525cccf50f9.png)

Then after clicking on the second board + icon.
![image](https://user-images.githubusercontent.com/86563553/195747049-4887b11f-f243-41fb-812f-bedad86cc4e8.png)

neilenns:
@JaimeLeon2 does this still happen? There've been several changes to this dialog since this bug was opened.

DocMoebiuz:
I will close this issue because not able to reproduce

PR #1001, opened 2022-11-05
Enable the upload button after opening a config

### LINKED ISSUE #1000: Upload config button doesn't activate after opening a config
**Describe the bug**
Upload config button doesn't activate after opening a config

**Steps To Reproduce**
1. Go to `Mobiflight Modules`
2. Select one of your boards
3. Open a config file

Button

**Expected behavior**
The `Upload config` button activates so you can upload the config.

**Actual behavior**
The `Upload config` button doesn't activate

**Details**
 - MobiFlight Version: 9.6.0

**Additional context**
If you expand the module and select any of the devices that will cause the button to activate.

neilenns:
Fixes #1000 

Enable the upload button whenever a config is loaded successfully.

### DIFF
--- a/UI/Panels/Settings/MobiFlightPanel.cs
+++ b/UI/Panels/Settings/MobiFlightPanel.cs
@@ -848,6 +848,7 @@ namespace MobiFlight.UI.Panels.Settings
 
                 moduleNode.ImageKey = "Changed";
                 moduleNode.SelectedImageKey = "Changed";
+                uploadToolStripButton.Enabled = true;
             }
         }
 


UI/Panels/Settings/MobiFlightPanel.cs
```diff
@@ -848,6 +848,7 @@ private void openToolStripButton_Click(object sender, EventArgs e)
 
                 moduleNode.ImageKey = "Changed";
                 moduleNode.SelectedImageKey = "Changed";
+                uploadToolStripButton.Enabled = true;
```
DocMoebiuz, line 851: there is nothing in the code that disables the button on opening inside this function. That's why I find the proposed solution doesn't fix the problem.

In your screen recoding the device is not selected by the user, but when you click the open button- i would guess that "afterselect"-event is not properly triggered in this case.

So the real fix should be to properly "select" the node and trigger the `mfModulesTreeView_AfterSelect` correctly.

Expected behavior:
* If no board is selected, the toolbar should be deactivated.
* If one board is selected, the toolbar items become active.
* If I open a new config, the board should stay selected as active, and all buttons in the toolbar therefore active too.

DocMoebiuz (changes requested):
see comment

DocMoebiuz:
Possibly related to #925

neilenns:
@DocMoebiuz I agree the fix in this PR isn't great. I couldn't find a good way to fire the actual existing method for when selection changes. I tried forcing the selection to the active board but that didn't work. Am open to other ideas, or if you want to try coding up a fix.

DocMoebiuz:
I don't want to spend too much effort on this but I wonder why the UI doesn't behave correctly in the first place. At what circumstance is it greying out the button? I looked through the AfterSelect method - as soon as the board is a MF board and gets selected then the toolbar should become enabled.

We have the other issue #925 where this could be the same root cause that triggers this behavior.

To reproduce this, what is needed?

A fresh MF board with no devices?
what else?

neilenns:
@DocMoebiuz My suspicion is one of these two situations are what makes it happen:

* Only one board connected
* The board connected has no devices on it yet

Haven't tested it deeply to find out which case causes it though.

When I was debugging I did confirm that AfterSelect is triggered when I first open the dialog, but it never gets called again after I click on the Open button. Forcing it to get called after the Open dialog closes is a good solution but I couldn't get that to work.

PR #1010, opened 2022-11-15
7 segment support decimal

### LINKED ISSUE #1009: Support decimal display on 7-segments
Now that we are working internally with floats, it would be great to be able to display floats on 7-segment displays with the decimal point in the right position.

I can display a decimal value on the 7-segment modules, and the decimal point will be set correctly automatically.

Manually setting the decimal point and having to multiply and round the value.

**Additional context**
As part of the float support.

DocMoebiuz:
fixes #1009 

Confirmations:
* decimal value can be sent to the 7-segment display
* decimal point is displayed in the correct position next to the the number without using an additional 7-segment digit
* masking is correctly applied, e.g. if we display the number only on the last three digits of an 8 digit module the decimal point is placed correctly

### DIFF
--- a/MobiFlight/MobiFlightLedModule.cs
+++ b/MobiFlight/MobiFlightLedModule.cs
@@ -66,12 +66,10 @@ namespace MobiFlight
 
             var command = new SendCommand((int)MobiFlightModule.Command.SetModule);
 
-            var strValueParts = value.Split('.');
-            value = String.Join("", strValueParts);
-
-            if (strValueParts.Length > 1)
+            if (value.IndexOf(".") >=0 )
             {
-                points = CalculateCorrectPoints(String.Join(".", strValueParts), mask);
+                points = CalculateCorrectPoints(value, mask);
+                value = value.Replace(".", "");
             }
 
             // clamp and reverse the string
@@ -101,6 +99,7 @@ namespace MobiFlight
         static public byte CalculateCorrectPoints(string value, byte mask)
         {
             byte points = 0;
+            bool lastDigitWasPoint = false;
 
             // we start with the last character in the value string
             int positionInValue = value.Length-1;
@@ -123,7 +122,13 @@ namespace MobiFlight
                     points |= (byte)(1 << digit);
                     
                     // then we stay on the digit one more time
-                    digit--;
+                    // but only if we have not had a decimal point
+                    // on the last digit.
+                    if (!lastDigitWasPoint)
+                        digit--;
+                } else
+                {
+                    lastDigitWasPoint = false;
                 }
 
                 // walk one character to the left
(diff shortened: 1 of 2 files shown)


MobiFlight/MobiFlightLedModule.cs
```diff
@@ -65,6 +66,14 @@ public void Display(int subModule, String value, byte points, byte mask)
 
             var command = new SendCommand((int)MobiFlightModule.Command.SetModule);
 
+            var strValueParts = value.Split('.');
+            value = String.Join("", strValueParts);
```
neilenns, line 70: Instead of calling this `value` and then having to re-join the string later if there is a decimal, why not call this `valueWithoutDecimals`?

MobiFlight/MobiFlightLedModule.cs
```diff
@@ -65,6 +66,14 @@ public void Display(int subModule, String value, byte points, byte mask)
...
+
+            if (strValueParts.Length > 1)
+            {
+                points = CalculateCorrectPoints(String.Join(".", strValueParts), mask);
```
neilenns, line 74: Then here instead of having to re-join you can just pass in the original `value`.

MobiFlight/MobiFlightLedModule.cs
```diff
@@ -89,6 +98,40 @@ public void Display(int subModule, String value, byte points, byte mask)
...
 
+        static public byte CalculateCorrectPoints(string value, byte mask)
+        {
+            byte points = 0;
+
+            // we start with the last character in the value string
+            int positionInValue = value.Length-1;
+
+            // we go over all 8 potential digits
+            for (byte digit=0; digit<8; digit++)
```
neilenns, line 109: Should this be `var` instead of `byte`?

MobiFlight/MobiFlightLedModule.cs
```diff
@@ -89,6 +98,40 @@ public void Display(int subModule, String value, byte points, byte mask)
...
+                if (positionInValue < 0)
+                    break;
+
+                // when we have a decimal point at the current position
+                if (value[positionInValue]=='.')
+                {
+                    // activate the point at the current digit
+                    points |= (byte)(1 << digit);
+                    
+                    // then we stay on the digit one more time
```
neilenns, line 125: I don't understand this code comment. Why does it say "stay on the digit" but decrement `digit`?

neilenns (changes requested):
This also works with reversed display?

MobiFlight/MobiFlightLedModule.cs
neilenns: Thinking about this even more, why even do this at all? Just pass `value` to `CalcuateCorrectPoints()` and when you need the string without decimals do `value.Replace('.', '')`;

DocMoebiuz:
> This also works with reversed display?

good point. I don't know exactly but I hope that we will get the feedback during the beta phase.

PR #1011, opened 2022-11-16
Auto-Retrigger can be performed on Run

### LINKED ISSUE #1004: Initiate Synchronisation at startup
After the start of Mobiflight or a restart of the simulator or the corresponding software, the state of the hardware switches can differ from the state in the software.

There is already a function that can be triggered but you have to assign it to an input of your hardware.

It would be nice if there could be a checkbox in the settings dialogue to enable an automatic synchronisation on startup.

DocMoebiuz:
fixes #1004 

- [x] Re-trigger is performed automatically on "Run"
- [x] Auto-Retrigger can be controlled in settings (On, Off)
- [x] Auto-Retrigger is set to "On" (True) by default
- [x] Localization
- [ ] Documentation

mobiflight-auto-retrigger-built-in

### DIFF
--- a/MobiFlight/ExecutionManager.cs
+++ b/MobiFlight/ExecutionManager.cs
@@ -260,6 +260,7 @@ namespace MobiFlight
         {
             simConnectCache.Start();
             xplaneCache.Start();
+            OnStartActions();
             timer.Enabled = true;
         }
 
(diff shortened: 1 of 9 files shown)


neilenns:
Nice. Off by default for backwards compatibility purposes?

neilenns (approved):

DocMoebiuz:
> Nice. Off by default for backwards compatibility purposes?

Yes, that’s how I typically do it. But you’re right. Maybe we should turn it on by default since that is what users expect and if it bothers them they can turn it off.

I think I like that better.

PR #1015, opened 2022-11-16
Allow multiple baud rates for a given board

neilenns:
The Arduino Nano poses a problem: there are two variations in the wild that are the same board but with different bootloaders that use different baud rates.

Right now MobiFlight only supports a single baud rate for a given board. We try flashing at that baud rate and if it fails it fails and the user is out of luck.

We've discussed resolving this by shipping two board definition files with two different baud rates in it but that's silly. It's the same board with the same firmware and same pins. The only difference is the baud rate for initial firmware flashing.

This PR is an attempt to solve this by allowing multiple baud rates in a single board definition file. Flashing loops over all baud rates until it finds one that works or all fails.


DocMoebiuz:
<@414827083601281044> why do we have to try and error? I don't understand why we need to try baudrates. The baud rate that varies for the nano is only on uploading the firmware. this means it's purely an avrdude flash parameter. At this point Mobiflight is not involved in the communication. Once the firmware is uploaded MobiFlight will connect with the standard baud rate like all other boards

DocMoebiuz:
is it not acceptable from your POV to have two nano entries?

neilenns:
> <@414827083601281044> why do we have to try and error? I don't understand why we need to try baudrates. The baud rate that varies for the nano is only on uploading the firmware. this means it's purely an avrdude flash parameter. At this point Mobiflight is not involved in the communication. Once the firmware is uploaded MobiFlight will connect with the standard baud rate like all other boards

This change is specifically about loading the firmware, not about future connections. With the Nano it could be either of two baud rates and we have no way to know which one to use.

Supplying two different `.board.json` files that differ only in baud rate means wouldn't make for a great user experience. For some % of boards the flash would fail and the board would be considered ambiguous. They'd have to go into the modules dialog, right click, and pick the other board definition to try and flash (which could still fail if they don't know what kind of bootloader they ave).

With this change (assuming it works! I don't have the nanos to try) MobiFlight can just try each listed baud rate until one works. It's completely transparent to the user, doesn't require maintaining multiple board definition files, and doesn't expose two board definitions to the user for what in their mind is just a "Nano".

neilenns:
Will re-open with new details

PR #1016, opened 2022-11-22
Support Nanos

### LINKED ISSUE #579: Support Arduino Nano

It's basically an Uno. We should support the Nano :)

Support the Nano

Don't support the Nano

**Additional context**

Once #553 is done this is trivial.

neilenns:
Fixes #579 

This adds support for trying different baud rates when flashing a given board. It is primarily intended to resolve the "which baud rate should we use for a Nano" problem, but is done in a general fashion that could apply to any device that uses avrdude to flash.

Changes include:

1. Version 7.0 of avrdude to get the `-x attempts` command line parameter. (I pulled the x86 version of avrdude). We need this for nano flashing so we can reduce the attempts from `10` to `1`. Otherwise Nano flash failure takes way, way, way too long.
2. New `TimeoutForFirmwareUpdate` property in the `.board.json` file to remove a hard-coded timeout value I missed in the original .board.json implementation
3. New `Attempts` property in the `.board.json` file to control how many times avrdude should attempt connections to a board
4. Change `BaudRate` property to `BaudRates` (an array) in the `.board.json` files
5. Schema updates to reflect the property additions/changes 
6. New `Migrate()` method on the `Board` object to ensure any random old `.board.json` files that might exist will still work
7. Update `RunAvrDude()` to use the `Attempts` property and to loop over the `BaudRates` until one works. This really is the meat of the change to make this all work and is pretty small. 

The baud rates are attempted in the order they are listed in the `.board.json` file which makes it easy for us to prioritize a given baud rate. For the Nano we'll likely want to list `115200` first and `57600` second assuming the new, fast, bootloader is more common.

Changes tested by @elral (both old and new Nanos) and Jamie (new Nano) and me (new Nano).


Boards/arduino_mega.board.json
```diff
@@ -1,8 +1,9 @@
 {
   "$schema": "./mfboard.schema.json",
   "AvrDudeSettings": {
+    "Attempts":  3,
```
neilenns, line 4: avrdude defaults to 10 attempts. If we want to keep exact behaviour with the prior release of MobiFlight we could make this 10 for all non-Nano boards. Seems excessive though, either the connection works or it fails. Setting this to 3 (or 2 or even 1) would make the non-Nano boards fail faster during the flashing process.

Boards/mfboard.schema.json
```diff
@@ -1,24 +1,44 @@
 {
   "definitions": {},
-  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$schema": "https://json-schema.org/draft/2019-09/schema#",
```
neilenns, line 3: Newer release of the JSON Schema to get access to the `deprecated` property.

Boards/mfboard.schema.json
```diff
@@ -176,7 +203,6 @@
         "MaxOutputs",
         "MaxServos",
         "MaxSteppers",
-        "MaxInputShifters",
```
neilenns, line 179: This was a duplicate entry in the schema I cleaned up.

MobiFlight/Board.cs
```diff
@@ -91,6 +101,11 @@ public class Connection
         /// Maximum size of a CmdMessenger message, in bytes.
         /// </summary>
         public int MessageSize;
+
+        /// <summary>
+        /// Number of milliseconds to wait for the firmware update to complete before attempting to call GetInfo on the board.
+        /// </summary>
+        public int TimeoutForFirmwareUpdate = 15000;
```
neilenns, line 108: `15000` is the original hardcoded value so it's used as the default here. If it is omitted from the `.board.json` file the default gets used.

MobiFlight/MobiFlightFirmwareUpdater.cs
```diff
@@ -89,9 +89,6 @@ public static bool UpdateFirmware(MobiFlightModule module, String FirmwareName)
 
         public static void RunAvrDude(String Port, Board board, String FirmwareName) 
         {
-            String ArduinoChip = board.AvrDudeSettings.Device;
-            String Bytes = board.AvrDudeSettings.BaudRate;
-            String C = board.AvrDudeSettings.Programmer;
```
neilenns, line 94: These three lines should have been removed back when I did the .board.json work. They aren't needed and the properies are now just referenced directly in the `anyCommand` string construction.

MobiFlight/MobiFlightFirmwareUpdater.cs
```diff
@@ -107,33 +104,36 @@ public static void RunAvrDude(String Port, Board board, String FirmwareName)
...
-            proc1.UseShellExecute = true;
-            proc1.WorkingDirectory = $@"""{FullAvrDudePath}""";
-            proc1.FileName = $@"""{FullAvrDudePath}\bin\avrdude""";
-            proc1.Arguments = anyCommand;
-            proc1.WindowStyle = ProcessWindowStyle.Hidden;
-            Log.Instance.log($"{proc1.FileName}.", LogSeverity.Debug);
-            Log.Instance.log($"{anyCommand}.", LogSeverity.Debug);
-            Process p = Process.Start(proc1);
-            if (p.WaitForExit(board.AvrDudeSettings.Timeout))
+            foreach (var baudRate in board.AvrDudeSettings.BaudRates)
```
neilenns, line 107: The meat of the change. One additional `foreach` loop to try all the baud rates. So simple!

MobiFlight/MobiFlightFirmwareUpdater.cs
```diff
@@ -107,33 +104,36 @@ public static void RunAvrDude(String Port, Board board, String FirmwareName)
...
-                message = $"avrdude timed out! Something went wrong when flashing with command \n {proc1.FileName} {anyCommand}.";
+                var proc1 = new ProcessStartInfo();
+                string anyCommand =
+                    $@"-C""{FullAvrDudePath}\etc\avrdude.conf"" {verboseLevel} -x attempts={board.AvrDudeSettings.Attempts} -p{board.AvrDudeSettings.Device} -c{board.AvrDudeSettings.Programmer} -P{Port} -b{baudRate} -D -Uflash:w:""{FirmwarePath}\{FirmwareName}"":i";
+                proc1.UseShellExecute = true;
+                proc1.WorkingDirectory = $@"""{FullAvrDudePath}""";
+                proc1.FileName = $@"""{FullAvrDudePath}\bin\avrdude""";
+                proc1.Arguments = anyCommand;
+                proc1.WindowStyle = ProcessWindowStyle.Hidden;
+                Log.Instance.log($"{proc1.FileName} {anyCommand}", LogSeverity.Debug);
```
neilenns, line 117: This used to be two log lines. Combined to one for ease of viewing in the logs.

DocMoebiuz (approved):
Really cool change. Thanks!

DocMoebiuz (changes requested):
I think it would make sense to include the nano board json for which this is all done and then we could also call #579 done.

I also would propose to set all other boards' retry value to 1 because we don't expect that a retry is required. So if it fails, it can fail fast

neilenns:
Json added, updated retries on the other boards.

neilenns:
Once this merges we'll still need to update the firmware builds to build the nano firmware, update the vs project to include it in desktop builds, and I think that's it?

DocMoebiuz:
> Once this merges we'll still need to update the firmware builds to build the nano firmware, update the vs project to include it in desktop builds, and I think that's it?

Yeah, why not do that and then we can include the .hex file right away with this PR? and then this is really the complete nano support.

DocMoebiuz:
i really would like to merge it but without the nano hex file it feels incomplete.

neilenns:
Nano firmware PR: https://github.com/MobiFlight/MobiFlight-FirmwareSource/pull/223

DocMoebiuz (approved):
Mercy boocoo!

PR #1027, opened 2022-12-07
Support proper Honeycomb Button and Axis Labels

DocMoebiuz:
- [x] Buttons have proper names, e.g. AP Mode - HDG
- [x] Axis have proper names
- [x] Backward compatible with loading old configs
- [x] Configs store the internal button name, so that when the label changes, the config will still load correctly
- [ ] Documentation
- [ ] ~~~i18n~~~

## Note
I also added a joystick sub class for the Saitek Aviator Stick (i have it here on my desk conveniently) - axis and buttons can simply be put in the one label file.

### DIFF
--- a/MobiFlight/Joystick.cs
+++ b/MobiFlight/Joystick.cs
@@ -7,6 +7,7 @@ using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using System.Web.SessionState;
+using HidSharp.Utility;
 using SharpDX;
 using SharpDX.DirectInput;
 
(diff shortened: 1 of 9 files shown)


MobiFlight/Joystick.cs
```diff
@@ -115,23 +118,26 @@ private void EnumerateDevices()
...
+                    var FriendlyAxisName = name;
                     try
                     {
                         OffsetAxisName = GetAxisNameForUsage(usage);
+                        FriendlyAxisName = GetFriendlyAxisName(name);
+
                     } catch (ArgumentOutOfRangeException ex)
                     {
-                        Log.Instance.log($"Axis can't be mapped: {joystick.Information.InstanceName} Aspect: {aspect} Offset: {offset} Usage: {usage} Axis: {name}.", LogSeverity.Error);
+                        Log.Instance.log($"Axis can't be mapped: {joystick.Information.InstanceName} Aspect: {aspect} Offset: {offset} Usage: {usage} Axis: {name} Label {FriendlyAxisName}.", LogSeverity.Error);
```
neilenns, line 129:
```suggestion
                        Log.Instance.log($"Axis can't be mapped: {joystick.Information.InstanceName} Aspect: {aspect} Offset: {offset} Usage: {usage} Axis: {name} Label: {FriendlyAxisName}.", LogSeverity.Error);
```

MobiFlight/Joystick.cs
```diff
@@ -115,23 +118,26 @@ private void EnumerateDevices()
...
                         continue;
                     }
-                    Axes.Add(new JoystickDevice() { Name = AxisPrefix + OffsetAxisName, Label = name, Type = JoystickDeviceType.Axis });
-                    Log.Instance.log($"Added {joystick.Information.InstanceName} Aspect {aspect} + Offset: {offset} Usage: {usage} Axis: {name}.", LogSeverity.Debug);
+                    Axes.Add(new JoystickDevice() { Name = AxisPrefix + OffsetAxisName, Label = FriendlyAxisName, Type = JoystickDeviceType.Axis });
+                    Log.Instance.log($"Added {joystick.Information.InstanceName} Aspect {aspect} + Offset: {offset} Usage: {usage} Axis: {name} Label {FriendlyAxisName}.", LogSeverity.Debug);
```
neilenns, line 133:
```suggestion
                    Log.Instance.log($"Added {joystick.Information.InstanceName} Aspect {aspect} + Offset: {offset} Usage: {usage} Axis: {name} Label: {FriendlyAxisName}.", LogSeverity.Debug);
```

MobiFlight/Joysticks/HoneycombBravo.cs
```diff
@@ -43,39 +107,37 @@ protected override void SendData(byte[] data)
...
-            Lights.Add(new JoystickOutputDevice() { Label = "AP Mode - NAV",                Name = "AP.nav",                Byte = 1, Bit = 1 });
-            Lights.Add(new JoystickOutputDevice() { Label = "AP Mode - APR",                Name = "AP.apr",                Byte = 1, Bit = 2 });
-            Lights.Add(new JoystickOutputDevice() { Label = "AP Mode - REV",                Name = "AP.rev",                Byte = 1, Bit = 3 });
-            Lights.Add(new JoystickOutputDevice() { Label = "AP Mode - ALT",                Name = "AP.alt",                Byte = 1, Bit = 4 });
-            Lights.Add(new JoystickOutputDevice() { Label = "AP Mode - VS",                 Name = "AP.vs",                 Byte = 1, Bit = 5 });
-            Lights.Add(new JoystickOutputDevice() { Label = "AP Mode - IAS",                Name = "AP.ias",                Byte = 1, Bit = 6 });
-            Lights.Add(new JoystickOutputDevice() { Label = "AP Mode - On/Off",             Name = "AP.autopilot",          Byte = 1, Bit = 7 });
+            Lights.Add(new JoystickOutputDevice() { Label = "AP Mode - HDG", Name = "AP.hdg", Byte = 1, Bit = 0 });
```
neilenns, line 110: This really should be a loop that goes over an array of objects and adds them... but you don't have to do it with this PR. When I convert to reading from a JSON file the JSON file will have an array of these JoystickOutputDevice objects in it, I'll read it in, and then loop across them to add it.

MobiFlight/Joystick.cs
```diff
@@ -1,10 +1,13 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics.SymbolStore;
```
neilenns, line 3: Why did this get included?

MobiFlight/Joystick.cs
```diff
@@ -1,10 +1,13 @@
...
 using System.Linq;
+using System.Runtime.Remoting.Messaging;
```
neilenns, line 5: Same question for this, is it actually used?

MobiFlight/Joysticks/HoneycombBravo.cs
DocMoebiuz: that's what i also thought!

MobiFlight/Joystick.cs
DocMoebiuz: nice catch

PR #1033, opened 2022-12-08
Support flashing picos

### LINKED ISSUE #1032: Support Raspberry Pi Pico boards
Sure would be nice to support these amazing boards...

neilenns:
Fixes #1032

* Adds new information to board config files to support flashing Picos
* Updates the firmware flashing to do either avrdude for arduinos or USB file copy for picos
* Adds support for disabling board reset on boards that don't support it
* Adds board reset info for Picos, including refactoring where the reset firmware filename comes from in the .board.json file
* Adds detection for USB devices in BOOTSEL mode during initial startup

Known issues:

* If a board is in BOOTSEL mode when mobiflight runs it won't get detected as a COM port after flashing the firmware. Users will need to run MobiFlight again for the board to appear.


neilenns:
Will re-open

PR #1036, opened 2022-12-09
Make Attempts an optional setting

### LINKED ISSUE #1034: Mega firmware flashing is broken in beta 9.6.2
Attempting to flash mega boards using 9.6.2 fails. The error is:

```
avrdude.exe: wiring_parseextparms(): invalid extended parameter 'attempts=1'
```

From the avrdude documentation the `attempts` extended parameter is only valid with the `arduino` programmer. This works fine for all the boards *except* the mega, which uses the `wiring` programmer.

Fix is to make the attempts property optional in the board.json and don't send the extra parameter to avrdude if it isn't specified in the board.json file.

neilenns:
Fixes #1034

* Makes `Attempts` an optional setting in the schema
* Makes `Attempts` a nullable value in the `Board` object
* Update the avrdude command line to only pass `-x attempts=` if the `Attempts` property is set in the board.json

Verified the Mega can now be flashed and the Nano still flashes correctly.

### DIFF
--- a/Boards/arduino_mega.board.json
+++ b/Boards/arduino_mega.board.json
@@ -1,7 +1,6 @@
 {
   "$schema": "./mfboard.schema.json",
   "AvrDudeSettings": {
-    "Attempts":  1,
     "Device": "atmega2560",
     "BaudRates": [ "115200" ],
     "Programmer": "wiring",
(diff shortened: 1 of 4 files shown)


DocMoebiuz (approved):
Cool, thanks for the quick fix.

PR #1037, opened 2022-12-09
Add check for capslock when creating a new input config

### LINKED ISSUE #730: First character in Event Description Field comes in Lower Case
**Describe the bug**
When entering a new event in MF with Caps Lock on, the first letter comes out in lower case.

**Steps To Reproduce**
1. Go to the bottom of MF main list view
2. Have your keyboard with Caps Lock on.
3. Single click on the last line to add a new event (I know it says to double click, but the same occurs with an existing config)
4. Begin typing the name of the config... for example BATTERY MASTER ON
5. Note that the first letter is in lower case for no apparent reason.

image

**Expected behavior**
Caps lock is on, so all letters should be entered as upper case.

**Actual behavior**
The first letter comes up as lower case wit the rest of the text in upper case.

**Details**
 - MobiFlight Version: 9.3.0
 - FlightSim: irrelevant
 - AddOn: irrelevant

**Additional context**
The issue happens only when entering the description without having double clicked first and with Caps lock on.

### LINKED ISSUE #1018: Input config ignores caps lock state when entering the first character into the description field
**Describe the bug**
This bug was fixed several months ago by Neil, 
[https://github.com/MobiFlight/MobiFlight-Connector/issues/730](https://github.com/MobiFlight/MobiFlight-Connector/issues/730)
but somehow has crept back into Mobiflight.

**Steps To Reproduce**
1. Go to Inputs tab
2. Set Caps Lock key to ON
3. Select an input config (without entering the description field)
4. By default text entered opens the description field
5. Enter text into description
6. Note first character is in lower case, the rest being in upper case.

**Expected behavior**
Text entered should respect Caps Lock state.

**Actual behavior**
First character ignores Caps Lock state.

**Details**
 - MobiFlight Version: 9.6.0
 - FlightSim: irrelevant
 - AddOn: irrelevant

**Additional context**
Only affects input, output config behaves as expected.  Bug does not happen if you enter the description field.

neilenns:
Fixes #1018

Easy fix, same as the one I did to fix #730 (PR: #739)

### DIFF
--- a/UI/Panels/InputConfigPanel.cs
+++ b/UI/Panels/InputConfigPanel.cs
@@ -275,7 +275,7 @@ namespace MobiFlight.UI.Panels
                     dgv.BeginEdit(true);
                     if (e.KeyCode != Keys.F2)
                     {
-                        (dgv.EditingControl as TextBox).Text = (e.Shift) ? e.KeyCode.ToString() : e.KeyCode.ToString().ToLower();
+                        (dgv.EditingControl as TextBox).Text = (e.Shift || Control.IsKeyLocked(Keys.CapsLock)) ? e.KeyCode.ToString() : e.KeyCode.ToString().ToLower();
                         (dgv.EditingControl as TextBox).Select(1, 0);
                     }
                 }


MobiFlight-Admin (approved):
Nice!

PR #1043, opened 2022-12-10
Resolve crash with interpolation dialog

### LINKED ISSUE #1042: Floating point interpolation values cause an exception when opening config dialog
**Describe the bug**

If an interpolation value is a floating point number it is possible to get an exception when opening the config dialog.

**Steps To Reproduce**
1. Load the attached .txt file as a .mcc file
2. Open the `LED Backlighting` output

**Expected behavior**
No exception

**Actual behavior**
Exception: 'The given key was not present in the dictionary.' InterpolationPanel.cs:56

**Additional details**

I looked at this in the debugger and it appears to be a rounding issue caused by a type mismatch. `i.GetValues().Keys` is a `Double` and returns `0.987999975681305`. `Key` is a `float` and returns `0.988`.

image

[Pedestal-Connector-Config.txt](https://github.com/MobiFlight/MobiFlight-Connector/files/10200699/Pedestal-Connector-Config.txt)

neilenns:
Fixes #1042 

Change the data type from float to double. No more crashy crash.

### DIFF
--- a/UI/Panels/Config/InterpolationPanel.cs
+++ b/UI/Panels/Config/InterpolationPanel.cs
@@ -49,7 +49,7 @@ namespace MobiFlight.UI.Panels.Config
 
             dt.Rows.Clear();
             
-            foreach (float Key in i.GetValues().Keys)
+            foreach (double Key in i.GetValues().Keys)
             {
                 DataRow row = dt.NewRow();
                 row["Input"] = Key.ToString();


UI/Panels/Config/InterpolationPanel.cs
```diff
@@ -49,7 +49,7 @@ internal void syncFromConfig(Interpolation i)
 
             dt.Rows.Clear();
             
-            foreach (float Key in i.GetValues().Keys)
+            foreach (double Key in i.GetValues().Keys)
```
neilenns, line 52: `Keys` are defined as a `double` on the Interploation object so update the foreach to use the same data type for `Key`

DocMoebiuz: oh! good catch. I wonder if there are other spots where floats are used that used to work when internally a int was used. I will have to search for "float" in the entire project and check

DocMoebiuz:
Thanks!

PR #1049, opened 2022-12-16
Check for active in ExecutionManager:ExecuteConfig()

### LINKED ISSUE #1048: Blink modifier keeps on blinking
**Describe the bug**
Blink modifier in mcc file does not stop blinking with active="False"

**Steps To Reproduce**
1. Set blink modifier properly in mcc file with 
`<blink active="True" blinkValue="0" onOffSequence="500,500" />`
2. Check that blink works
3. Set blink to `active="False"`
4. Note blinking continues to work. 

**Expected behavior**
Blink should stop with `active="False"`

**Actual behavior**
LEDs continue to blink even with `active="False"`

**Details**
 - MobiFlight Beta 9.6.0.2
 - FlightSim: irrelevant
 - AddOn: irrelevant

**Additional context**
Workaround was to set `onOffSequence="0"`

neilenns:
Fixes #1048 

Instead of adding a test for `Active` inside every `Apply()` method of every current and future modifier, move the test into `ExecuteConfig()` and just pre-filter out any modifier that doesn't have `Active` set to `true`.

Verified this works when `<modifiers>` is empty inside the config file. Verified it picks up active and inactive transforms, comparisons, and blinks.

### DIFF
--- a/MobiFlight/ExecutionManager.cs
+++ b/MobiFlight/ExecutionManager.cs
@@ -485,7 +485,7 @@ namespace MobiFlight
                 try
                 {
                     processedValue = value;
-                    foreach (var modifier in cfg.Modifiers.Items)
+                    foreach (var modifier in cfg.Modifiers.Items.Where(m => m.Active))
                     {
                         processedValue = modifier.Apply(processedValue, configRefs);
                     }
--- a/MobiFlight/Modifier/Comparison.cs
+++ b/MobiFlight/Modifier/Comparison.cs
@@ -164,11 +164,6 @@ namespace MobiFlight.Modifier
             string result = connectorValue.String;
             string value = connectorValue.String;
 
-            if (!Active)
-            {
-                return connectorValue.String;
-            }
-
             string comparisonValue = Value;
             string comparisonIfValue = IfValue;
             string comparisonElseValue = ElseValue;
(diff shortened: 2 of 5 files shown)


MobiFlight/Modifier/Comparison.cs
```diff
@@ -169,11 +164,6 @@ private string ExecuteStringComparison(ConnectorValue connectorValue)
             string result = connectorValue.String;
             string value = connectorValue.String;
 
-            if (!Active)
-            {
-                return connectorValue.String;
-            }
-
```
neilenns, line 176: Even though this isn't strictly related to the `Apply()` bug, I removed it anyway for consistency. The only place this method gets called in the entire codebase is from `Comparison.cs:Apply()` so it never had any real effect anyway.

MobiFlight/ExecutionManager.cs
```diff
@@ -485,7 +485,7 @@ private void ExecuteConfig()
                 try
                 {
                     processedValue = value;
-                    foreach (var modifier in cfg.Modifiers.Items)
+                    foreach (var modifier in cfg.Modifiers.Items.Where(m => m.Active))
```
neilenns, line 488: The meat of the fix. I love LINQ.

DocMoebiuz:
Is the order of the modifiers maintained?

Another option which would be more analog to other areas of the code is to simply add

```
if (!modifier.Active) continue;
```

DocMoebiuz (commented):
see my last comment, https://github.com/MobiFlight/MobiFlight-Connector/pull/1049#issuecomment-1365455230

neilenns:
Yes the order is maintained

DocMoebiuz (approved):
TY!

PR #1057, opened 2022-12-20
Support flashing Picos

### LINKED ISSUE #1032: Support Raspberry Pi Pico boards

neilenns:
Fixes #1032

* Adds new information to board config files to support flashing Picos
* Updates the firmware flashing to do either avrdude for arduinos or USB file copy for picos
* Adds support for disabling board reset on boards that don't support it
* Adds board reset info for Picos, including refactoring where the reset firmware filename comes from in the .board.json file
* Adds detection for USB devices in BOOTSEL mode during initial startup
* Renames some variables to remove references to "Arduino" when it is really about COM port devices
* Removed "Arduino" from some of the UI dialogs so the messages are more generic

Known issues:

* If a board is in BOOTSEL mode when mobiflight runs it won't get detected as a COM port after flashing the firmware. Users will need to run MobiFlight again for the board to appear.


Boards/arduino_mega.board.json
```diff
@@ -5,7 +5,6 @@
     "BaudRates": [ "115200" ],
     "Programmer": "wiring",
     "FirmwareBaseName": "mobiflight_mega",
-    "ResetFirmwareFile": "reset.arduino_mega_1_0_2.hex",
```
neilenns, line 8: This property got moved to the `Info` object so it can be shared by both COM and USB-based boards.

Boards/arduino_mega.board.json
```diff
@@ -35,10 +34,12 @@
   ],
   "Info": {
     "CanInstallFirmware": true,
+    "CanResetBoard": true,
```
neilenns, line 37: This new property controls whether the reset board menu option is available for the device. Originally the Pico couldn't be reset so I added this in. Pico can be reset now but it seemed worth leaving this in for future use.

Boards/mfboard.schema.json
```diff
@@ -319,6 +332,39 @@
           }
         }
       }
+    },
+    "UsbDriveSettings": {
```
neilenns, line 336: The new settings for USB flashing.

MobiFlight/MobiFlightCache.cs
```diff
@@ -45,7 +45,7 @@ public class MobiFlightCache : MobiFlightCacheInterface, ModuleCacheInterface
         public event EventHandler LookupFinished;
 
 
-        private List<MobiFlightModuleInfo> connectedArduinoModules = null;
+        private List<MobiFlightModuleInfo> connectedComModules = null;
```
neilenns, line 48: Renamed since this will now contain Picos as well.

MobiFlight/MobiFlightCache.cs
```diff
@@ -197,7 +197,7 @@ private static List<PortDetails> getSupportedPorts()
             return result;
         }
 
-        private async  Task<List<MobiFlightModuleInfo>> LookupAllConnectedArduinoModulesAsync()
+        private async  Task<List<MobiFlightModuleInfo>> LookupAllConnectedComModulesAsync()
```
neilenns, line 200: Renamed since this will now find Picos as well.

MobiFlight/MobiFlightFirmwareUpdater.cs
```diff
@@ -34,66 +34,89 @@ public static bool IsValidFirmwareFilepath(string filepath)
...
                 Log.Instance.log($"Firmware reset requested for {module.Board.Info.MobiFlightType} ({module.Port}) however no reset settings were specified in the board definition file. Module reset skipped.", LogSeverity.Error);
                 return false;
             }
 
-            var FirmwareName = module.Board.AvrDudeSettings.ResetFirmwareFile;
-            return UpdateFirmware(module, FirmwareName);
+            return UpdateFirmware(module, module.Board.Info.ResetFirmwareFile);
         }
 
         public static bool UpdateFirmware(MobiFlightModule module, String FirmwareName)
```
neilenns, line 67: Changes to this method are best viewed in side-by-side mode instead of inline mode.

MobiFlight/MobiFlightFirmwareUpdater.cs
```diff
@@ -34,66 +34,89 @@ public static bool IsValidFirmwareFilepath(string filepath)
...
+            catch
             {
-                Log.Instance.log($"Firmware update requested for {module.Board.Info.MobiFlightType} ({module.Port}) however no update settings were specified in the board definition file. Module update skipped.", LogSeverity.Error);
+                result = false;
             }
             return result;
         }
 
-        public static void RunAvrDude(String Port, Board board, String FirmwareName) 
+        public static void RunAvrDude(String port, Board board, String firmwareName)
```
neilenns, line 113: Changed capitalization for consistency with other methods.

MobiFlight/MobiFlightFirmwareUpdater.cs
```diff
@@ -137,5 +160,85 @@ public static void RunAvrDude(String Port, Board board, String FirmwareName)
 
             throw new Exception(message);
         }
+
+        public static void FlashViaUsbDrive(String port, Board board, String firmwareName)
```
neilenns, line 164: The work to do the actual flashing. It's a lotta lines of code to copy a file 😂

UI/MainForm.cs
```diff
@@ -332,6 +333,37 @@ private void CheckForWasmModuleUpdate()
             
         }
 
+        /// <summary>
+        /// Returns a list of connected USB drives that are supported with MobiFlight and are in flash mode already,
+        /// as opposed to being connected as COM port.
+        /// </summary>
+        /// <returns>The list of connected USB drives supported by MobiFlight.</returns>
+        private static List<MobiFlightModuleInfo> FindConnectedUsbDevices()
```
neilenns, line 341: This is the method used to find devices that are already in BOOTSEL mode when MobiFlight runs. Seems very strange to have it in a `MainForm.cs` file but I didn't know anywhere else better to put it.

DocMoebiuz: A better place is MobiflightCache because it contains a list of all modules.

DocMoebiuz (changes requested):
Please check my comments.
Great job with all the changes. In a next step I will refactor this process so that we have AvrDude and PicoFlash as two implementations of the firmware update as two individual strategies. That might help organizing the code.

A similar approach will make sense for "detecting" boards. Instead of having that all in one class, we can separate that and raise events whenever one of the classes detects a board. But I will now leave this idea out of the scope of this PR.

Boards/arduino_mega.board.json
neilenns: I moved `FirmwareBaseName` and `FirmwareExtension` into the `Info` object since it was already there and meant less disruption to the schema. There was already firmware-related stuff there too, like the latest firmware version.

UI/MainForm.cs
neilenns: Moved to MobiFlightCache.cs as a static method.

neilenns:
@DocMoebiuz Feedback addressed, thanks for the comments. I agree with your ideas around a future refactor.

DocMoebiuz (approved):

elral:
@neilenns @DocMoebiuz seems that my comment (see above) reageding the max. devices didn't find the way into the main branch ;)

neilenns:
@elral I don't see any comments from you on this PR?

elral:
@neilenns  Sorry, seems that I did again something wrong on Github.
The .json file has the max devices from the mega.

neilenns:
@elral Open a bug and PR with the corrected file 😀 Does it really matter? It's not like the Pico hurts for memory.

elral:
What could I have done wrong?
image
This is what I see...

neilenns:
You didn't hit request changes and complete review I think. See how it says "Pending"?

PR #1079, opened 2023-01-20
Add ?. to handle possible null object

### LINKED ISSUE #1078: Log message for device not ready might still cause a crash
The log message added in #1075 might itself cause a crash if for some reason the `device` object is null.

neilenns:
Fixes #1078 

Added ?. to the two object dereferences in the log message just to be safe.

### DIFF
--- a/MobiFlight/MobiFlightCache.cs
+++ b/MobiFlight/MobiFlightCache.cs
@@ -98,9 +98,7 @@ namespace MobiFlight
                 }
                 catch (Exception ex)
                 {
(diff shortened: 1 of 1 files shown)


DocMoebiuz (changes requested):
I want to suggest to add quotes around name and label so that it is easier to spot when the values are actually null... quotes or brackets make it obvious in the log message that there should be some value displayed

neilenns:
This isn't necessary now that I think about it. `drive.IsReady` would also throw an exception, and the MSDN sample doesn't protect against it either.

PR #1083, opened 2023-01-20
Update check happens first thing after startup

### LINKED ISSUE #1081: Update check should happen first thing after startup
Right now we do a bunch of things, like scanning for connected devices, before running the update check. This means if someone (not me, I swear) introduces a bug that causes a crash *before* the update check happens (not me, I swear!!!) we can't push out an auto-update to fix it.

Sebastian suggested in Discord that we move the update check earlier in the flow. This issue tracks doing that.

Pitj3:
Moved the update check to when the app is shown instead of after the modules have been loaded. This will allow us to fix bugs introduced easier with a new update.

There a few other locations we could check this for, personally I think this is the most straight forward place to do it.

fixes #1081

### DIFF
--- a/UI/MainForm.cs
+++ b/UI/MainForm.cs
@@ -120,6 +120,11 @@ namespace MobiFlight.UI
 
         private void MainForm_Shown(object sender, EventArgs e)
         {
+            // Check for updates before loading anything else
+#if (!DEBUG)
+            AutoUpdateChecker.CheckForUpdate(false, true);
+#endif
+
             if (Properties.Settings.Default.Started == 0)
             {
                 OnFirstStart();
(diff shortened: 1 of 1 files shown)


DocMoebiuz:
> There a few other locations we could check this for, personally I think this is the most straight forward place to do it.

What are you referring to? What did you have in mind?

DocMoebiuz:
```
public MainForm()
        {
            // this shall happen before anything else
            InitializeUILanguage();
            // then initialize components
            InitializeComponent();
            // then restore settings
            InitializeSettings();
            // finally set up logging (based on settings)
            InitializeLogging();
            // Initialize the board configurations
            BoardDefinitions.Load();
            // configure tracking correctly
            InitializeTracking();
        }
```

yes, there is a bunch of stuff happening before that which could also still break, Boarddefinitions or so... but it's been fine for a long time (until @neilenns broke something) - so I am fine with the quick win.

neilenns:
To be fair the board definition stuff is littered in try/catch blocks, and has survived busted board config files before 😂

Pitj3:
I think this is sufficiently "early" enough in the application startup where if it still breaks we would notice before pushing the update out to people.

DocMoebiuz (approved):
let's give it a try

Issue #1110, opened 2023-02-01
Refactor the initialization of device pins in Config objects


DocMoebiuz:
**Is your feature request related to a problem? Please describe.**
I reviewed the code of how the different devices are intialized while reviewing the issue related to #1106 and I came to the conclusion that there is some opportunities to refactor code

**Describe the solution you'd like**
- Devices are not initialized with negative pin numbers
- Device are initialized with a state that it is clear that they are not set yet
- The initialization code from the MultiplexerDevice shall be reviewed, maybe some lines can be removed
- OutputShifter, InputShifter and Multiplexer have a lot of copied code
- Review using strings for pins and places where we do Parsing of pins to numbers

**Describe alternatives you've considered**
Don't refactor and leave things as they are.

PR #1119, opened 2023-02-04
Fixed handling of MultiplexDriver deregistration

### LINKED ISSUE #1105: Crash when initializing a multiplexer
**Describe the bug**
User in Discord reported a crash when attempting to use a multiplexer in version 9.6.0.

```
System.OverflowException: Value was either too large or too small for an unsigned byte.
   at System.Byte.Parse(String s, NumberStyles style, NumberFormatInfo info)
   at MobiFlight.Config.MultiplexerDriver.Initialize(String[] pinNos)
   at MobiFlight.Config.MultiplexerDriver.FromInternal(String value)
```

image

The offending line of code is here:

https://github.dev/MobiFlight/MobiFlight-Connector/blob/b4cf773cef7f5c48de6fb215470f1b299ae3bc44/MobiFlight/Config/MultiplexerDriver.cs#L56-L57

Still not sure *why* this is happening but the code can definitely be tweaked to not crash.

GioCC:
Fixes #1105
Modified:
- Fixed registration calls/reg. counter updates in MultiplexerDriver
- Modified functions to lookup an existing MultiplexDriver in MobiFlightPanel; changed their names to reflect the modified purpose
- Added or corrected a few comments about Multiplexer Driver handling

The first commit contains only code changes; the second commits contains only comments, so it should be easier to review.

### DIFF
--- a/MobiFlight/Config/Config.cs
+++ b/MobiFlight/Config/Config.cs
@@ -157,9 +157,11 @@ namespace MobiFlight.Config
                                 // the FromInternal() of the client (in this case InputMultiplexer) will provide them.
                                 // The MultiplexerDriver is registered as a "ghost" device in Config's items list; it won't be shown in the GUI tree.
                                 Items.Add(multiplexerDriver);
-                            } else {
-                                multiplexerDriver.registerClient();
+                            //} else {
+                            //    multiplexerDriver.registerClient();
                             }
+                            multiplexerDriver.FromInternal(InputMultiplexer.GetMultiplexerDriverConfig(item + BaseDevice.End));
+
                             currentItem = new MobiFlight.Config.InputMultiplexer(multiplexerDriver);
                             currentItem.FromInternal(item + BaseDevice.End);
                             break;
--- a/MobiFlight/Config/Config.cs
+++ b/MobiFlight/Config/Config.cs
@@ -155,6 +155,8 @@ namespace MobiFlight.Config
                                 // (in this case InputMultiplexer) will provide them
                                 // Treat the MultiplexerDriver as a regular device (add it to the items list), except it won't be shown in the GUI tree.
                                 Items.Add(multiplexerDriver);
+                            } else {
+                                multiplexerDriver.registerClient();
                             }
                             currentItem = new MobiFlight.Config.InputMultiplexer(multiplexerDriver);
                             currentItem.FromInternal(item + BaseDevice.End);
--- a/MobiFlight/Config/Config.cs
+++ b/MobiFlight/Config/Config.cs
@@ -61,7 +61,7 @@ namespace MobiFlight.Config
         {
             String[] items = value.Split(BaseDevice.End);
 
-            // Need to set aside the MultiplexerDriver reference (for subsequent devices) when we find it 
+            // Need to set aside the MultiplexerDriver reference (for subsequent client devices) when we find it 
             MobiFlight.Config.MultiplexerDriver multiplexerDriver = null;
 
             foreach (String item in items)
@@ -148,12 +148,14 @@ namespace MobiFlight.Config
                             break;
 
                         case DeviceType.InputMultiplexer:
+                            // If the multiplexerDriver is to be implicitly defined by clients:
                             // Build multiplexerDriver if none found yet 
                             if (multiplexerDriver == null) {
+                                // Store it, so another clients will not create a new one
                                 multiplexerDriver = new MobiFlight.Config.MultiplexerDriver();
-                                // multiplexerDriver is not yet init'ed with pin numbers: the FromInternal() of the client
-                                // (in this case InputMultiplexer) will provide them
-                                // Treat the MultiplexerDriver as a regular device (add it to the items list), except it won't be shown in the GUI tree.
+                                // multiplexerDriver is not yet initialized with pin numbers:
+                                // the FromInternal() of the client (in this case InputMultiplexer) will provide them.
+                                // The MultiplexerDriver is registered as a "ghost" device in Config's items list; it won't be shown in the GUI tree.
                                 Items.Add(multiplexerDriver);
                             } else {
                                 multiplexerDriver.registerClient();
@@ -172,6 +174,18 @@ namespace MobiFlight.Config
                             currentItem.FromInternal(item + BaseDevice.End);
                             break;
 
+                        // If the multiplexerDriver is to be explicitly defined by its own config line,
+                        // following 'case' is required:
+
+                        //case DeviceType.MultiplexerDriver:
+                            //if (multiplexerDriver == null) {
+                            //  multiplexerDriver = new MobiFlight.Config.MultiplexerDriver();
+                            //  currentItem = multiplexerDriver;
+                            //  currentItem.FromInternal(item + BaseDevice.End);
+	                        //} else {
+	                        //  multiplexerDriver.registerClient();
+                            //}
+                            //break;
                     }
 
                     if (currentItem != null)
--- a/MobiFlight/Config/Config.cs
+++ b/MobiFlight/Config/Config.cs
@@ -153,8 +153,6 @@ namespace MobiFlight.Config
                             if (multiplexerDriver == null) {
                                 // Store it, so another clients will not create a new one
                                 multiplexerDriver = new MobiFlight.Config.MultiplexerDriver();
-                                // multiplexerDriver is not yet initialized with pin numbers:
-                                // the FromInternal() of the client (in this case InputMultiplexer) will provide them.
                                 // The MultiplexerDriver is registered as a "ghost" device in Config's items list; it won't be shown in the GUI tree.
                                 Items.Add(multiplexerDriver);
                             //} else {
--- a/MobiFlight/Config/MultiplexerDriver.cs
+++ b/MobiFlight/Config/MultiplexerDriver.cs
@@ -31,12 +31,8 @@ namespace MobiFlight.Config
 
         public bool registerClient()
         {
-            bool res = false;
-            if (_initCounter > 0) {
-                _initCounter++;
-                res = true;
-            }
-            return res;
+            _initCounter++;
+            return true;
         }
 
         public bool unregisterClient()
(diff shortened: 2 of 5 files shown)


MobiFlight/Config/Config.cs
```diff
@@ -155,6 +155,8 @@ public Config FromInternal(String value, bool throwException = false)
                                 // (in this case InputMultiplexer) will provide them
                                 // Treat the MultiplexerDriver as a regular device (add it to the items list), except it won't be shown in the GUI tree.
                                 Items.Add(multiplexerDriver);
+                            } else {
```
DocMoebiuz, line 160: I would like to keep the symmetry here with calling .FromInternal() for the MultiplexerDriver.
What stands out to me is that we interact with the `registerClient` method from the outside (here), and the inside (inside FromInternal)

``` C#
// Build multiplexerDriver if none found yet 
if (multiplexerDriver == null) {
    multiplexerDriver = new MobiFlight.Config.MultiplexerDriver();
    Items.Add(multiplexerDriver);
}
// then we can always call 
multiplexerDriver.FromInternal(item + BaseDevice.End);
```

Inside from internal the "registerClient" is called, and the pins can be set, if required (we can do it once, or we can do it every time, it doesn't matter if the config was created through mobiflight the base pins will be the same)

MobiFlight/Config/MultiplexerDriver.cs
```diff
@@ -31,12 +31,8 @@ public bool isInitialized()
 
         public bool registerClient()
         {
-            bool res = false;
-            if (_initCounter > 0) {
-                _initCounter++;
-                res = true;
-            }
-            return res;
+            _initCounter++;
```
DocMoebiuz, line 34: This should be a `private` message because this is "inner" state that no one should worry about outside.

MobiFlight/Config/MultiplexerDriver.cs
```diff
@@ -31,12 +31,8 @@ public bool isInitialized()
...
+            return true;
         }
 
         public bool unregisterClient()
```
DocMoebiuz, line 38: This should be a `private` message because this is "inner" state that no one should worry about outside.

MobiFlight/Config/MultiplexerDriver.cs
```diff
@@ -109,10 +103,9 @@ override public bool FromInternal(String value)
                 pins[1] = paramList[2];
                 pins[2] = paramList[3];
                 pins[3] = paramList[4];
-                res = Initialize(pins);
-            } else {
-                _initCounter++;
+                res = SetPins(pins);
```
DocMoebiuz, line 106: Why not setting the paramList result to the PinSx?

We don't have a logic that checks the pin values coming from the board in any other device either since we can assume that the configs are correct.

We an entirely remove SetPins(pins);

UI/Panels/Settings/MobiFlightPanel.cs
```diff
@@ -335,7 +335,7 @@ private TreeNode mfModulesTreeView_initNode(MobiFlightModuleInfo moduleInfo, Tre
                     // MultiplexerDrivers should not appear in the tree, therefore they are stored in a dictionary 
                     // (by module name) for easy retrieval
                     if(device.Type == DeviceType.MultiplexerDriver) {
-                        moduleMultiplexerDrivers.Add(moduleInfo.Name, device as MobiFlight.Config.MultiplexerDriver);
+                        moduleMultiplexerDrivers.Add(moduleNode.Text, device as MobiFlight.Config.MultiplexerDriver);
```
DocMoebiuz, line 339: What is wrong with moduleInfo.Name?
I know both are the same at this point. What was the intention to change it?

MobiFlight/Config/MultiplexerDriver.cs
```diff
@@ -109,10 +103,9 @@ override public bool FromInternal(String value)
...
             }
+            registerClient();
```
DocMoebiuz, line 108: I would say, we make this either called from the outside (not my favorite) or only from here (my favorite)

UI/Panels/Settings/MobiFlightPanel.cs
```diff
@@ -1081,10 +1067,12 @@ private void unregisterMuxClient()
         {
             MobiFlight.Config.MultiplexerDriver multiplexerDriver;
             
-            multiplexerDriver = getModuleMultiplexerDriver();
-            multiplexerDriver.unregisterClient();
-            if (!multiplexerDriver.isInitialized()) {
-                moduleMultiplexerDrivers.Remove(getModuleNode().Name);
+            multiplexerDriver = findModuleMultiplexerDriver(false);
```
DocMoebiuz, line 1083: In general I would propose to rethink about the approach to handle a lot of this MultiplexerDriver logic in the View component.

I would like to handle the unregisterClient Logic inside the Config object.

It seems to me that there is too much knowledge of the Driver required here.

DocMoebiuz (changes requested):
it is hard for me to tell what actual situation or state we are really addressing with all the changes made. How critical is this PR? We have had Multiplexer for a while.

I had created this issue recently: https://github.com/MobiFlight/MobiFlight-Connector/issues/1110

And i think some parts are touching this logic too. 

In any case, please review my other comments.

GioCC:
> it is hard for me to tell what actual situation or state we are really addressing with all the changes made. How critical is this PR? We have had Multiplexer for a while.

As stated in the PR text, issue #1105 (which is incorrectly marked as resolved with a patch that gets rid of the crash but leaves the issue). As for how critical, it's a bug that consistently appears if multiplexer inputs ared added/deleted/readded.

> I had created this issue recently: #1110
> And i think some parts are touching this logic too.

Probably yes, since they refer to the same part of the code. However, this is not the issue I meant to address (nor I even tried to).
 
> In any case, please review my other comments.

MobiFlight/Config/Config.cs
GioCC: This would definitely work; calling `MultiplexerDriver.FromInternal()` from here (rather than from `WhateverClient.FromInternal()`) would apparently simplify the code.

The only drawback is that the client should have a method to supply the part of its config string to be passed to `multiplexerDriver.FromInternal()` (in the correct format); therefore I had to add a `GetMultiplexerDriverConfig()` method, any other suggestion welcome.

Unfortunately, this now breaks the symmetry with the creation of the corresponding objects in `MobiFlightPanel`, where the call to `registerClient()` is explicit (there are no methods to hide it in). 

On a related note, I still stand by my opinion that having the Multiplexer driver defined in its own specific configuration line only would be a better choice.
The current situation, besides being dangerous because of the redundant definition of the same object in several places (which is required and assumed - without any enforcement or check in place - to be consistent), also brings to additional needless clutter in the configuration file, which will only get worse if further mux client devices are ever implemented. As for the advantages, I fail to see any, but I'd be glad to see it proven otherwise.
(BTW, the above mentioned `GetMultiplexerDriverConfig()` method would also become unnecessary - another simplification.)

MobiFlight/Config/MultiplexerDriver.cs
GioCC: Could you explain why should this be "inner"? (The variable obviously is.)
Un/Registering a new client for the driver is essential to keep count when - in the process of adding/deleting devices - the driver can be invalidated and its pins released; this is obviously an operation that has to be exposed.

MobiFlight/Config/MultiplexerDriver.cs
GioCC: > Why not setting the paramList result to the PinSx?

They are shifted by one; once shifted, we already have `SetPins()` available. If the issue is removing the validation check, see next comment;

> We don't have a logic that checks the pin values coming from the board in any other device either since we can assume that the configs are correct.

That's true - but that is also the reason why this bug could be caught in the first place. Maybe it's the other way around, checks could be implemented also in other places given the occasion.
 
> We an entirely remove SetPins(pins);

Do you mean the call or the method? Because it is used when initializing a new driver in the GUI.

UI/Panels/Settings/MobiFlightPanel.cs
GioCC: All other properties here are set using `moduleNode`; exactly because `moduleNode.Text` is the same as `moduleInfo.Name` makes it more logical and less misleading to use it.

UI/Panels/Settings/MobiFlightPanel.cs
GioCC: Could you elaborate? If I'm not mistaken, the saved/uploaded configuration is sourced from the View component before the corresponding Config object exists; and this latter is rebuilt by reading back the board config. Therefore the driver is created (or deleted) first in the GUI.
In that place, it is required that each new client un/registers itself for the reasons mentioned above; I can't see any useless operation there.
Could you make an example of what the `getModuleMultiplexerDriver()` should do differently (or how it could possibly be removed)?

MobiFlight/Config/MultiplexerDriver.cs
GioCC: As mentioned in other replies (https://github.com/MobiFlight/MobiFlight-Connector/pull/1119#discussion_r1097854556), I find it hard to imagine how we could get rid of the explicit call (from the outside) in the GUI; and for consistency I'd also be inclined to "extract" it from here (in the only remaining call from `Config`.)

UI/Panels/Settings/MobiFlightPanel.cs
DocMoebiuz: let's schedule an interactive coding session and try some stuff. it costs a lot of time to discuss this here

GioCC: Agreed, I'd be very happy to. I'll PM you for arrangements.

GioCC:
As I mentioned in a PM, After all, this PR fixes an obvious bug, and it does not change multiplexer handling from the way it already is now.
Therefore, in this sense I see no reason to defer it until a possible refactoring is discussed, agreed and implemented; in the meantime, the fix has to be resynced with the evolving main branch with the danger that it introduces new bugs.

GioCC:
@DocMoebiuz Too much time (and too many commits) have passed since this fix; I'm having a hard time keeping all changes correctly merged.
I'll make one last try, if unsuccessful I'll delete the PR and leave the bug for somebody other to fix.

PR #1121, opened 2023-02-05
Dynamically load labeled joysticks from config files

### LINKED ISSUE #992: Support defining joysticks via json files
Right now joysticks get detected but with generic input names. We're about to add support for specific input names for the Honeycomb Bravo, but what about all the other joysticks?

Provide a way to specify joystick information via json files that are loaded at startup so additional joysticks can be defined with friendly names for inputs/outputs without having to recompile Mobiflight.

This is too cool to consider alternatives.

**Additional context**
This would be similar to how Boards are handled. Make sure to support joysticks that don't have a customized definition file too.

neilenns:
Fixes #992 

This adds support for dynamically loading LabeledJoysticks from JSON files. By adding a JSON file to the `HIDs` folder it will magically get labeled inputs and support outputs.

Changes made:

* Add a `HidDefinitions` class to manage loading the definition files and finding a matching definition based on the device's instance name.
* Add a `HidDefinition` class that contains the information about a labeled joystick.
* Update `LabeledJoystick` to take in a `HidDefinition` as part of the constructor and use that to populate labels
* Move the code from `HoneycombBravo.cs` that connected to the device to send output information, and to enumerate output devices, over to `LabeledJoystick`
* Update `JoystickManager` to load the definition files, search for a matching definition, and create a `LabeledJoystick` if one was found


neilenns:
So at this point I *think* the custom Saitek class could just go away, since all it does is load the labels. That can move into the base Joystick class.

What in the Bravo class is special now that couldn't be moved to the base class too @DocMoebiuz ?

MobiFlight/JoystickManager.cs
```diff
@@ -11,13 +11,14 @@ public class JoystickManager
     {
         public event EventHandler Connected;
         public event ButtonEventHandler OnButtonPressed;
-        Timer PollTimer = new Timer();
-        List<Joystick> joysticks = new List<Joystick>();
+        readonly Timer PollTimer = new Timer();
+        readonly List<Joystick> joysticks = new List<Joystick>();
```
neilenns, line 15: Cleaning up code stuff suggested by Visual Studio when I was editing the file.

MobiFlight/JoystickManager.cs
```diff
@@ -11,13 +11,14 @@ public class JoystickManager
...
 
         public JoystickManager ()
         {
             PollTimer.Interval = 50;
             PollTimer.Tick += PollTimer_Tick;
+            HidDefinitions.Load();
```
neilenns, line 21: Load all the joystick definitions when JoystickManager is created.

MobiFlight/JoystickManager.cs
```diff
@@ -34,7 +35,7 @@ private void PollTimer_Tick(object sender, EventArgs e)
                         js?.Update();
                     }
                 }
-            } catch (InvalidOperationException ex)
+            } catch (InvalidOperationException)
             {
```
neilenns, line 39: Cleaning up code stuff suggested by Visual Studio when I was editing the file.

MobiFlight/JoystickManager.cs
```diff
@@ -77,12 +78,14 @@ public void Connect(IntPtr Handle)
...
-                } else if (d.InstanceName == "Saitek Aviator Stick")
+
+                // Look up the custom device definition file for the joystick. If it exists then a LabeledJoystick
+                // can get created with nicely named inputs and outputs.
+                var definition = HidDefinitions.GetHidByInstanceName(d.InstanceName);
+
+                if (definition != null)
                 {
-                    js = new Joysticks.SaitekAviatorStick(new SharpDX.DirectInput.Joystick(di, d.InstanceGuid));
+                    js = new Joysticks.LabeledJoystick(new SharpDX.DirectInput.Joystick(di, d.InstanceGuid), definition);
```
neilenns, line 88: The meat of the change in JoystickManager. Look up a matching board definition by instance name and if it's found use that to create a `LabeledJoystick` object. If no definition is found then fall back to making a generic `Joystick`.

MobiFlight/Joysticks/LabeledJoystick.cs
```diff
@@ -1,14 +1,45 @@
...
+using System;
 using System.Collections.Generic;
 using System.Linq;
 
 namespace MobiFlight.Joysticks
 {
     public class LabeledJoystick : Joystick
     {
+        HidStream Stream { get; set; }
+        HidDevice Device { get; set; }
```
neilenns, line 11: This moved here from the old HoneycombBravo.cs file.

MobiFlight/Joysticks/LabeledJoystick.cs
```diff
@@ -1,14 +1,45 @@
...
+
         protected Dictionary<string, string> Labels = new Dictionary<string, string>();
+        readonly public HidDefinition Definition;
+
+        public LabeledJoystick(SharpDX.DirectInput.Joystick joystick, HidDefinition definition) : base(joystick) 
+        {
+            Definition = definition;
+            Definition?.Inputs?.ForEach(input => Labels.Add(input.Name, input.Label));
```
neilenns, line 19: Populate the input labels here instead of in the prior derived classes.

MobiFlight/Joysticks/LabeledJoystick.cs
```diff
@@ -1,14 +1,45 @@
...
+        protected override void SendData(byte[] data)
+        {
+            if (!RequiresOutputUpdate) return;
+            if (Stream == null)
+            {
+                Connect();
+            };
+            Stream.SetFeature(data);
+            base.SendData(data);
+        }
```
neilenns, line 42: Moved into this class from the old HoneycombBravo class.

MobiFlight/Joysticks/LabeledJoystick.cs
```diff
@@ -18,10 +49,15 @@ protected override List<JoystickDevice> GetButtonsSorted()
             return buttons;
         }
 
-        public override string MapDeviceNameToLabel(string name)
+        protected override void EnumerateOutputDevices()
         {
-            var result = name;
+            base.EnumerateOutputDevices();
+            Definition?.Outputs?.ForEach(output => Lights.Add(new JoystickOutputDevice() { Label = output.Label, Name = output.Name, Byte = output.Byte, Bit = output.Bit }));
+        }
```
neilenns, line 56: Moved into this class from the old HoneycombBravo class. Create all the lights by enumerating the outputs in the definition file.

neilenns:
I have a lot more refactoring I came up with that was simplify this further. Will reopen.

PR #1126, opened 2023-02-06
Add exception handling when querying WMI

### LINKED ISSUE #1122: System.Management.ManagementException trying to get connected ports
Reported by a user in Discord. We crash here: https://github.dev/MobiFlight/MobiFlight-Connector/blob/1fa9794cb2d716bf8e5b38e26a3b7301188da88d/MobiFlight/MobiFlightCache.cs#L172-L173

Crash log:

```
System.Management.ManagementException: Invalid namespace 
   at System.Management.ThreadDispatch.Start()
   at System.Management.ManagementScope.Initialize()
   at System.Management.ManagementObjectSearcher.Initialize()
   at System.Management.ManagementObjectSearcher.Get()
   at MobiFlight.MobiFlightCache.getSupportedPorts()
   at MobiFlight.MobiFlightCache.<LookupAllConnectedComModulesAsync>d__31.MoveNext()
```

Reading around online it appears this happens when the WMI registry is somehow damaged or corrupted. There doesn't appear to be anything we can do other than catch the exception and move on. That would mean on this user's machine no COM devices would get detected though.

neilenns:
Fixes #1122

Adds a try/catch for the specific WMI error reported in this bug, as well as a more general try/catch in case something else goes wrong.

### DIFF
--- a/MobiFlight/MobiFlightCache.cs
+++ b/MobiFlight/MobiFlightCache.cs
@@ -168,81 +168,94 @@ namespace MobiFlight
             var regex = new Regex(@"(?<id>VID_\S*)"); // Pattern to match the VID/PID of the connected devices
 
             // Code from https://stackoverflow.com/questions/45165299/wmi-get-list-of-all-serial-com-ports-including-virtual-ports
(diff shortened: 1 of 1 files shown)


DocMoebiuz (approved):
Thanks

PR #1132, opened 2023-02-08
Allow for device names (e.g. joysticks) to contain "/"

### LINKED ISSUE #1131: "Not Implemented" error when trying to use MFG Crosswind pedals
Create new input config line, select MFG Crosswind pedals from Module list.

You get the following dialog and the device list remains empty.

image

Log shows: Error    02/07/2023 19:49:11    Index was outside the bounds of the array.

DocMoebiuz:
fixes #1131 

The problem was the "/" in the device name. So I made sure that we always cut off at the last slash.
I also replaced all code snippets that were not using the `SerialNumber` class and the methods it provides yet.

- [x] Add unit tests for the problem cases
- [x] All unit tests are still working

### DIFF
--- a/Base/SerialNumber.cs
+++ b/Base/SerialNumber.cs
@@ -34,12 +34,19 @@ namespace MobiFlight.Base
 
         public static bool IsMobiFlightSerial(string serial)
         {
+            if (serial == null || serial == "") return false;
             return (serial.IndexOf("SN") == 0);
         }
 
         public static bool IsJoystickSerial(string serial)
         {
+            if (serial == null || serial == "") return false;
             return (serial.IndexOf(Joystick.SerialPrefix) == 0);
+        }
+
+        public static bool IsRawSerial(string serial)
+        {
+            return (serial != null && serial.Contains(SerialSeparator));
         }
     }
 }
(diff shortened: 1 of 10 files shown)


tigert:
This works.

Also, when testing the Crosswind, the two buttons don't seem to do anything . there might be an option for simple on/off brake switch for some models?

Also, 

- X Axis is left brake
- Y Axis is right brake
- Z-rotation is Rudder

DocMoebiuz:
you are saying there are two buttons found, but you cannot assign them? can you explain a bit more please.

DocMoebiuz:
> This works.
> 
> Also, when testing the Crosswind, the two buttons don't seem to do anything . there might be an option for simple on/off brake switch for some models?
> 
> Also,
> 
> * X Axis is left brake
> * Y Axis is right brake
> * Z-rotation is Rudder

I don't understand what you mean.
i guess you are trying to say that there are two buttons reported, but the pedals dont have any buttons.

with the second "also", i think you are trying to say that friendly names for your axis should be "Left Brake", "Right Brake" & "Rudder" - renaming of labels is a different topic, it doesn't have to do with this PR. 

I don't consider the reported buttons as a defect related to this PR either. So I will merge this.
We can think of adding a "hide" function to the joystick.json files later.

tigert:
The USB device exposes two buttons, but they do not react to anything on the pedals.

I guess they use the same usb module and might have on/off brakes (or some other functionality) for those buttons.

PR #1137, opened 2023-02-12
Allow modifiers to appear after interpolation

### LINKED ISSUE #1136: Modifiers placed after interpolation modifiers never get loaded
**Describe the bug**
If a modifier is placed after the interpolation modifier in the config XML then it never gets loaded.

**Steps To Reproduce**
Create an output config with modifiers ordered like this (you will need to hand edit the XML to get it to this state):

```xml
        <modifiers>
          <transformation active="True" expression="$*M" substrStart="0" substrEnd="7" />
          <interpolation active="True">
            <value x="0" y="0" />
            <value x="100" y="255" />
          </interpolation>
          <comparison active="False" value="" operand="" ifValue="" elseValue="" />
        </modifiers>
```

**Expected behavior**
The comparison modifier gets loaded.

**Actual behavior**
The last modifier to get loaded is interpolation. The comparison modifier never gets loaded.

neilenns:
Fixes #1136 

I'm 99% sure this is the correct fix. There doesn't seem to be any need to do a `reader.Read()` at the end of processing the interpolation section of the XML.

### DIFF
--- a/MobiFlight/Modifier/Interpolation.cs
+++ b/MobiFlight/Modifier/Interpolation.cs
@@ -79,9 +79,6 @@ namespace MobiFlight.Modifier
                     reader.ReadToNextSibling("value");
                 } while (reader.LocalName == "value");
             }
-
-            if (reader.LocalName == "interpolation")
-                reader.Read(); // this closes the interpolation node
         }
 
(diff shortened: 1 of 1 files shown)


neilenns:
This fix doesn't work.

PR #1148, opened 2023-02-14
Load Joystick definitions from json files

neilenns:
Re-opening for review.


Koseng:
Current definition of inputs is like:
```
  "Inputs": [
    {
      "Name": "Button 1",
      "Label": "Button - Left Hand - White - Back"
    },
    {
      "Name": "Button 2",
      "Label": "Button - Left Hand - White - Front"
    },
```
This does not work on many systems, since the display names "Name" are language dependent. On my german system everything is called differently:

JoystickMobiflight

So instead the configuration should be like:
```
  "Inputs": [
    {
      "InputType": "Button",
      "Id": 1,
      "Label": "Button - Left Hand - White - Back"
    },
    {
      "InputType": "Button",
      "Id": 2,
      "Label": "Button - Left Hand - White - Front"
    },
```
with InputType being an enum of the available options. And then a way needs to be found to do the mapping based on that information. Also the "Name" needs to be auto generated then.

neilenns:
Lots of discussion about this in Discord, it was a bug that existed in main that carried over to my code. I've updated this code to use the prefix name instead.

Koseng:
Now:
image

Koseng:
Nice! 
And since there is no inheritance any more you can also replace the rest of the "protected virtual" stuff in Joystick.cs with "private".

neilenns:
> Nice! And since there is no inheritance any more you can also replace the rest of the "protected virtual" stuff in Joystick.cs with "private".

Oooh yes, missed that. I'll clean it up!

MobiFlight/Joysticks/MFJoystick.cs
```diff
@@ -149,21 +177,18 @@ protected virtual void EnumerateDevices()
...
+        public string MapDeviceNameToLabel(string deviceName)
         {
-            return MapDeviceNameToLabel(Regex.Replace(name, @"\d+", v.ToString()).ToString());
-        }
-
-        public virtual string MapDeviceNameToLabel(string deviceName)
-        {
-            var result = deviceName;
+            // First try and look for a custom label.
+           var input = Definition?.FindInputByName(deviceName);
```
DocMoebiuz, line 183: wrong indentation :D

DocMoebiuz (changes requested):
Awesome refactoring. But the MFJoystick renaming, I don't understand the reason and it lead to many changed files. I am curious to hear why.

If there is no compelling reason, then let's revert the MFJoystick renaming.

Was there a conflict with the Joystick class from the library?

Base/AppTelemetry.cs
```diff
@@ -86,7 +86,7 @@ public void ConfigLoaded(ConfigFile configFile)
             foreach (InputConfigItem item in inputConfigs)
             {
                 String key = "input." + item.Type;
-                if (item.ModuleSerial.Contains(Joystick.SerialPrefix))
+                if (item.ModuleSerial.Contains(MFJoystick.SerialPrefix))
```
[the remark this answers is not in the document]
neilenns: So I went back and remembered why I did this. The problem is:

```C#
        private readonly SharpDX.DirectInput.Joystick joystick;
```

In Joystic.cs It should be `Joystick` per the property naming conventions. But you can't have a property have the same name as the class it is in, and the class is already called `Joystick`. So I renamed the class I guess.

We can go the other way and rename the property. What would be something instead of `Joystick` to use?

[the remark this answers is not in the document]
neilenns: Undid this change and instead renamed the property to `DXJoystick` inside the class.

neilenns:
@DocMoebiuz Comments addressed and changes from `main` merged in.

MobiFlight/Joysticks/LabeledJoystick.cs
neilenns, line 1: Refactored away, it's now just part of `Joystick.cs`

MobiFlight/Joysticks/SaitekAviatorStick.cs
neilenns, line 1: Refactored away, it's now just part of `Joystick.cs`

DocMoebiuz (approved):
YEAH!

PR #1154, opened 2023-02-15
Replace Pico reset firmware with flash_nuke.uf2

### LINKED ISSUE #1153: Pico reset board firmware isn't flash_nuke.uf2
I'm not sure where the firmware for resetting the Pico came from, but it's not the official "flash_nuke.uf2" file from Raspberry Pi that truly resets the board back to factory default and brings it up in bootsel mode.

We should replace the existing reset firmware with flash_nuke.uf2 from here: https://cdn-learn.adafruit.com/assets/assets/000/099/419/original/flash_nuke.uf2?1613329170

neilenns:
Fixes #1153 

Replace the file with flash_nuke.uf2


neilenns:
Didn't rename it to actually be flash_nuke.uf2 to maintain consistency with our other reset firmware filenames. But it's the right one now, only 25k in size.

elral:
Wouldn't it be better to keep the original name and changing the board.json file?
Then it's more clear where the file is coming from respectively it is not coming from the Mobiflight-EEPROM-clear repo.

neilenns:
> Wouldn't it be better to keep the original name and changing the board.json file? Then it's more clear where the file is coming from respectively it is not coming from the Mobiflight-EEPROM-clear repo.

You're right. I renamed it to `reset.raspberry_pico_flash_nuke.uf2`.

PR #1156, opened 2023-02-16
Reuse Usb drive detection logic when flashing COM-connected Picos

### LINKED ISSUE #1155: Flashing a COM-connected Pico might fail if network drives/optical media are connected
In #1074 we realized optical drives that weren't ready could cause a crash when the volume label was ready.
In #1089 we realized connected network devices take FOREVER to return their drive name.

Both of these were fixed but only for the startup sequence. The same fundamental `Where()` clause is also used in the flashing flow when toggling an existing, COM connected, Pico then attempting to flash: https://github.dev/MobiFlight/MobiFlight-Connector/blob/cea64103ec105ed58a261305315a98122e1494b5/MobiFlight/MobiFlightFirmwareUpdater.cs#L189

It won't crash because of the try/catch, but should still be updated to check for `IsReady` and `DeviceType.Removable`.

neilenns:
Fixes #1155 

To ensure consistency in how USB drives are detected get rid of the detection code when flashing COM-connected Picos and reuse the same code from the Mobiflight startup sequence.

### DIFF
--- a/MobiFlight/MobiFlightFirmwareUpdater.cs
+++ b/MobiFlight/MobiFlightFirmwareUpdater.cs
@@ -194,7 +194,7 @@ namespace MobiFlight
                 // USB devices. What's needed for flashing however is a single USB drive whose volume label
                 // matches the volume lable in the .board.json of the device we toggled the COM port on.
                 // Attempt to find it.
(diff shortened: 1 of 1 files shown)


DocMoebiuz (approved):
I am Sebastian, and I approve this PR

Issue #1173, opened 2023-03-05
Burst of Logging with active LogPanel leads to CPU spike


Koseng:
**Describe the bug**
When LogPanel is active, bursts in logging still lead to a spike in CPU load and lag towards the sim. On my system a burst of 130 log messages leads to a CPU load of 6% for about 3 seconds.

Reason are the many redrawings in the LogPanel.

**Expected behavior**
CPU load should show no spike.

**Actual behavior**
On my system a burst of 130 log messages leads to a CPU load of 6% for about 3 seconds.

**Details**
 - MobiFlight Version: 9.7

PR #1174, opened 2023-03-05
Redraw LogPanel at most every 50ms

Koseng:
This PR solves #1173 . Many log entries in a short period of time are no problem any more. 

On my system CPU load now stays pretty normal at around 1% when showing the LogPanel. Before a burst led to 6% for about 3 seconds

### DIFF
--- a/UI/Panels/LogPanel.cs
+++ b/UI/Panels/LogPanel.cs
@@ -1,24 +1,31 @@
 ﻿using System;
-using System.Collections.Generic;
-using System.ComponentModel;
-using System.Data;
-using System.Drawing;
-using System.Linq;
 using System.Reflection;
-using System.Text;
-using System.Threading.Tasks;
 using System.Windows.Forms;
 
 namespace MobiFlight.UI.Panels
 {
     public partial class LogPanel : UserControl, ILogAppender
     {
+        private bool IsDrawingSuspended = false;
+        private System.Windows.Forms.Timer CheckDrawingSuspendedTimer = new Timer();
         public LogPanel()
         {
             InitializeComponent();
             listView1.Items.Clear();
             listView1.Resize += ListView1_Resize;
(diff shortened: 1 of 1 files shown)


UI/Panels/LogPanel.cs
```diff
@@ -1,24 +1,31 @@
...
     {
+        private bool IsDrawingSuspended = false;
+        private System.Windows.Forms.Timer CheckDrawingSuspendedTimer = new Timer();
         public LogPanel()
         {
             InitializeComponent();
             listView1.Items.Clear();
             listView1.Resize += ListView1_Resize;
             listView1.DoubleBuffered(true);
+            CheckDrawingSuspendedTimer.Interval = 50;
```
neilenns, line 17: Any reason you picked 50ms? That's still way faster than any human could process a log event, why not go with 250ms (to pick another random number :)

Koseng: Yes. Any slower with a delay like 100ms or 200ms did not improve cpu or lag any further. But user experience is a bit less nice. And faster makes no sense, because the maximum accuracy of windows forms timer is around 55ms.

DocMoebiuz (commented):
@Koseng have you tried an approach like this to not utilize the timer for these one-off runs? Not sure if there is a performance drawback:

``` c#
public void log(string message, LogSeverity severity)
{
    if (!IsDrawingSuspended)
    {
        IsDrawingSuspended = true;
        listView1.BeginUpdate();
        Task.Delay(1000).ContinueWith(t => {
            IsDrawingSuspended = false;
            listView1.EndUpdate();
        }, TaskScheduler.FromCurrentSynchronizationContext());
    }
    ListViewItem item = new ListViewItem();
```

No timer needed. This block could be put in a method like `SuspendAndAutoResume`

PR #1179, opened 2023-03-14
New directory "Docs" created

GioCC:
Multiplexer info document updated and moved there from root; 
also updated and added another document (Input matrix with multiplexers) which was "in the wind" until now


DocMoebiuz:
Thanks @GioCC for the docs - however, I won't include PDF files here. Please be so kind and create a Wiki page for the information. I don't want PDF files for documentation in general but since you already created one, you can also attach it to the Wiki. But this shall be an exception.

I will remove the PDF file from the root... i simply missed that this was added there before.

MobiFlight-Admin:
> > since you already created one, you can also attach it to the Wiki. But this shall be an exception.
> 
> No reason to, I can convert both files to markdown and put it on the wiki.

That would be awesome!!!

GioCC:
Are you OK with me adding a "Technical notes" section to the wiki page? These documents are not exactly aimed at novice users, and they don't really fit elsewhere.
I was thinking about adding also an actual beginner's guide for multiplexers; that's still missing, and would go under "Examples and tutorials".

MobiFlight-Admin:
a beginners guide would be great.  definitely

elral:
I have also a couple of documentation regarding the firmware I am looking for where to store. So a "Technical notes" section would be great.

PR #1181, opened 2023-03-14
Prevent the exception when sorting by column header

### LINKED ISSUE #1162: MobiFlight shows exception dialog when user selects last row and then tries to sort the gird view
**Describe the bug**
MobiFlight crashes when user selects last row and then tries to sort the gird view

**Steps To Reproduce**
1. Select the last "empty" line from the gridView
2. Click on any of the column headers for sorting

**Expected behavior**
GridView content is sorted by column values

**Actual behavior**
An exception is thrown.

**Details**
 - MobiFlight Version: [e.g. 8.2.0]
 - FlightSim: [e.g. MSFS2020]

**Additional context**
This problem exists for Output and Input Grid Views

** Suggested fix **
Add `?`

``` C#
DataRow currentRow = (row.DataBoundItem as DataRowView)?.Row;
if (currentRow != null)
   SelectedGuids.Add(currentRow["guid"].ToString());
```

image

DocMoebiuz:
fixes #1162 

- [x] Add `?` to the places where DataBoundItem is access and not yet checked for `null`

### DIFF
--- a/UI/Panels/InputConfigPanel.cs
+++ b/UI/Panels/InputConfigPanel.cs
@@ -116,10 +116,11 @@ namespace MobiFlight.UI.Panels
                     } //if
 
                     InputConfigItem cfg;
-                    DataRow row = null;
                     bool create = false;
+                    DataRow row = (inputsDataGridView.Rows[e.RowIndex].DataBoundItem as DataRowView)?.Row;
 
-                    row = (inputsDataGridView.Rows[e.RowIndex].DataBoundItem as DataRowView).Row;
+                    if (row == null) 
+                        break;
 
                     // the row had been saved but no config object has been created
                     // TODO: move this logic to an appropriate event, e.g. when leaving the gridrow focus of the new row
@@ -202,7 +203,9 @@ namespace MobiFlight.UI.Panels
                 // duplicate it
                 // duplicate row 
                 // link to new config item 
-                DataRow currentRow = (row.DataBoundItem as DataRowView).Row;
+                DataRow currentRow = (row.DataBoundItem as DataRowView)?.Row;
+                if (currentRow == null) continue;
+
                 DataRow newRow = inputsDataTable.NewRow();
 
                 foreach (DataColumn col in inputsDataTable.Columns)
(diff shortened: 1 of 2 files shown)


UI/Panels/InputConfigPanel.cs
```diff
@@ -116,10 +116,11 @@ private void inputsDataGridView_CellContentClick(object sender, DataGridViewCell
                     } //if
 
                     InputConfigItem cfg;
-                    DataRow row = null;
                     bool create = false;
+                    DataRow row = (inputsDataGridView.Rows[e.RowIndex].DataBoundItem as DataRowView)?.Row;
```
neilenns, line 120: I think this should be `var` instead of `DataRow` (style-wise)

neilenns: And elsewhere in this PR as well.

neilenns (approved):

UI/Panels/InputConfigPanel.cs
DocMoebiuz: I want to keep changes to the  minimum at this point. Thanks for pointing this out though!

DocMoebiuz: @neilenns I was thinking about this particular case, and I must say: I cannot easily defer the type of `.Row` and so putting `DataRow` in front of the variable makes that crystal clear.

I also started to use `var` when initializing variables and when it is clear by looking at the right hand side, what data type is used. Makes it shorter and easier to read.

neilenns: Went and looked at what MSFT says about this since I was curious: https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions.

> Use [implicit typing](https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/implicitly-typed-local-variables) for local variables when the type of the variable is obvious from the right side of the assignment, or when the precise type is not important.

Then they say to use the explicit type if it isn't clear from the right side (as you mention above). I guess the question is whether knowing it is a DataRow is important :D

PR #1182, opened 2023-03-14
Float64 numbers are displayed correctly in the `Sim Value` column

### LINKED ISSUE #1161: Float64 numbers are not displayed correctly in the `Sim Value` column
**Describe the bug**
With 9.6, internally Float64 (or 'Number') is used as long as possible to allow for proper interpretation of the type in transformations and modifiers.

The problem is, that the resolution is higher than the default .ToString() - so when a number eventually is converted into a string for display in the Sim Value column, the last two digits will get cut off. 

If a user uses the displayed value from the Sim Value column in a comparison expression, then the value will not match, and the user will be confused. There is no way at the moment to see the actual internal value.

**Steps To Reproduce**
1. Create a config that reads the trailing flaps left value
2. Load the DA20 in MSFS2020
3. Run MobiFlight

**Expected behavior**
Flaps value displayed in Sim Var column is: `0.37037035822868347` 

**Actual behavior**
Flaps value displayed in Sim Var column is: `0.370370358228683` 

**Details**
 - MobiFlight Version: 9.7

DocMoebiuz:
fixes #1161 

- [x] Adding "R" as format type for ToString()

### DIFF
--- a/Base/ConnectorValue.cs
+++ b/Base/ConnectorValue.cs
@@ -21,10 +21,10 @@ namespace MobiFlight
                     result = String;
                     break;
                 case FSUIPCOffsetType.Integer:
-                    result = Float64.ToString();
+                    result = Float64.ToString("R");
                     break;
                 case FSUIPCOffsetType.Float:
-                    result = Float64.ToString();
+                    result = Float64.ToString("R");
                     break;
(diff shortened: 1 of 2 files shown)


neilenns (approved):
Fancy!

neilenns (approved):

PR #1183, opened 2023-03-14
Fixing the Honeycomb LEDs for float values

### LINKED ISSUE #1175: HID Output LED (Bravo Throttle) not accept Floats
**Describe the bug**
Using a Float Output Value to Controll a LED (On Bravo Throttle) result in a STOP of MF
"Input string in wrong format"

**Steps To Reproduce**
1. Create a Output Config
2. Use GEAR LEFT POSITION Preset (generic)
3. Set to a Bravo LED
4. RUN MF and move gear in Sim

**Expected behavior**
Value goes from 0,0 too 100,0 as a float
LED accept it... MF without Device accept it.    Bravo should also do.

**Actual behavior**
STOP in MF when we try to send a Float to Bravo

**Details**
 - MobiFlight Version: 9.7.0

**Additional context**
I expect also a Config with transform   1.234 should make this reproduceable.

DocMoebiuz:
fixes #1175 

- [x] applying same logic as for LEDs and OutputShiftRegister
- [x] changed `SetOutputDeviceState` function param to `byte`
- [x] fixed test for value change and required update

### DIFF
--- a/MobiFlight/Joystick.cs
+++ b/MobiFlight/Joystick.cs
@@ -399,8 +399,6 @@ namespace MobiFlight
             foreach(var light in Lights)
             {
                 if (light.Label != name) continue;
-
-                state = (byte)(state > 0 ? 1 : 0);
                 if (light.State == state) continue;
 
                 light.State = state;
--- a/MobiFlight/Joystick.cs
+++ b/MobiFlight/Joystick.cs
@@ -392,16 +392,18 @@ namespace MobiFlight
             return UsageMap[usage];
         }
 
-        public void SetOutputDeviceState(string name, string value)
+        public void SetOutputDeviceState(string name, byte state)
         {
-            byte state = byte.Parse(value);
+            double tempValue;
 
             foreach(var light in Lights)
             {
                 if (light.Label != name) continue;
+
+                state = (byte)(state > 0 ? 1 : 0);
                 if (light.State == state) continue;
 
-                light.State = (byte) (state > 0 ? 1 : 0);
+                light.State = state;
                 RequiresOutputUpdate = true;
                 return;
             }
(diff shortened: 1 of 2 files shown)


neilenns:
NoOOOOOooooOOOOOO now I will have merge conflicts in my joystick PR 😭

MobiFlight/Joystick.cs
```diff
@@ -392,16 +392,18 @@ private String GetAxisNameForUsage(int usage)
...
+        public void SetOutputDeviceState(string name, byte state)
         {
-            byte state = byte.Parse(value);
+            double tempValue;
 
             foreach(var light in Lights)
             {
                 if (light.Label != name) continue;
+
+                state = (byte)(state > 0 ? 1 : 0);
```
neilenns, line 403: Why not have this take a Bool instead and demand the callers provide the correctly formatted data in the first place? Then you can just cast it to a byte without all the other stuff. The only place this is called now anyway is ExecutionManager, and that's already explicitly only sending 0 or 1.

MobiFlight/Joystick.cs
```diff
@@ -392,16 +392,18 @@ private String GetAxisNameForUsage(int usage)
             return UsageMap[usage];
         }
 
-        public void SetOutputDeviceState(string name, string value)
+        public void SetOutputDeviceState(string name, byte state)
         {
-            byte state = byte.Parse(value);
+            double tempValue;
```
neilenns, line 397: This doesn't seem used anywhere?

neilenns (changes requested):

MobiFlight/Joystick.cs
DocMoebiuz: yes. true. With that reasoning in mind i made the change of the function signature and I am providing a byte now. I will remove it.

MobiFlight/Joystick.cs
DocMoebiuz: No. Great eye! 👀

neilenns (approved):

PR #1184, opened 2023-03-14
FSUIPC presets will now show integer values again

### LINKED ISSUE #1172: FSUIPC presets will now show decimal values (floats)
With the introduction of internally using float as type for number values, a lot of the FSUIPC presets that require transformation will now be treated as floats. This might cause issues and confusion for users.

Problematic are Heading values on Autopilot which are displayed on 7-segment dislpays. Here the new float treatment (automatic decimal point) will kick in and break existing configs.

**Additional context**
A solution is to actually floor () "integer" type, which is set for FSUIPC

DocMoebiuz:
fixes #1172 

- [x] Maintain `FSUIPCOffsetType.Integer` in Connector Value
- [x] if of type `FSUIPCOffsetType.Integer`, ToString() now applies `Math.Round()` to the internal Float64 value 
- [x] Updated unit tests
- [x] Test in sim with FSUIPC7 and Autopilot Heading - was not working before fix, now it is working.

### DIFF
--- a/Base/ConnectorValue.cs
+++ b/Base/ConnectorValue.cs
@@ -21,7 +21,7 @@ namespace MobiFlight
                     result = String;
                     break;
                 case FSUIPCOffsetType.Integer:
-                    result = Float64.ToString("R");
+                    result = Math.Round(Float64).ToString();
                     break;
                 case FSUIPCOffsetType.Float:
                     result = Float64.ToString("R");
--- a/MobiFlight/Modifier/Comparison.cs
+++ b/MobiFlight/Modifier/Comparison.cs
@@ -152,7 +152,6 @@ namespace MobiFlight.Modifier
             try
             {
                 result.Float64 = Double.Parse(comparisonResult);
-                result.type = FSUIPCOffsetType.Float;
                 if (result.Float64.ToString()!=comparisonResult.ToString())
                 {
                     result.String = comparisonResult;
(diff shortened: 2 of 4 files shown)


Base/ConnectorValue.cs
```diff
@@ -21,7 +21,7 @@ public override string ToString()
                     result = String;
                     break;
                 case FSUIPCOffsetType.Integer:
-                    result = Float64.ToString("R");
+                    result = Math.Round(Float64).ToString();
```
neilenns, line 24: Why is this PR doing it this way and [the other PR used "R"?](https://github.com/MobiFlight/MobiFlight-Connector/pull/1182/files)?

neilenns (approved):

Base/ConnectorValue.cs
MobiFlight-Admin: Connector Value remembers what the original value type was. Internally it uses always floats.

At the moment, Mobiflight converts the ConnectorValue using ToString() before sending it to a display.

ToString() now converts the number to a whole number (round) when it originally was a FSUIpCOffsetType.Integer

MobiFlight-Admin: this really only applies to FSUIPC offsets and the weird transformations (eg, heading is $*360/65000) but these values are commonly used in the context of FSUIPC.

PR #1185, opened 2023-03-14
Prevent Error message when opening an output config

### LINKED ISSUE #1145: "Trying to show config but display type Output not present." when opening an output config
**Describe the bug**
When opening an output config I get the error "Trying to show config but display type Output not present." when opening an output config.

This appears to be from the renaming that happened to rename `Output` to `LED / Output` in the dialog.

The line of code where this gets logged is here:: https://github.dev/MobiFlight/MobiFlight-Connector/blob/cea64103ec105ed58a261305315a98122e1494b5/UI/Panels/OutputWizard/DisplayPanel.cs#L131-L134. At that point in the code `config.DisplayType` is `Output` but the combo box has `LED / Output` in it.

DocMoebiuz:
fixes #1145 

- [x] Use correct helper-method to select by value and not label

### DIFF
--- a/UI/Panels/OutputWizard/DisplayPanel.cs
+++ b/UI/Panels/OutputWizard/DisplayPanel.cs
@@ -367,7 +367,7 @@ namespace MobiFlight.UI.Panels.OutputWizard
 
                 // third tab
                 if (config.DisplayType != null && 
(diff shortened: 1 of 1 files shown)


neilenns (approved):
Easy fix!

PR #1191, opened 2023-03-27
Add debounce to HubHop search input

### LINKED ISSUE #1192: Difficult to type on Search input for Filter Preset List
**Describe the bug**
As the "Filter Preset List" logic happens for every keystroke on the Search input, the UI stutters with every key pressed.

Ideally, the UI should wait for the user to finish typing, and only then execute the filter, to avoid unintended typos caused by the user thinking the UI hasn't registered the key typed when it's just a temporary UI freeze happening

**Steps To Reproduce**
1. Go to Outputs (or Inputs)
2. Click on '....' (edit)
3. Select SimConnect (MSFS2020) selected as "Variable Type"
4. Click on "Reset" button (for increased effect)
5. Start typing something that matches lots of presets like: `indicator`, `on` or `_` 

**Expected behavior**
The UI should not freeze on every keystroke because of filtering, this helps with typing, avoiding typos and unnecessary wait or cleaning up with backspace

**Actual behavior**
The UI stutters for every keystroke

**Details**
 - MobiFlight Version: [e.g. 9.7.0]
 - FlightSim: MSFS2020 SimConnect

**Additional context**
In a broader UI context, MobiFlight should (ideally) leverage the background thread in WinForms in an effort to avoid main UI thread freezes (which can be interpreted as problems or "bugs"), but this is not something fixable with a small tweak, though.

Also, while https://github.com/MobiFlight/MobiFlight-Connector/issues/1118 would be a great UI change as well, this would make the issue being reported here even more noticeable, since the user would always interact with the "Search" box first, and only add more filters later (which is a common practice even in e-commerces)

lscarneiro:
## Why
In the current implementation, it can be slow to input search terms in both Output and Input Preset filters if you don't narrow down your search categories first.

fixes #1192

## How
The Idea here is to add a "debounce" of `300ms` (which can be tuned up or down if needed) so we can defer the heavy filter logic to a moment after the user has finished typing its search term

## Extra
During my initial compilation of the Solution, I noticed a mismatch in the `Newtonsoft.Json` version, which was failing the build, so I updated it, and now it compiles successfully locally. I can remove this part of the change if it's needed

> This is my first contribution to MobiFlight =D


UI/Panels/Config/HubHopPresetPanel.cs
```diff
@@ -510,7 +511,25 @@ private void MinimizeSimVarNameTextBox()
 
         private void textBox1_TextChanged(object sender, EventArgs e)
         {
-            FilterPresetList();
+            // Cancel any previous timer
```
DocMoebiuz, line 514: no need to comment every line. they are self-explanatory. 
Please remove them.

MobiFlight-Admin (approved):
cool!

DocMoebiuz (approved):

DocMoebiuz:
@lscarneiro updating libraries is better done in a specific pull request - you are right about your comment.

I am wondering why it was failing for you - and why is the "hint" section still referring to net45 when the target framework is 4.8 - never noticed that

DocMoebiuz:
> @lscarneiro updating libraries is better done in a specific pull request - you are right about your comment.
> 
> I am wondering why it was failing for you - and why is the "hint" section still referring to net45 when the target framework is 4.8 - never noticed that

we make it right the next time

PR #1195, opened 2023-04-01
Initial midiboard support

Koseng:
This PR implements initial generic midi board support for Mobiflight similar to the joystick support. See #597.

**Includes configuration files for:**

- [Behringer X-TOUCH MINI](https://www.behringer.com/product.html?modelCode=P0B3M)
- [KORG nanoKONTROL2](https://www.korg.com/us/products/computergear/nanokontrol2/)
- [Akai MPD218](https://www.akaipro.com/mpd218)
- [Akai MPK MINI PLAY](https://www.akaipro.com/mpk-mini-play-mpkminiplay)
- [Intech EF44](https://intech.studio/shop/modules/ef44)

Without a configuration file for a type of board, inputs can be seen in the logging and tested  with the “Scan for input” feature. For serious usage of a midi board type, a configuration file should be created.

**Implementation remarks:**

- The used midi library is not available as a nuget package. Therefore the dll is included.
- The only reliable identification of a midi board is its name. If several boards of same type are connected, they all have the same name. So a postfix to the name shown must be generated. Also output and input device for one midi board can only be correlated via the name. The used serial number in mobiflight is generated from the name.
- For midi boards connected during startup of mobiflight, also later disconnect and reconnect during runtime is supported.
- Renamed Start() to Startup() in JoystickManager and MidiboardManager because it is executed only once on startup of application and not after each Start in Mobiflight.
- Output identification is currently done via label like in joystick.
- It is not possible to query midi devices for its capabilities, like it is done with joysticks.

Thanks to Jaime for doing some tests on the Akai MPD218 and thanks to Ape42 for doing some tests on the KORG nanoKONTROL2.

**Planned next steps:**

- Write documentation in Mobiflight wiki
- Output identification via id instead of label

Also implementation wise think about where to use common interfaces between module, joystick and midiboard.


MobiFlight/MidiBoard/MidiBoardDefinition.cs
```diff
@@ -0,0 +1,50 @@
...
+    public class MidiBoardDefinition
+    {
+        // From deserialization of config
+        public string InstanceName;
+
+        // From deserialization of config
+        public string DifferingOutputName;
+
+        // From deserialization of config
+        public int EncoderNeutral = 0;
```
neilenns, line 14: The schema said this is a position if I remember right. I suggest calling this `EncoderNeutralPosition` to be super clear on its use.

neilenns: Also there should be doc comments on each of the properties to match what was listed in the json file.

MobiFlight/MidiBoard/MidiBoardDeviceType.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System.Runtime.Serialization;
+
+namespace MobiFlight
+{
+    public enum MidiBoardDeviceType
+    {
+        [EnumMember(Value = "Button")]
```
neilenns, line 7: Is this needed to serialize/deserialize from JSON? I thought it would Just Work(tm) with the strings, no decorations required.

MobiFlight/MidiBoard/MidiBoardManager.cs
```diff
@@ -0,0 +1,273 @@
...
+                        errorMessages.Add($"{item.Label}: Label does not contain % placeholder.");
+                }
+            }
+            return errorMessages;
+        }
+
+        private List<string> CollectErrorsInOutputDefinition(MidiBoardDefinition def)
+        {
+            List<string> errorMessages = new List<string>();
+            if (def.Outputs != null)
```
neilenns, line 50: I don't believe this if statement is necessary. The following foreach just won't iterate if `def.Outputs` is null.

MobiFlight/MidiBoard/MidiBoardManager.cs
```diff
@@ -0,0 +1,273 @@
...
+            {
+                foreach (var item in def.Outputs)
+                {
+                    if (item.LabelIds.Length != item.MessageIds.Length)
+                        errorMessages.Add($"{item.Label}: Number of LabelIds unequal to number of MessageIds");
+                    if (!item.Label.Contains("%"))
+                        errorMessages.Add($"{item.Label}: Label does not contain % placeholder.");
+                    if (!string.IsNullOrEmpty(item.RelatedInputLabel))
```
neilenns, line 58: This nested if statement can be unfolded by changing this to `if (string.IsNullOrEmpty(item.RelatedInputLabel)) continue;`.

MobiFlight/MidiBoard/MidiBoardManager.cs
```diff
@@ -0,0 +1,273 @@
...
+                        if (!item.RelatedInputLabel.Contains("%"))
+                            errorMessages.Add($"{item.Label}: RelatedInputLabel does not contain % placeholder.");
+                    }
+                }
+            }
+            return errorMessages;
+        }
+
+
+        private bool IsErrorInDefinition(MidiBoardDefinition def)
```
neilenns, line 71: It's a little weird to have a function called `IsErrorInDefinition()` both test if there is an error (the return statement) *and* log all the error messages.

MobiFlight/MidiBoard/MidiBoardManager.cs
```diff
@@ -0,0 +1,273 @@
...
+                    }                    
+                }
+                catch (Exception ex)
+                {
+                    Log.Instance.log($"Unable to load {definitionFile}: {ex.Message}", LogSeverity.Error);
+                }
+            }         
+        }
+
+        public bool MidiBoardsConnected()
```
neilenns, line 104: Could be renamed to `AreMidiBoardsConnected()` to make it clear it's a yes/no instead of a count.

MobiFlight/MidiBoard/MidiBoardManager.cs
```diff
@@ -0,0 +1,273 @@
...
+        }
+      
+        public void Startup()
+        {
+            ProcessTimer.Start();
+        }
+
+        private void ProcessTimer_Tick(object sender, EventArgs e)
+        {    
+            if (BoardsToBeRemoved.Count > 0)
```
neilenns, line 116: I don't think this if is needed. You can just run the `foreach` and if there are no boards to remove it'll do nothing. There's no issue calling `BoardsToBeRemoved.Clear()` every tick either since it'll noop if there are no boards to be removed.

MobiFlight/MidiBoard/MidiBoardManager.cs
```diff
@@ -0,0 +1,273 @@
...
+            foreach (var mb in MidiBoards)
+            {
+                mb.ProcessMessages();
+            }
+            UpdateOnAttachedOrRemovedMidiBoards();           
+        }
+
+        private void UpdateOnAttachedOrRemovedMidiBoards()
+        {
+            if (CheckAttachedRemovedCounter >= 40) // Check every 2 seconds
```
neilenns, line 133: Similar comment to above. You can remove a layer of if statements by converting this to:

`if (CheckAttachedRemovedCounter < 40) return;`

MobiFlight/MidiBoard/MidiBoardManager.cs
```diff
@@ -0,0 +1,273 @@
...
+                mb.ProcessMessages();
+            }
+            UpdateOnAttachedOrRemovedMidiBoards();           
+        }
+
+        private void UpdateOnAttachedOrRemovedMidiBoards()
+        {
+            if (CheckAttachedRemovedCounter >= 40) // Check every 2 seconds  
+            {               
+                if (MidiDevice.InputsCount != MidiBoards.Count)
```
neilenns, line 135: This can also be unwound and converted to:

`if (MidiDevice.InputsCount == MidiBoards.Count) return;`

MobiFlight/MidiBoard/MidiInputDefinition.cs
```diff
@@ -0,0 +1,43 @@
+﻿using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace MobiFlight
+{
+    public class MidiInputDefinition
+    {
+        public string Label;
```
neilenns, line 8: All these properties should get doc comments explaining what they are.

MobiFlight/MidiBoard/MidiOutputDefinition.cs
```diff
@@ -0,0 +1,36 @@
+﻿using Newtonsoft.Json;
+using Newtonsoft.Json.Converters;
+
+namespace MobiFlight
+{
+    public class MidiOutputDefinition
+    {
+        public string Label;
```
neilenns, line 8: Doc comments :)

UI/MainForm.cs
```diff
@@ -1247,6 +1247,56 @@ private void _checkForOrphanedJoysticks(bool showNotNecessaryMessage)
...
+            {
+                if (item.ModuleSerial.Contains(MidiBoard.SerialPrefix) &&
+                    !serials.Contains(item.ModuleSerial) &&
+                    !NotConnectedMidiBoards.Contains(item.ModuleSerial))
+                {
+                    NotConnectedMidiBoards.Add(item.ModuleSerial);
+                }
+            }
+
+            if (NotConnectedMidiBoards.Count > 0)
```
neilenns, line 1282: This can be unwound, changed to `if (NotConnectedMidiBoards.Count == 0) return`;

neilenns (changes requested):
Mostly looked at code style stuff. Commented in several places where if statements can be unpacked to reduce the depth of the code (#nevernester!), which makes it easier to read long term. Here's a good video about it: https://www.youtube.com/watch?v=CFRhGnuXG-4

MobiFlight/MidiBoard/MidiBoardManager.cs
neilenns: Hmm, it is necessary. Instead this should be `if (def.Outputs == null) return errorMessages`;

MobiFlight/MidiBoard/MidiBoardDefinition.cs
Koseng: Renaming to EncoderNeutralPosition - I like.

Regarding the doc comment topic in general. I'm in the camp of trying to not write any method or variable comments at all, even if they are public. In general I would argue if a comment is necessary, the variable name or method name is not clear enough. Or the method is too long and does too many things and therefore a good name cannot be found. In my opinion when I feel a comment is necessary I should think about a renaming or restructuring instead.

Nevertheless I'll check where I think additional comments really make sense.

MobiFlight/MidiBoard/MidiBoardDeviceType.cs
Koseng: Thanks, I'm not sure. I'll check. There are other classes where it was done like that.

MobiFlight/MidiBoard/MidiBoardManager.cs
Koseng: Ok, this is a question of style and taste. I'm in the camp which likes only 1 return statement per function. In the end it is a minor thing and not worth endless discussions. Here some literature on that topic :-).

https://hackerchick.com/religious-war-48293-single-vs-multiple/
https://stschindler.medium.com/having-only-one-return-statement-is-better-and-more-safe-5cd379b7e87c
https://www.anthonysteele.co.uk/TheSingleReturnLaw.html

MobiFlight/MidiBoard/MidiBoardManager.cs
Koseng: Yep I like that suggestion.

(Even that's also a little bit like the single vs. multiple return statement discussion where people don't like code being skipped.)

Koseng: Ok tried it, I do like the current version better. I think current version makes it clearer that the other two if statements are clearly related to the existance of the RelatedInputLabel. Probably also a matter of style and what you are used to.

MobiFlight/MidiBoard/MidiBoardManager.cs
Koseng: Agree that should be handled and named differently.

MobiFlight/MidiBoard/MidiBoardManager.cs
Koseng: Good suggestion.

MobiFlight/MidiBoard/MidiBoardManager.cs
Koseng: Yeah, I can remove the check. The same check is done within the .Clear() Method in the framework.

MobiFlight/MidiBoard/MidiBoardDefinition.cs
neilenns: It's not a huge deal but I like adding doc comments to objects that are serialized/deserialized. As is there's no way to know on the code side which of these are optional, for example.

MobiFlight/MidiBoard/MidiBoardManager.cs
Koseng: Same style discussion.

MobiFlight/MidiBoard/MidiBoardManager.cs
Koseng: Same style discussion.

MobiFlight/MidiBoard/MidiBoardManager.cs
neilenns: I agree it's a matter of style. I read those and I don't think any of them argued against what I suggested though 🤣 Ultimately it's up to @DocMoebiuz but the current codebase does tend to prefer early returns/continues over nested ifs when it comes to validating variables before executing the core part of a function.

UI/MainForm.cs
Koseng: Same style discussion. I need to think about in which cases I like something like that and when not.

MobiFlight/MidiBoard/MidiInputDefinition.cs
Koseng: I'll think about which ones really need a comment and are not self explanatory.

MobiFlight/MidiBoard/MidiOutputDefinition.cs
Koseng: I'll think about which ones really need a comment and are not self explanatory.

MobiFlight/MidiBoard/MidiBoardManager.cs
Koseng: Yes, I know. I justed wanted to show that there is an argument going on. In the end short methods are most important and prevent most crazy stuff. Rest is small optimizations.

MobiFlight/MidiBoard/MidiBoardDefinition.cs
Koseng: Good point, being optional is a good reason for a comment.

neilenns: And once you comment one of them you might as well comment all of them 🤣

MobiFlight/MidiBoard/MidiBoardDeviceType.cs
Koseng: Works. Thumbs up.

MobiFlight/MidiBoard/MidiBoardDefinition.cs
Koseng: jaja

MobiFlight/MidiBoard/MidiBoardManager.cs
DocMoebiuz: I defintely prefer early returns, it makes the code easier to read. Please adhere to the same style whenever possible.

MobiFlight/MidiBoard/MidiBoardManager.cs
DocMoebiuz: Please do.

MobiFlight/MidiBoard/MidiBoardManager.cs
DocMoebiuz: Please do.

Koseng:
Tested now with two midi boards of same type connected. Handling and naming needed to be changed. Also fixed issue with automatic reconnect when several midi boards are connected at the same time. Finally added configuration for [Akai MPK MINI PLAY](https://www.akaipro.com/mpk-mini-play-mpkminiplay).

MidiBoards/intech_ef44.midiboard.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "./mfmidiboard.schema.json",
+  "InstanceName": "Intech Studio: AC  ",
+  "Inputs": [
+    {
+      "Label": "Knob %",
+      "LabelIds": [ "1", "2", "3", "4" ],
+      "InputType": "EndlessKnob",
```
DocMoebiuz, line 8: Is there a reason this is not called "Encoder"?

MidiBoards/intech_ef44.midiboard.json
```diff
@@ -0,0 +1,30 @@
...
+      "LabelIds": [ "1", "2", "3", "4" ],
+      "InputType": "Button",
+      "MessageType": "CC",
+      "MessageChannel": 1,
+      "MessageIds": [ 0, 1, 2, 3 ]
+    },
+    {
+      "Label": "Slider %",
+      "LabelIds": [ "1", "2", "3", "4" ],
+      "InputType": "Fader",
```
DocMoebiuz, line 24: We could reuse "axis" or "analog input"

MobiFlight/MidiBoard/MidiOutputDefinition.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        /// <summary>
+        /// Midi message value for turning on the LED. Required.
+        /// </summary>
+        public byte ValueOn;
+
+        /// <summary>
+        /// Midi message value for putting LED to blink mode. Optional.
+        /// </summary>
+        public byte? ValueBlinkOn;
```
DocMoebiuz, line 47: How does blink work for midi?

DocMoebiuz (approved):
Thanks! Awesome PR!!!

PR #1204, opened 2023-04-30
PWM Pin option is shown only for PWM capable pins

### LINKED ISSUE #1201: PWM Mode in output configs remains selectable with non-PWM capable pins
**Describe the bug**
The PWM Mode checkbox in Output configs of LED/Output type remains on even if the selected pin does not have PWM capability.  This has already caused some confusion with one user trying to control brightness with a non-PWM capable pin. 

**Steps To Reproduce**
1. Create a LED device in any board with a non-PWM pin assigned.
2. Create output config for this LED
3. Select Display tab to configure the LED pin
4. Check on the PWM Mode checkbox
5. Note the box is active and selectable, but the pin used does not have this capability. 

**Expected behavior**
The Display tab config wizard should only activate the PMW Mode checkbox when the selected pin to display the LED status is actually capable of sending a PWM signal.

**Actual behavior**
The PWM Mode checkbox is active for all pins.

**Details**
 - MobiFlight Version: 9.7.1
 - FlightSim: irrelevant
 - AddOn: irrelevant

**Additional context**
I think the PWM mode checkbox used to be active only with PWM capable pins in the past.

DocMoebiuz:
fixes #1201 

- PWM pin option only displayed when a PWM-capable pin is selected
- PWM option now also available for multi-selection when all checked pins support PWM
- Fixed `Dock`-property and added padding to multiselect
- Multi-select option is displayed correctly

1201-pwm-option

- Added `IsArcazeSerial` as method to make it clear and similar to all the other boards. Arcaze Serial is detected when all other Serial Types are returning `false` (MobiFlight, Joystick, MidiBoard)
- Added unit tests (also for Midi Board Serial)

### DIFF
--- a/Base/SerialNumber.cs
+++ b/Base/SerialNumber.cs
@@ -33,6 +33,12 @@ namespace MobiFlight.Base
             return String.Join("", tokens).Trim();
         }
 
+        public static bool IsArcazeSerial(string serial)
+        {
+            if (serial == null || serial == "") return false;
+            return !IsMidiBoardSerial(serial) && !IsMobiFlightSerial(serial) && !IsJoystickSerial(serial);
+        }
+
         public static bool IsMobiFlightSerial(string serial)
         {
             if (serial == null || serial == "") return false;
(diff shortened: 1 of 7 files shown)


neilenns:
I guess I approve this 😆

PR #1206, opened 2023-05-01
Move LoadConfig() out of connect() for stable board detection

### LINKED ISSUE #1205: Boards are not recognized properly as MobiFlight boards
**Describe the bug**
Users report that during startup their boards are not detected properly and show as Arduino boards without MobiFlight firmware. After a couple of restarts the boards might all show ok.

**Steps To Reproduce**
1. Start MobiFlight

**Expected behavior**
All MobiFlight boards are detected properly

**Actual behavior**
A firmware upload popup appears and proposes to upload the MobiFlight firmware

**Details**
 - MobiFlight Version: 9.6+

DocMoebiuz:
related to #1205 

This is a potential improvement because during initial connect the config is not immediately loaded, and thereby we prevent to send a lot of data at once just right after connecting.

Changes:
- `LoadConfig` removed from `connect` method
- `LoadConfig` now explicitly called only for MobiFlight boards, after `GetInfo`

Feedback from the user who reported #1205 - it didn't make a difference, but I think it is an optimization to keep and see how it behaves for other users

### DIFF
--- a/MobiFlight/MobiFlightCache.cs
+++ b/MobiFlight/MobiFlightCache.cs
@@ -392,6 +392,7 @@ namespace MobiFlight
                     ModuleConnecting?.Invoke(this, "Connecting to MobiFlight Modules", progressValue);
                     module.Connect();
                     module.GetInfo();
+                    module.LoadConfig();
                     ModuleConnecting?.Invoke(this, "Connecting to MobiFlight Modules", progressValue);
                 }));
             }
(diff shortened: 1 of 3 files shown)


JaimeLeon2:
Could this be the reason for the non-detection of the Nano boards? MF 9.7.1 detects them OK.

elral:
I am still wondering why this does not happen for the other boards. I see no topic where they differ in board detection.
But it could be a try worth.

DocMoebiuz:
as i wrote in my comment: it seems to make sense but not sure if it has any unforeseen side effects - maybe @JaimeLeon2 found one of those

PR #1226, opened 2023-05-22
Set Arduino Mega connection delay to 1750

### LINKED ISSUE #1225: Mega boards need a 1500 connection delay
We've seen a bunch of issues connecting to Nanos lately. @elral chased down one of the (possibly many) issues.

Since there are CH340-based Megas out there it means there's a shared vid/pid between megas and nanos, and on initial connection to a nano the values from arduino_mega.board.json will get used (we use the first one alphabetically based on vid/pid when we don't know the mobiflight type).

This means the ConnectionDelay value used on initial connection is 1250, instead of the longer 1750 that seems necessary for CH340-based Nano boards.

Fix is to set the mega connection delay to 1750 as well.

While this doesn't solve the issue of the buggy CH340 windows driver, it does help resolve some of the failures we're seeing where a flashed Nano doesn't get seen as a Mobiflight board.

neilenns:
Fixes #1225 

Sets the delay for the mega to 1750 to match the Nano connection delay. Seems like CH340-based boards need the longer time.

### DIFF
--- a/Boards/arduino_mega.board.json
+++ b/Boards/arduino_mega.board.json
@@ -2,12 +2,12 @@
   "$schema": "./mfboard.schema.json",
   "AvrDudeSettings": {
     "Device": "atmega2560",
-    "BaudRates": [ "115200" ],
+    "BaudRates": ["115200"],
     "Programmer": "wiring",
     "Timeout": 15000
   },
   "Connection": {
-    "ConnectionDelay": 1250,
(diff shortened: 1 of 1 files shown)


DocMoebiuz (approved):
Ok. it's 500ms - so it doesn't matter.
A default board settings file might be interesting for the first connect when we don't know what kind of board it is yet.

PR #1236, opened 2023-06-03
Add missing pins A6 and A7 to nano

### LINKED ISSUE #1228: Increase analog input support in Pro Micro and Nano
The Pro Micro and Nano boards have 9 and 8 analog capable pins respectively, but are limited in Mobiflight to only 5 and 6 analog inputs.  I propose to increase the Mobiflight limit to each board's true capacities.  

Increase the number of analog inputs in Pro Micro to 9.
Increase the number of analog inputs in Nano to 8.

Only alternative is to use more boards.  This is not feasible in throttle quadrants where space is typically limited and use of many potentiometers is needed.

**Additional context**
It seems potentiometers are very pin efficient, just requiring one pin to function.  There are ways to use more analog input in rotary switches and groups of buttons, attaching resistors in series.

neilenns:
Fixes #1228 

Adds the missing A6 and A7 pins to the nano board file and increases the max number of analog pins to 8.

While the two pins work without a firmware change, the firmware also needs to update to have a max analog pin count of 8 for all the pins to be usable at the same time (see https://github.com/MobiFlight/MobiFlight-Connector/issues/1228)

### DIFF
--- a/Boards/arduino_micro.board.json
+++ b/Boards/arduino_micro.board.json
@@ -135,12 +135,6 @@
       "Pin": 16
     },
     {
-      "isAnalog": false,
-      "isI2C": false,
-      "isPWM": false,
-      "Pin": 17
-    },
-    {
       "isAnalog": true,
       "isI2C": false,
       "isPWM": false,
--- a/Boards/arduino_micro.board.json
+++ b/Boards/arduino_micro.board.json
@@ -32,7 +32,7 @@
     "ResetFirmwareFile": "reset.arduino_promicro_1_0_2.hex"
   },
   "ModuleLimits": {
-    "MaxAnalogInputs": 5,
+    "MaxAnalogInputs": 9,
     "MaxButtons": 18,
     "MaxEncoders": 9,
     "MaxInputShifters": 2,
@@ -73,7 +73,8 @@
       "isAnalog": true,
       "isI2C": false,
       "isPWM": false,
-      "Pin": 4
+      "Pin": 4,
+      "Name": "A6"
     },
     {
       "isAnalog": false,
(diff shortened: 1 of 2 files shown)


elral:
@neilenns could you add the ProMicro on this PR to increase the max. numbers of Analog Input to 9 as this PR is still open?

I have issued a PR for the firmware change on Nano and ProMicro, see PR [248](https://github.com/MobiFlight/MobiFlight-FirmwareSource/pull/248).

neilenns:
@elral Updated the PR with the Pro Micro max increased to 9, and also added names to the analog pins that didn't have analog names attached.

I used this for reference: https://learn.sparkfun.com/tutorials/pro-micro--fio-v3-hookup-guide

elral:
I double checked the pin assignemt.

The ProMicro does not have Pin D17, but this one is on the json file.
The "original" board is just named "Micro", and this board has additional pins (11/12/13/17/22/23).
We could
* add these additional pins to the json file. If one isn't available on the board, it can not be connected
* "just" support the ProMicro, but then we have to delete Pin 17.

@neilenns @DocMoebiuz Any idea how to handle this?

The Nano has Pin 20 and 21, but they can only be used as Anlog In. I guess we do not have a marker to avoid that they show up as digital in. Furthermore these both pins are marked as I2C in this PR, but they aren't.

neilenns:
* Removed pin D17 from the ProMicro file.
* Fixed the incorrect I2C marking for A6 and A7 on the nano

Regarding Micro vs. Pro Micro, Mobiflight website says we support the *pro micro* so the file should reflect that. you are correct, there is no way currently to avoid having the analog pins show up as digital.

DocMoebiuz:
@neilenns i saw you added analog pins to the Pro Micro, I looked at the sparkfun pinout
image

And I was confused by the labels and the analog pins that they show compared what our JSON file has.

I always tried to name the labels for the pins the same as they are named on the board silkscreen.
Looks like we have a mistake in our json - the analog pins and pin numbers are not the same as on the pinout.

PR #1239, opened 2023-06-05
Build the connector by default

neilenns:
Add MobiFlight-Connector to the build configuration manager (debug and release)


DocMoebiuz (commented):
So what does this acutally mean? You say that AnyCpu shall use the settings from x86? Is that what we want? 

The ArcazeHid.dll is at least 32-bit only. I think if you compile with AnyCPU then your fine as long as you don't activate Arcaze support which will try to load the 32-bit DLL and then fail.

I think the right fix for this is to remove "AnyCPU" alltogether. First time checkout will be x86 and this will work fine.

DocMoebiuz (changes requested):
See last comment.

neilenns:
All I did was check a box in the UI to say build it. I didn't change any of the CPU things, those values are what they were before I checked the box.

DocMoebiuz (approved):

PR #1241, opened 2023-06-08
Add a test for bad CH340 chips

### LINKED ISSUE #1410: Add a warning for counterfeit CH340 chips
Right now when people have board connection issues due to counterfeit CH340 chips the only way for us to find out is to ask them to send photos of the boards.

We should log an error when a counterfeit board is detected so we can just have users give us logs instead of photos.

neilenns:
Fixes #1410

Added a check of the device's reported name against the garbled name for the known bad CH340 chips and log an error.

Also cleaned up two style warnings for the method I was touching.

### DIFF
--- a/MobiFlight/MobiFlightCache.cs
+++ b/MobiFlight/MobiFlightCache.cs
@@ -178,6 +178,17 @@ namespace MobiFlight
             return Modules.Values;
         }
 
+        private static string GetDeviceProperty(ManagementObject managementObject, string propertyName)
+        {
+            var args = new object[] { new string[] { "DEVPKEY_Device_BusReportedDeviceDesc" }, null };
+            var r = managementObject.InvokeMethod("GetDeviceProperties", args);
+            var mbos = (ManagementBaseObject[])args[1];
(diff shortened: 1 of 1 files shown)


MobiFlight/MobiFlightCache.cs
```diff
@@ -188,7 +211,7 @@ private static List<PortDetails> getSupportedPorts()
             try
             {
                 ManagementObjectSearcher searcher = new ManagementObjectSearcher("root\\CIMV2", "SELECT * FROM Win32_PnPEntity WHERE ClassGuid=\"{4d36e978-e325-11ce-bfc1-08002be10318}\"");
-                foreach (ManagementObject queryObj in searcher.Get())
+                foreach (ManagementObject queryObj in searcher.Get().Cast<ManagementObject>())
```
neilenns, line 214: Style cleanup as recommended by VS.

DocMoebiuz:
@neilenns did this really work? PR is in Draft and I am wondering if we are going to include this or if we close it.

neilenns:
It works. It is up to you if you want to include it.

Issue #514, opened 2021-10-04
Support 4-digit TM1637-based displays


neilenns:
**Is your feature request related to a problem? Please describe.**
Support displays like this: https://www.amazon.com/s?k=TM1637+display&ref=nb_sb_noss_2

**Describe the solution you'd like**
Arduino library: https://github.com/avishorp/TM1637

**Describe alternatives you've considered**
MAX7219s

**Additional context**
This has come up twice in the last couple of weeks in Discord.

MobiFlight-Admin:
Max7219 are darn cheap and available to everyone. What are the benefits of TM1637?

PR #1252, opened 2023-07-10
Added support for TM1637

GioCC:
Requires firmware including modified display library "LedControl_Dual" (branch: MOD_TM1637_DualLib).
Solves [closed] issue #514 (changed to discussion #604.


GioCC:
An important point posted in the Discord channel:
I expect the max number of configurable displays to be a somehow controversial point.
RAM is not actually a constraint in this case; the number could be easily extended to 6 or more.
The real conundrum is that MAXes are chainable, so having 4 "devices" (Mega and Pico; 1 for other boards) really means having 4 _chains_ of 4 physical modules each, which is plenty. Even for other boards, where the limit is 1, 4 physical modules are usually enough.
The problem in extending this number for MAXes is that the data transfer time becomes _very_ long; for _every single digit_ to be changed, a very long sequence must be transmitted, so IMHO even a 4-device chain should be preferably avoided, let alone having many of them.
On the other hand, TMs can't be chained (so they don't have the above issue), therefore a higher maximum number of devices - I'd say 8 or more - is practically compulsory; but it's apparent that allowing such a high number of displays whatever the type could cause unpleasant problems to the unaware user.
I haven't found an answer to this problem as of yet, so I would gladly accept any idea or suggestion.

elral:
> An important point posted in the Discord channel: I expect the max number of configurable displays to be a somehow controversial point. RAM is not actually a constraint in this case; the number could be easily extended to 6 or more. The real conundrum is that MAXes are chainable, so having 4 "devices" (Mega and Pico; 1 for other boards) really means having 4 _chains_ of 4 physical modules each, which is plenty. Even for other boards, where the limit is 1, 4 physical modules are usually enough. The problem in extending this number for MAXes is that the data transfer time becomes _very_ long; for _every single digit_ to be changed, a very long sequence must be transmitted, so IMHO even a 4-device chain should be preferably avoided, let alone having many of them. On the other hand, TMs can't be chained (so they don't have the above issue), therefore a higher maximum number of devices - I'd say 8 or more - is practically compulsory; but it's apparent that allowing such a high number of displays whatever the type could cause unpleasant problems to the unaware user. I haven't found an answer to this problem as of yet, so I would gladly accept any idea or suggestion.

I would increase the max number of displays to 6 (or 8?) so that more than 4 TM1637 displays can be connected. For the MAX displays I would give a clear statement in the wiki, that the max numbers of devices should be limited. For now we can connect up to 4 x 4(daisy chained) = 16 displays. This would mean that the limitation should be still 16 displays (what I think it could be still too much), that could be 8 devices each daisy chained with 2 displays, or 6 devices with 4 devices and 3 daisy chained plus 2 devices with 2 daisy chained. According the formula devices multiplied with daisy chained displays must not exceed 16.
At least for the MAX displays we are backwards compatible.

I still do not now how many displays would be sensefull on 1 board. I see it in the way that 1 board should be used for one instrument (COM, NAV, ...). So even 16 would be much to high, but haven't figured out what the max numbers would be by using this logic

DocMoebiuz:
This PR was superseded by #1325

Thank You @GioCC

PR #1257, opened 2023-07-13
Displays now work with more than 6 Max7219 modules

### LINKED ISSUE #1256: Problem with more than 6 Max7219 modules
**Describe the bug**
When you have more than 6 max7219 modules connected to MobiFlight, then module 7 & 8 might not work consistently. 
You notice it on "Stop" of MobiFlight because the last two will still stay on. Or displays are garbled.

Smells like: the amount of messages (burst) probably provokes a buffer overflow on the arduino side.

**Steps To Reproduce**
1. Add 8 max chips to your config
2. Create an individual output config for each
3. Hit the Run button
4. Hit the Stop button
5. See error

**Expected behavior**
All displays are turned off

**Actual behavior**
The last two displays are still showing numbers.

**Details**
 - MobiFlight Version: 9.7.1

DocMoebiuz:
fixes #1256

### DIFF
--- a/MobiFlight/MobiFlightLedModule.cs
+++ b/MobiFlight/MobiFlightLedModule.cs
@@ -92,6 +92,7 @@ namespace MobiFlight
             Log.Instance.log($"Command: SetModule <{(int)MobiFlightModule.Command.SetModule},{this.ModuleNumber},{subModule},{value},{points},{mask};>.", LogSeverity.Debug);
 
             // Send command
+            System.Threading.Thread.Sleep(1);
             CmdMessenger.SendCommand(command);
         }
 


DocMoebiuz:
@neilenns - can you confirm that it fixes the issue for you too?

DocMoebiuz:
I would not consider this a final fix, rather a workaround / quick fix. I think the CmdMessenger should have some throttling built in so that the Arduinos have the chance to catch up with the Serial Input (if that is really the root cause).

neilenns:
I'll give it a try later today

elral:
> I would not consider this a final fix, rather a workaround / quick fix. I think the CmdMessenger should have some throttling built in so that the Arduinos have the chance to catch up with the Serial Input (if that is really the root cause).

In this case the command messenger has to know which time is required for each output device and wait accordingly until the next command is send. Additionally it also depends if it's an Arduino or Pico.
Another solution could be that each output device sends a confirmation after execution.
Or the command messenger reports that the buffer is filled. But then an additional report is required that the buffer can be filled again. Might be too much effort.

DocMoebiuz:
so maybe it is even ok to deal with it on the mobiflight code side - but instead of always sleeping for 1ms only do it when there is a burst of messages

Issue #1219, opened 2023-05-15
Exception on selecting item in input type ComboBox. ALT Btn


JaimeLeon2:
**Describe the bug**
Error message appears in the log when opening any input config
`Error	05/15/2023 16:48:45	Exception on selecting item in input type ComboBox. ALT Btn
`

**Steps To Reproduce**
1. Mobiflight in logging mode, Info Level.
2. open any input config
3. close input config
4. see error generated in the log

**Expected behavior**
No error should be generated

**Actual behavior**
Consistent error generated with opening every input config.

**Details**
 - MobiFlight Version: 9.7.1.1
 - FlightSim: irrelevant
 - AddOn: irrelevant

**Additional context**
Several users have asked in the Discord server about very similar errors.

![image](https://github.com/MobiFlight/MobiFlight-Connector/assets/86563553/025a611d-54d9-4195-8795-7001898c1e72)

PR #1273, opened 2023-08-13
First refactor of Module, Joystick, Midi: Use common device interface

Koseng:
Solves #1219 and is a first step in bringing Mobiflight-Modules, Joysticks and MidiBoards closer together via a common IBaseDevice interface.

In #1219 the issue was different usage in regard to ToString() overload in devices of Mobiflight-Modules, Joysticks and MidiBoards. On MidiBoards and Joysticks it was used for the Label/Name system. On Modules it is already used differently. Based on the ToString() pattern it is not possible to resolve the issues. The problem was introduces with the Label/Name system.

_Solution:_ 
Do not use any more device.ToString() overload for InputTypeCombobox at all. Use a common device interface.

**Changelog:**

- Introduced new common interface IBaseDevice
- Moved JoystickDevice and JoystickDeviceType to separate files
- Replaced JoystickDeviceType with general DeviceType in Joystick like in Module-BaseDevice and MidiBoardDevice
     - JoystickDeviceType only necessary for InputDefnitions in JoystickInput
- Removed unused OnAxisChanged in Joystick
- Removed initial set strings in InputConfigWizard inputTypeCombobox (Pin, Display Module, BCD4056). Why do/did they exist?
    - Side effect? @DocMoebiuz  
- Simplified InputConfigWizard - determineCurrentDeviceType
- Introduced new ComboBoxHelper SetSelectedDeviceByDeviceName. Use that to set selected item in InputTypeComboBox.
- Reverted ComboBoxHelper.SetSelectedItem to ComboBoxHelper.SetSelectedItemByValue in DisplayPanel. => That was a bug and also led to an error message in logging.
 
**Discussion:**

Keep ListItem stuff in JoystickDevice and Joystick Classes or not?  Or better handle it outside of those classes? That means handle it like in Modules and MidiBoard or handle it like in Joystick?
- I am unsure what is better. Why offering especially ListItem in Joystick class? What if another control does need another type?
- On the other hand it does make the usage code more compact. But it is only used once in the codebase.
- My gut feeling is that we should remove the ListItem stuff in Joystick class. What if we move to another ui framework and there something else is needed?

What do you guys think @neilenns, @DocMoebiuz?

Depending on that decision the ListItem handling of Modules, Joysticks and MidiBoards should be unified in a followup PR.


MobiFlight/Joysticks/JoystickManager.cs
```diff
@@ -154,8 +154,7 @@ public void Connect(IntPtr Handle)
                 Log.Instance.log($"Adding attached joystick device: {d.InstanceName} Buttons: {js.Capabilities.ButtonCount} Axis: {js.Capabilities.AxeCount}.", LogSeverity.Info);
                 js.Connect(Handle);
                 joysticks.Add(js);
-                js.OnButtonPressed += Js_OnButtonPressed;
-                js.OnAxisChanged += Js_OnAxisChanged;
```
neilenns, line 158: Why did OnAxisChanged get removed?

Koseng:
New common interface:

ibasedevice

MobiFlight/Joysticks/JoystickManager.cs
Koseng: If I was not misled, it is dead uncalled code.

Koseng:
And for my discussion point:

Like a) keep ListItem Handling in Joystick
discussion1

Or like b) do ListItem Handling outside of Joystick
discussion2

Koseng:
ok, that merge did not turn out as intended....
will redo

Koseng:
Created new PR.

Issue #1242, opened 2023-06-12
Use keyboard keys for stepper calibration 


DocMoebiuz:
**Is your feature request related to a problem? Please describe.**
When calibrating stepper zero position with Mobiflight you have to use the mouse to move the stepper and select amount of steps.

This requires that you can look at the screen, use the mouse and see the gauge - all at the same time.

**Describe the solution you'd like**
Being able to control the stepper steps and moving the stepper by using the keyboard.

Eg. arrow keys up and down increase/decrease amount of steps slider - and arrow keys left right make the stepper move.

**Additional context**
Brought up on discord: https://discord.com/channels/608690978081210392/1117639821326819369

CaptainBobSim:
Maybe Enter could set it in place?

DocMoebiuz:
Implemented by #1292

PR #1292, opened 2023-09-17
Add keyboard support for stepper calibration

simontrhodes:
Enhancement Issue #1242
Use keyboard keys for stepper calibration.
Arrow keys already worked but user needs to click on the trackbar control for keyboard to work. 
Added extra keys:
Enter key or 0 keys will zero.
Space bar will move the stepper motor.
Fixed spelling mistakes.

### DIFF
--- a/UI/Panels/Output/StepperPanel.cs
+++ b/UI/Panels/Output/StepperPanel.cs
@@ -136,17 +136,17 @@ namespace MobiFlight.UI.Panels
         {
             if (!(sender as Control).Parent.Enabled) return;
 
-            String value = ((TextBox)sender).Text.Trim();
+            String value = (sender as TextBox).Text.Trim();
 
             if (value == "") e.Cancel = true;
             if (e.Cancel)
             {
-                displayError((Control)sender, i18n._tr("uiMessagePanelsStepperInputRevolutionsMustNonEmpty"));
+                displayError(sender as Control, i18n._tr("uiMessagePanelsStepperInputRevolutionsMustNonEmpty"));
                 return;
             }
             else
             {
-                removeError((Control)sender);
+                removeError(sender as Control);
             }
 
             try
@@ -159,18 +159,18 @@ namespace MobiFlight.UI.Panels
             catch (Exception ex)
             {
                 e.Cancel = true;
-                displayError((Control)sender, i18n._tr("uiMessageValidationMustBeNumber"));
+                displayError(sender as Control, i18n._tr("uiMessageValidationMustBeNumber"));
                 return;
             }
 
             if (e.Cancel)
             {
-                displayError((Control)sender, i18n._tr("uiMessagePanelsStepperInputRevolutionsMustBeGreaterThan0"));
+                displayError(sender as Control, i18n._tr("uiMessagePanelsStepperInputRevolutionsMustBeGreaterThan0"));
                 return;
             }
             else
             {
-                removeError((Control)sender);
+                removeError(sender as Control);
             }
         }
 
--- a/UI/Panels/Output/StepperPanel.cs
+++ b/UI/Panels/Output/StepperPanel.cs
@@ -11,6 +11,7 @@ namespace MobiFlight.UI.Panels
         public event EventHandler<ManualCalibrationTriggeredEventArgs> OnManualCalibrationTriggered;
         public event EventHandler<StepperConfigChangedEventArgs> OnSetZeroTriggered;
         public event EventHandler<StepperConfigChangedEventArgs> OnStepperSelected;
+        
         protected StepperProfilePreset StepperProfile;
 
         int[] StepValues = { -50, -10, -1, 1, 10, 50 };
@@ -67,7 +68,7 @@ namespace MobiFlight.UI.Panels
             stepperAddressesComboBox.SelectedValue = value;
         }
 
-        public void SetAdresses(List<ListItem> pins)
+        public void SetAddresses(List<ListItem> pins)
         {
             stepperAddressesComboBox.DataSource = new List<ListItem>(pins);
             stepperAddressesComboBox.DisplayMember = "Label";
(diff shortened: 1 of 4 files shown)


tigert:
This is cool, but very very undiscoverable! :-)

Would it be possible to do two things:

- A small text description to make it more obvious:

image

Also, is it possible to grab the key events when the stepper display tab is open, so that you can just do the adjustment without having to click the "Move steps" slider? Or at least this is what I had to do to make the shortcuts work?

UI/Panels/Output/StepperPanel.cs
```diff
@@ -84,40 +81,72 @@ public void SetAdresses(List<ListItem> pins)
...
+        }
+        
+        
         private void inputRevTextBox_Validating(object sender, CancelEventArgs e)
         {
-            if (!(sender as Control).Parent.Enabled) return;
+            if (!((Control)sender).Parent.Enabled) return;
 
-            String value = (sender as TextBox).Text.Trim();
+            var value = ((TextBox)sender).Text.Trim();
```
DocMoebiuz, line 139: We use the `object as specificType` syntax pretty much everywhere in mobiflight.
if the object is not of that type, it will be null.

You code will throw an exception. 

Please revert to old syntax.

UI/Panels/Output/StepperPanel.cs
```diff
@@ -84,40 +81,72 @@ public void SetAdresses(List<ListItem> pins)
...
 
             if (value == "") e.Cancel = true;
             if (e.Cancel)
             {
-                displayError(sender as Control, i18n._tr("uiMessagePanelsStepperInputRevolutionsMustNonEmpty"));
+                DisplayError((Control)sender, i18n._tr("uiMessagePanelsStepperInputRevolutionsMustNonEmpty"));
```
DocMoebiuz, line 144: Same.

Your code changes also are no changes. Please don't introduce such changes if we don't have a real benefit.

DocMoebiuz (changes requested):
There are many changes solely because of upper case vs. lower case changes. I am not a big fan of mixing this in a PR because it distracts so much from the actual changes.

I would prefer that you do these renamings in a separate PR where no further logic is changed.
Even the Designer is inconsistent with the Microsoft recommendations for property vs method camelCasing or Pascal casing.

Please simply have that in mind for your next PR. Thanks!

UI/Panels/Output/StepperPanel.cs
simontrhodes: Thank you.

UI/Panels/Output/StepperPanel.cs
simontrhodes: Yes, I understand.

UI/Panels/Output/StepperPanel.cs
```diff
@@ -84,40 +81,72 @@ public void SetAdresses(List<ListItem> pins)
...
+            }
+        }
+        
+        
         private void inputRevTextBox_Validating(object sender, CancelEventArgs e)
         {
             if (!(sender as Control).Parent.Enabled) return;
 
-            String value = (sender as TextBox).Text.Trim();
+            String value = ((TextBox)sender).Text.Trim();
```
DocMoebiuz, line 139: The changes are still present.

UI/Panels/Output/StepperPanel.cs
```diff
@@ -84,40 +81,72 @@ public void SetAdresses(List<ListItem> pins)
...
             if (e.Cancel)
             {
-                displayError(sender as Control, i18n._tr("uiMessagePanelsStepperInputRevolutionsMustNonEmpty"));
+                displayError((Control)sender, i18n._tr("uiMessagePanelsStepperInputRevolutionsMustNonEmpty"));
                 return;
             }
             else
             {
-                removeError(sender as Control);
+                removeError((Control)sender);
```
DocMoebiuz, line 149: The changes are still present.

UI/Panels/Output/StepperPanel.cs
```diff
@@ -130,18 +159,18 @@ private void inputRevTextBox_Validating(object sender, CancelEventArgs e)
             catch (Exception ex)
             {
                 e.Cancel = true;
-                displayError(sender as Control, i18n._tr("uiMessageValidationMustBeNumber"));
+                displayError((Control)sender, i18n._tr("uiMessageValidationMustBeNumber"));
```
DocMoebiuz, line 162: The changes are still present.

UI/Panels/Output/StepperPanel.cs
```diff
@@ -130,18 +159,18 @@ private void inputRevTextBox_Validating(object sender, CancelEventArgs e)
...
                 return;
             }
 
             if (e.Cancel)
             {
-                displayError(sender as Control, i18n._tr("uiMessagePanelsStepperInputRevolutionsMustBeGreaterThan0"));
+                displayError((Control)sender, i18n._tr("uiMessagePanelsStepperInputRevolutionsMustBeGreaterThan0"));
```
DocMoebiuz, line 168: The changes are still present.

UI/Panels/Output/StepperPanel.cs
```diff
@@ -130,18 +159,18 @@ private void inputRevTextBox_Validating(object sender, CancelEventArgs e)
...
                 return;
             }
             else
             {
-                removeError(sender as Control);
+                removeError((Control)sender);
```
DocMoebiuz, line 173: The changes are still present.

DocMoebiuz (changes requested):
See comments

UI/Panels/Output/StepperPanel.cs
simontrhodes: Sorry, changed back now.

UI/Panels/Output/StepperPanel.cs
simontrhodes: Sorry, changed back now.

UI/Panels/Output/StepperPanel.cs
simontrhodes: Sorry, changed back now.

neilenns:
@simontrhodes Can you update the title of this to be clearer? Maybe "Add keyboard support for stepper calibration" or somthing?

DocMoebiuz (approved):
Thanks!

PR #1295, opened 2023-09-17
Use a combination of PR number and run count for the connector version.

### LINKED ISSUE #1294: GitHub CI should make smarter filenames for zips
So I find the automatic zip creation super useful for testing PR's, but there are two things that bug me and / or would be helpful to change slightly.

- The zip file is wrapped inside another zip file, which is a bit strange and tedious to unpack

- The zip contains a folder path with some kind of an ID ("MobiFlightConnector-0.0.0-482" for example) - it would be nicer if that was the pull request number, in case you are testing a few different PR's at the same time. 

This would make it easier to delete old test builds when you can see the pull request number in the folder names.

We likely need to have some kind of an id on the zip and folder name, to differentiate between updated zips in the same pull request in case the PR branch gets more commits, but we could still have the id in the end - like maybe MobiFlightConnector-PR#1242-001 or something like that where the number increments.
 
So ideally we could have something like MobiFlightConnector-PR1234-001.zip that you could just right click -> Extract all, and it would default to:

C:\Downloads\MobiFlight-Testing\MobiFlightConnector-PR1234-001\

.. so you could just replace C:\Downloads\ with D:\ for example and have all test zips unpacked neatly in one folder you can clean up easily when needed.

neilenns:
Fixes #1294 

* Updates the build version to a combination of the PR number and the run number (which increments on every PR run). The build version winds up being `0.0.<prnumber>.<runnumber>`.
* Change the artifact upload to only upload that specific zip file, instead of a wildcard of everything in the release folder.
* Update the checkout action to use v3 to get rid of a warning

There doesn't seem to be a way to tell the artifact uploader to only upload a single file instead of zipping all the artifacts, so no way to avoid the zip-inside-a-zip problem.

### DIFF
--- a/MobiFlightConnector.csproj
+++ b/MobiFlightConnector.csproj
@@ -1690,7 +1690,7 @@
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <PropertyGroup>
-    <PostBuildEvent>if "$(ConfigurationName)" == "Release" ($(ProjectDir)Build\release.bat $(TargetDir) $(SolutionDir) 7.1.4)</PostBuildEvent>
+    <PostBuildEvent>if "$(ConfigurationName)" == "Release" ($(ProjectDir)Build\release.bat $(TargetDir) $(SolutionDir))</PostBuildEvent>
   </PropertyGroup>
   <PropertyGroup>
     <PreBuildEvent>xcopy /R /Y "$(ProjectDir)Lib\vJoyInterface.dll" "$(TargetDir)"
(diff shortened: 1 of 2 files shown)


tigert:
I am waiting for the checks to run :-)

MobiFlightConnector.csproj
```diff
@@ -1690,7 +1690,7 @@
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
   <PropertyGroup>
-    <PostBuildEvent>if "$(ConfigurationName)" == "Release" ($(ProjectDir)Build\release.bat $(TargetDir) $(SolutionDir) 7.1.4)</PostBuildEvent>
```
neilenns, line 1693: This isn't strictly required for this request, but I noticed it when working on the change. Doesn't seem like that version number needs to be there at all, the `release.bat` script only takes two parameters.

DocMoebiuz: you are right. this was like this in an early version of the script

DocMoebiuz (approved):
Cool!

PR #1299, opened 2023-09-22
Fix input config error log message 

### LINKED ISSUE #1219: Exception on selecting item in input type ComboBox. ALT Btn

DocMoebiuz:
fixes #1219 

- [x] Use ComboBoxHelper.SetSelectedItem
- [x] Use device names consistently in error messages

### DIFF
--- a/UI/Dialogs/InputConfigWizard.cs
+++ b/UI/Dialogs/InputConfigWizard.cs
@@ -286,7 +286,7 @@ namespace MobiFlight.UI.Dialogs
             }
 
             // second tab
-            if (!ComboBoxHelper.SetSelectedItemByValue(inputTypeComboBox, config.Name))
+            if (config.Name!=null && !ComboBoxHelper.SetSelectedItem(inputTypeComboBox, config.Name))
             {
                 // TODO: provide error message
                 Log.Instance.log($"Exception on selecting item in input type ComboBox. {config.Name}", LogSeverity.Error);
@@ -516,7 +516,7 @@ namespace MobiFlight.UI.Dialogs
                 }
 
                 // third tab
-                if (!ComboBoxHelper.SetSelectedItem(inputTypeComboBox, config.Name))
+                if (config.Name != null && !ComboBoxHelper.SetSelectedItem(inputTypeComboBox, config.Name))
                 {
                     Log.Instance.log($"Problem setting input type ComboBox. {config.Name}", LogSeverity.Error);
                 }
--- a/UI/Panels/Output/DisplayShiftRegisterPanel.cs
+++ b/UI/Panels/Output/DisplayShiftRegisterPanel.cs
@@ -29,9 +29,9 @@ namespace MobiFlight.UI.Panels
             // pre-select display stuff
             if (config.ShiftRegister != null && config.ShiftRegister.Address != null)
             {
-                if (!ComboBoxHelper.SetSelectedItem(shiftRegistersComboBox, config.ShiftRegister.Address.ToString()))
+                if (!ComboBoxHelper.SetSelectedItem(shiftRegistersComboBox, config.ShiftRegister.Address))
                 {
-                    Log.Instance.log("Exception on selecting item in Shift Register ComboBox.", LogSeverity.Error);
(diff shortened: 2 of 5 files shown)


Koseng:
Hi @DocMoebiuz , I suppose the changes regarding the ComboBox in InputConfigWizard.cs brake something and probably should be undone.

The whole reason I started the refactoring in #1273 was to solve the issue of #1219.

So far the behavior of arduino-modules, joysticks and midi-boards is different. So far only everyhing works because both ComboBoxHelpers are called in the process of opening the inputconfigWizard. SetSelectedItemByValue and SetSelectedItem. In some cases SetSelectedItemByValue is needed and then SetSelectedItem creates the error message and vice versa. Replacing one with the other will brake something (either for modules, joysticks or midiboards). The problem is, the arduino modules have so far no name/label handling.

I was annoyed by that error message, but it turned out, it could not be solved easily. Therefore the refactoring, which solves that issue and which is a first step in unifiying modules, joysticks and midiboards. There are more steps I would like to do, but one step after another.

DocMoebiuz:
i wanted this fix to solve the problem and then do the refactoring afterwards.

I had not the time to comment on your PR which i find is a great idea - I actually believe we might be able to use basedevice for all kinds of devices and derice specific classes from that - i saw that BaseDevice was only used for Arduino devices, not joysticks and midi devices

Koseng:
Sebastian, you did not understand what I wrote ;-).

With the change in the PR you created a big problem. Now I made the effort to test what you broke. You removed the error message, but made all joysticks and midiboards with labels unusable.

Try it with the bravo throttle quadrant. If you edit now an existing bravo throttle input with that the current PR [MobiFlightConnector.zip](https://github.com/MobiFlight/MobiFlight-Connector/suites/16440322449/artifacts/939494412), you always get an empty input config wizard. See result in the screenshot.

This issue is difficult to fix. **The main point of the first refactoring was to be able to fix this exact problem**. Without the refactoring the problem cannot be fixed. It needs bigger changes.

Regarding the refactoring, yes I also have many more ideas what we can improve, but I wanted to do small steps.

image

Koseng:
More detailed Explanation what is the problem with the selection in the InputTypeComboBox:

#### Bravo Throttle Quadrant example:

`ListItem.Label: AP-ALT`
`ListItem.Value: Button 5`  => (corresponds to **config.name**)


#### (Arduino) Module Example:
`ListItem.Label: Button4`   =>  (corresponds to **config.name**)
`ListItem.Value: Button:Button4 Pin:4`


#### ComboBoxHelper

`ComboboxHelper.SetSelectedItem`   === Selects Item by LABEL
`ComboBoxHelper.SetSelectedItemByValue`  === Selects Item by VALUE (**executes item.ToString()**)


### RESULT:
- For joysticks with differing Labels from Names `ComboBoxHelper.SetSelectedItemByValue` must be used
- For arduino modules `ComboboxHelper.SetSelectedItem` must be used
- And now we are in a deadlock. 
- Before that PR it only worked because BOTH were called. But only one worked and the other one throwed an error message.
- Straight forward would be to always use `ComboBoxHelper.SetSelectedItemByValue`. But that cannot be used easily with the arduino modules, because the toString() overload is already used for other purposes.

=> The current design problems are in different usage of the toString() overload in arduino modules and in joysticks.
=> To select the right element we should not use toString() at all. We should be more explicit.
=> #1273 addresses that with the introduction of  `ComboBoxHelper.SetSelectedDeviceByDeviceName` and the refactoring.

Issue #1305, opened 2023-10-04
Use Mobiflight on several PCs in network connected to one MSFS instance


Koseng:
**Is your feature request related to a problem? Please describe.**
Use Mobiflight on several PCs in network connected to one MSFS instance

**Describe the solution you'd like**
Run mobiflight on several pcs with adjusted simconnect config.
https://docs.flightsimulator.com/html/Programming_Tools/SimConnect/SimConnect_CFG_Definition.htm

To achieve that, the multiclient capabilities of mobiflight wasm need to be used.

PR #1306, opened 2023-10-04
Use multiclient wasm interface

Koseng:
Resolves: #1305 
Two wasm client configurations are introduced.

- InitClient +
- RuntimeClient

InitClient is only used to establish the first connection and register the RuntimeClient. Then RuntimeClient is used for all operations.

So far the code is still under the assumption, that the simconnect connection might be available before the wasm was initialized. 
I doubt that can happen, in the devmode console logging wasm is initialized long before first simconnect operations. 

To clarify I opened the ticket:
https://devsupport.flightsimulator.com/t/is-standalone-wasm-initialized-before-external-simconnect-becomes-available/6535


DocMoebiuz (approved):
Thanks! Really great changes

PR #1310, opened 2023-10-07
First refactor of Module, Joystick, Midi: Use common device interface

Koseng:
Merged with Main.

Solves #1219 and is a first step in bringing Mobiflight-Modules, Joysticks and MidiBoards closer together via a common IBaseDevice interface.

In #1219 the issue was different usage in regard to ToString() overload in devices of Mobiflight-Modules, Joysticks and MidiBoards. On MidiBoards and Joysticks it was used for the Label/Name system. On Modules it is already used differently. Based on the ToString() pattern it is not possible to resolve the issues. The problem was introduces with the Label/Name system.

_Solution:_ 
Do not use any more device.ToString() overload for InputTypeCombobox at all. Use a common device interface.

**Changelog:**

- Introduced new common interface IBaseDevice
- Moved JoystickDevice and JoystickDeviceType to separate files
- Replaced JoystickDeviceType with general DeviceType in Joystick like in Module-BaseDevice and MidiBoardDevice
     - JoystickDeviceType only necessary for InputDefnitions in JoystickInput
- Removed unused OnAxisChanged in Joystick
- Removed initial set strings in InputConfigWizard inputTypeCombobox (Pin, Display Module, BCD4056). Why do/did they exist?
    - Side effect? @DocMoebiuz  
- Simplified InputConfigWizard - determineCurrentDeviceType
- Introduced new ComboBoxHelper SetSelectedDeviceByDeviceName. Use that to set selected item in InputTypeComboBox.
- Reverted ComboBoxHelper.SetSelectedItem to ComboBoxHelper.SetSelectedItemByValue in DisplayPanel. => That was a bug and also led to an error message in logging.

New common interface:

ibasedevice


MobiFlight/Joysticks/Joystick.cs
```diff
@@ -157,7 +121,7 @@ protected virtual void EnumerateDevices()
                         Log.Instance.log($"Axis can't be mapped: {DIJoystick.Information.InstanceName} Aspect: {aspect} Offset: {offset} Usage: {usage} Axis: {name} Label: {axisLabel}.", LogSeverity.Error);
                         continue;
                     }
-                    Axes.Add(new JoystickDevice() { Name = axisName, Label = axisLabel, Type = JoystickDeviceType.Axis });
+                    Axes.Add(new JoystickDevice() { Name = axisName, Label = axisLabel, Type = DeviceType.AnalogInput });
```
DocMoebiuz, line 124: I don't like that we lose the specifics of a Joystick. An axis inherits from Analog Input maybe. It is an Enum right now, i know. but I find it sad that we use the base types here. It could be a AxisJoystickDevice.

MobiFlight/Joysticks/Joystick.cs
```diff
@@ -167,19 +131,19 @@ protected virtual void EnumerateDevices()
                     // when looking up names in the the .joystick.json file.
                     var buttonName = $"{ButtonPrefix} {device.Usage}";
                     var buttonLabel = MapDeviceNameToLabel(buttonName);
-                    Buttons.Add(new JoystickDevice() { Name = buttonName, Label = buttonLabel, Type = JoystickDeviceType.Button });
+                    Buttons.Add(new JoystickDevice() { Name = buttonName, Label = buttonLabel, Type = DeviceType.Button });
```
DocMoebiuz, line 134: That's fine. a button is a button, also on a Joystick.

MobiFlight/Joysticks/Joystick.cs
```diff
@@ -167,19 +131,19 @@ protected virtual void EnumerateDevices()
...
-                    POV.Add(new JoystickDevice() { Name = $"{PovPrefix} {name}UR", Label = $"{name} (↗)", Type = JoystickDeviceType.POV });
-                    POV.Add(new JoystickDevice() { Name = $"{PovPrefix} {name}R", Label = $"{name} (→)", Type = JoystickDeviceType.POV });
-                    POV.Add(new JoystickDevice() { Name = $"{PovPrefix} {name}DR", Label = $"{name} (↘)", Type = JoystickDeviceType.POV });
-                    POV.Add(new JoystickDevice() { Name = $"{PovPrefix} {name}D", Label = $"{name} (↓)", Type = JoystickDeviceType.POV });
-                    POV.Add(new JoystickDevice() { Name = $"{PovPrefix} {name}DL", Label = $"{name} (↙)", Type = JoystickDeviceType.POV });
-                    POV.Add(new JoystickDevice() { Name = $"{PovPrefix} {name}L", Label = $"{name} (←)", Type = JoystickDeviceType.POV });
-                    POV.Add(new JoystickDevice() { Name = $"{PovPrefix} {name}UL", Label = $"{name} (↖)", Type = JoystickDeviceType.POV });
+                    POV.Add(new JoystickDevice() { Name = $"{PovPrefix} {name}U", Label = $"{name} (↑)", Type = DeviceType.Button });
```
DocMoebiuz, line 139: I would really like to keep this detail. If we don't want to keep it as Type A POV button is special for Joysticks. PovJoysticDeviceType maybe as a solution for now.

MobiFlight/Joysticks/JoystickManager.cs
```diff
@@ -178,11 +177,6 @@ private bool HasAxisOrButtons(Joystick js)
                 js.Capabilities.ButtonCount > 0;
         }
 
-        private void Js_OnAxisChanged(object sender, InputEventArgs e)
```
DocMoebiuz, line 181: This should not get lost, the onButtonPress mapping should probably be done outside of the Joystick Manager in the Execution Manager.

DocMoebiuz (changes requested):
Everything else is looking really good. But these points are important to me.

MobiFlight/Joysticks/Joystick.cs
Koseng: Added extra JoystickDeviceType.

MobiFlight/Joysticks/JoystickManager.cs
Koseng: Added extra JoystickDeviceType.

MobiFlight/Joysticks/Joystick.cs
Koseng: Added extra JoystickDeviceType.

MobiFlight/Joysticks/Joystick.cs
Koseng: Added extra JoystickDeviceType.

DocMoebiuz (approved):
Thank YOU!

PR #1318, opened 2023-10-12
Add setting to control minimize on autorun

### LINKED ISSUE #1317: Add a setting to disable auto-minimize on auto-run
Right now if auto-run is enabled MobiFlight will always minimize to the system tray when it starts running. This is really annoying when you have MobiFlight open in another window because you want to see the logs, or are working to edit configurations.

A setting that disables the minimizing behaviour on auto-run.

Turning off auto-run, but then I lose the benefit of auto-run.

neilenns:
Fixes #1317 

* Rename an existing section of the general settings dialog to "Run options" so I don't have to add another section with a single setting in it
* Add a general setting "Minimize to system tray on AutoRun", which defaults to true to maintain existing behaviour by default
* Check that setting before attempting to minimize the app

image

@DocMoebiuz If you give me the German translation for that string I can add it to the string resources for the dialog

### DIFF
--- a/Properties/Settings.Designer.cs
+++ b/Properties/Settings.Designer.cs
@@ -12,7 +12,7 @@ namespace MobiFlight.Properties {
     
     
     [global::System.Runtime.CompilerServices.CompilerGeneratedAttribute()]
-    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.VisualStudio.Editors.SettingsDesigner.SettingsSingleFileGenerator", "17.3.0.0")]
+    [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.VisualStudio.Editors.SettingsDesigner.SettingsSingleFileGenerator", "17.5.0.0")]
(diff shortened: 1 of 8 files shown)


neilenns:
Just pushed a change to fix the tab order in the general settings as well.

DocMoebiuz:
> Fixes #1317
> 
> * Rename an existing section of the general settings dialog to "Run options" so I don't have to add another section with a single setting in it
> * Add a general setting "Minimize to system tray on AutoRun", which defaults to true to maintain existing behaviour by default
> * Check that setting before attempting to minimize the app
> 
> image
> @DocMoebiuz If you give me the German translation for that string I can add it to the string resources for the dialog

German translation: "Bei AutoRun in Task Leiste minimieren"

neilenns:
@DocMoebiuz Can I also get for "Run options"?

DocMoebiuz:
> @DocMoebiuz Can I also get for "Run options"?

"Run" Optionen

PR #1339, opened 2023-10-24
Update type number for MAX72xx devices

### LINKED ISSUE #1338: 7 segment displays don't work with beta 9.7.1.3
New 7 segment displays defined in the desktop app do not work with the beta 9.7.1.3. Displays configured with prior versions of the desktop app do.

From the Discord discussion:

> looks like the config written in the device by stable MF is : 4.2.4.3.15.1.LedModule:; <- this works
and the latest beta writes it as new type, likely due to the TM segments support: 16.1.2.4.3.15.1.LedModule: <- this does not work

neilenns:
Fixes #1338

### DIFF
--- a/MobiFlight/Config/LedModule.cs
+++ b/MobiFlight/Config/LedModule.cs
@@ -6,7 +6,7 @@ namespace MobiFlight.Config
 {
     public class LedModule : BaseDevice
     {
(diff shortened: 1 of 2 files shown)


DocMoebiuz (approved):
Thanks

PR #1366, opened 2023-10-28
Skip update check on pull request builds

### LINKED ISSUE #1365: Only make new version check if current major version is > 0
Current PR builds are great but they always suggest an upgrade and if you accidentally click OK (because who wouldnt want a NEW VERSION OF MOBIFLIGHT if it is available???!) then your PR folder is screwed and you need to download it again :-)

As suggested by @neilenns, check only if it is an actual release, PR builds are all 0.0.X.

neilenns:
Fixes #1365

Checks to see if the major version of the build is `0` and if so skips the update check.

```
Info	10/28/2023 12:38:24	Skipping update check since this is an unreleased build.
```

Since I was touching the lines close by anyway I also converted two old-style strings to new style string interpolation.

### DIFF
--- a/Base/AutoUpdateChecker.cs
+++ b/Base/AutoUpdateChecker.cs
@@ -17,8 +17,16 @@ namespace MobiFlight.UpdateChecker
             if (Properties.Settings.Default.CacheId == "0") Properties.Settings.Default.CacheId = Guid.NewGuid().ToString();
             String trackingParams = hash + "-" + Properties.Settings.Default.CacheId + "-" + Properties.Settings.Default.Started;
 
-            string CurVersion = Assembly.GetExecutingAssembly().GetName().Version.ToString();
-            string CommandToSend = "/check /version " + CurVersion + " /cacheId " + trackingParams;
+            var CurVersion = Assembly.GetExecutingAssembly().GetName().Version;
+            string CommandToSend = "/check /version " + CurVersion.ToString() + " /cacheId " + trackingParams;
+
+            // Issue 1365: Don't check for updates if the build came from a pull request. These builds are
+            // identified by the major version being 0.
+            if (CurVersion.Major == 0)
+            {
+                Log.Instance.log("Skipping update check since this is an unreleased build.", LogSeverity.Info);
+                return;
+            }
 
             if (Properties.Settings.Default.BetaUpdates)
             {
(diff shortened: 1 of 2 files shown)


Base/AutoUpdateChecker.cs
```diff
@@ -15,10 +15,18 @@ public static void CheckForUpdate(bool force = false, bool silent = false)
         {
             String hash = (Environment.UserName + Environment.MachineName).GetHashCode().ToString();
             if (Properties.Settings.Default.CacheId == "0") Properties.Settings.Default.CacheId = Guid.NewGuid().ToString();
-            String trackingParams = hash + "-" + Properties.Settings.Default.CacheId + "-" + Properties.Settings.Default.Started;
+            var trackingParams = $"{hash}-{Properties.Settings.Default.CacheId}-{Properties.Settings.Default.Started}";
 
-            string CurVersion = Assembly.GetExecutingAssembly().GetName().Version.ToString();
-            string CommandToSend = "/check /version " + CurVersion + " /cacheId " + trackingParams;
+            var CurVersion = Assembly.GetExecutingAssembly().GetName().Version;
+            var CommandToSend = $"/check /version {CurVersion} /cacheId {trackingParams}";
```
DocMoebiuz, line 21: this line could go under the block `if (CurVersion.Major == 0)` because it is only needed for the other cases.

DocMoebiuz (changes requested):
Just a minor thing. Very cool!

Base/AutoUpdateChecker.cs
neilenns: Done!

tigert:
works!

of course now if you want to work on the update checker feature, you got to remember this change :-)

neilenns:
@DocMoebiuz I went through and cleaned up the remaining three nag messages from Visual Studio in this file. Two fields get set to readonly which is easy. It also pointed out that the `force` parameter to the method never actually gets used, so I removed that which required tweaking the call to the method in MainForm.cs.

Base/AutoUpdateChecker.cs
```diff
@@ -9,16 +9,25 @@ namespace MobiFlight.UpdateChecker
 {
     static class AutoUpdateChecker
     {
-        static string mobiFlightInstaller = "MobiFlight-Installer.exe";
-        static int UpdateCheckTimeoutInMs = 5000;
-        public static void CheckForUpdate(bool force = false, bool silent = false)
+        static readonly string mobiFlightInstaller = "MobiFlight-Installer.exe";
+        static readonly int UpdateCheckTimeoutInMs = 5000;
+        public static void CheckForUpdate(bool silent = false)
```
DocMoebiuz, line 14: you missed one call to CheckForUpdate in line 610. It should remove true now entirely.

DocMoebiuz (changes requested):
See comment. THANKS!

Base/AutoUpdateChecker.cs
neilenns: Oooooops.

DocMoebiuz (approved):
Thanks!

Issue #1371, opened 2023-11-01
Feature request: Be able to configure different actions after short and long button press


Koseng:
**Is your feature request related to a problem? Please describe.**
Common user request and also a widely used feature in other software. The same button shall be used for two different actions, depending on whether it was short pressed or long pressed. Meaning whether the button release came quickly or slowly after the button press. One common usage is with a Button-Encoder for the Airbus autopilot panel. Usually Button-Encoders do not support a Push/Pull behavior like in the real plane. So the "Pull" is emulated with a long press and the "Push" is done with a short press.

It is not the same as #308 which focuses on repeating an action as long as a button is pressed.

**Describe the solution you'd like**
In the input config wizard for buttons, there should be a way to configure two different actions, depending on whether the button was released quickly or slowly.

**Current Workaround**
There is the current workaround, which probably is too complex for a lot of users.
https://github.com/MobiFlight/MobiFlight-Connector/wiki/MSFS2020-Short%2C-Long-and-Double-Press-Button-Configuration

**Additional context**
Many people, for example with the x-touch-mini, use this feature with other software.

PR #1372, opened 2023-11-01
Implement Button On Long Release

Koseng:
Solves #1371.

In Addition to "On Release", "On Long Release" is introduced. Currently the minimum Long Release delay is fix 350ms. In the future that can be made configurable.

For the most widely used scenario, https://github.com/MobiFlight/MobiFlight-Connector/wiki/MSFS2020-Short%2C-Long-and-Double-Press-Button-Configuration is not necessary any more.

For a "Short button press" - "On Release" can be used.
For a "Long button press" - "On Long Release" can be used.

If there is no defined "On Long Release", always "On Release" is executed.

**If someone tests the PR, please backup your configurations beforehand.**

image


Koseng:
Still need to check, whether all panel sizes are ok.

MobiFlight/InputConfig/ButtonInputConfig.cs
```diff
@@ -10,13 +8,15 @@ public class ButtonInputConfig : IXmlSerializable, ICloneable
...
         public InputAction onRelease;
+        public InputAction onLongRelease;
 
 
         public object Clone()
         {
             ButtonInputConfig clone = new ButtonInputConfig();
             if (onPress != null) clone.onPress = (InputAction)onPress.Clone();
             if (onRelease != null) clone.onRelease = (InputAction)onRelease.Clone();
+            if (onLongRelease != null) clone.onLongRelease = (InputAction)onLongRelease.Clone();
```
DocMoebiuz, line 19: Great you though about the unit tests.

https://github.com/MobiFlight/MobiFlight-Connector/blob/0944c0a61e394253a74654bda17383b30bc1e231/MobiFlightUnitTests/MobiFlight/InputConfig/ButtonInputConfigTests.cs#L27

We have to extend them for this new property accordingly.

MobiFlight/InputConfig/ButtonInputConfig.cs
```diff
@@ -69,20 +77,36 @@ public void WriteXml(System.Xml.XmlWriter writer)
             writer.WriteStartElement("onRelease");
             if (onRelease != null) onRelease.WriteXml(writer);
             writer.WriteEndElement();
+
```
DocMoebiuz, line 80: an option could be that we don't write the entire node if it is not set. Also true for the other nodes, would reduce the XML output.

In this case, we could even open the config in an old MF version in case no onReleaseLong is defined.

MobiFlight/InputConfig/ButtonInputConfig.cs
```diff
@@ -69,20 +77,36 @@ public void WriteXml(System.Xml.XmlWriter writer)
...
+            if (inputEvent == MobiFlightButton.InputEvent.PRESS && onPress != null)
             {
                 onPress.execute(cacheCollection, args, configRefs);
             }
-            else if (args.Value == 1 && onRelease != null)
+            else if (inputEvent == MobiFlightButton.InputEvent.RELEASE && onRelease != null)
             {
                 onRelease.execute(cacheCollection, args, configRefs);
             }
+            else if (inputEvent == MobiFlightButton.InputEvent.LONG_RELEASE)
```
DocMoebiuz, line 99: i would prefer to move the second if case to the other condition above... so we would have onRelease.execute only once.

When test become longer, i started to create a method for the test with a speaking name. e.g. ```IsLongReleaseButOnlyReleaseDefined()```

and this is implemented as 

```return inputEvent == MobiFlightButton.InputEvent.LONG_RELEASE && onLongRelease==null && onRelease!=null```

DocMoebiuz (changes requested):

MobiFlight/InputConfig/ButtonInputConfig.cs
Koseng: Done

MobiFlight/InputConfig/ButtonInputConfig.cs
Koseng: Done

MobiFlight/InputConfig/ButtonInputConfig.cs
Koseng: Done

DocMoebiuz (changes requested):
Thanks for all the effort with the unit tests.

Based on experience from the past, it would be great to have covered reading elements with:

- one onPress
- one onPress, one empty onRelease (backward compatibility)
- one onPress, one onLongPress (skipping one of the optional nodes in between)
- one empty <button /> element (this might be a theoretical state)

Koseng:
Test cases added.

DocMoebiuz (approved):
Thanks man!

PR #1374, opened 2023-11-04
Refactor Button Panel

Koseng:
Refactor `ButtonPanel` and remove all the long copy and paste switch cases. `ButtonPanel` is shrinked from 520 to 200 lines.

- All input panels now share the same interface `IPanelConfigSync`
- Input actions are mapped to input panels in `InputActionMapping.cs`
- In `ButtonInputConfig` two new methods are introduced:
   - `SetInputActionByName`
   - `GetInputActionByName`

If the refactoring is approved, it can also be applied to the other panels like `AnalogPanel` or `EncoderPanel`.

-> Now introducing a new tab or a new input panel is just a couple lines of code.

- [x] Resolve Issue of Setting Input Action back to "None".

PR #1383, opened 2023-11-09
Added configurable long button release and on hold with repeat

### LINKED ISSUE #308: Feature request: button repeat on long press
Use case: Many avionics like the G1000 have repeating button events for long press. Like, selecting the vertical speed or airspeed hold value \(there are two buttons instead of a rotary wheel\). The G1000 map joystick lets you pan the map when you deflect and hold the stick. An engine starter could invoke automatic start once you press and hold a button for a given time. Generally useful for cockpits with simple push buttons, for things that you do not want to do accidentally.

**Design**

‌

New tabs in InputConfigWizard: Long Press | Long Release

* Long press delay: if a button is held pressed for this long, the long press event is invoked. \(default: 250ms\)
* Repeat interval: \(default 125ms\)
* Action type:

    * Repeat “On Press” event \(default\)
    * Fire another event
    
        * similar config as in “On Press” tab
        
    * None
    

‌

[Issue created by tigert: 2021-02-13]
[Last updated on bitbucket: 2021-06-08]

[Comment created by tigert: 2021-06-08]
Been thinking about this a bit recently, and wanted to share some thoughts:

Basically we are talking about values that, when triggered, do not immediately change, but rather are passed to a “pipe” that manipulates the value over time in a given fashion, based on the input value. 

```
Simulator input --> [ filter: start value, end value, time, function ] --> Output
```

A **repeat** “filter” would have a timer and a function that turns the value on and off based on some parameters \(interval, on-time, off-time\). You could create a key repeat event with this to your GPS or autopilot button \(think G1000 “nose up/down” buttons\) or a blinking warning light.

Then you could have **animate** filter that would use a function to make the change happen over given amount of time. If it is a boolean, it just waits before doing a change. If it is a value change, it would gradually change over the given time. A good example of these is documented here for example: [https://developer.roblox.com/en-us/api-reference/enum/EasingStyle](https://developer.roblox.com/en-us/api-reference/enum/EasingStyle). This would be pretty useful for analog gauges 

For booleans, you don’t have any intermediate values, so maybe a **delay** would be the logical equivalent.

And I guess using the animation filter in a repeat would also make sense \(more natural looking flashing lights that simulate an incandescent light bulb instead of a led\).

I found this kind of library, though I am not sure if I can judge its usefulness, but it seems to have a test app on the demo repository: [https://github.com/mobius3/tweeny](https://github.com/mobius3/tweeny) and [https://github.com/mobius3/tweeny-demos](https://github.com/mobius3/tweeny-demos) for an example demo program.

Additionally, the above library appears to support creating loops, so maybe we could create the whole repeat functionality in it?

‌

[Comment created by tigert: 2021-03-09]
I wonder what would be a use case for keeping repeating buttons in ON position apart from making such a case accidentally, and how big of a problem that would be in practice?  
  
I mean one could for example do beacon and strobe lights with actual leds as a crazy example, using the repeat events of the light switch if those existed - or make a landing gear warning light logic based on throttle position, flap switch and landing gear position. But those are kind of a special case and I don’t think you would get very many of them at once?

On the other hand, having a generic repeating programmable on/off timer might also be useful, for things like blinking warning lights. 

‌

[Comment created by DocMoebiuz: 2021-03-02]
Interesting. I first thought of adding the long press / release as a new button event, so that it is actually tracked on the arduino side rather than on the mobiflight connector side \(pure software\).  
I am not really keen on making it configurable on the arduino side because this basically requires a new button type and the upgrade logic for old button types. 

Adding long press and long release is really a simple addition to the current firmware. Repeat however is more complicated because I am not sure how well it works if we have several switches in ON position and they all fire repeatedly.

So yeah, it looks tempting to realize it in Software only.

[Comment created by tigert: 2021-02-14]
… ”for things you don’t want to do accidentally, or do want to adjust the value for a larger amount, and there is no rotary encoder.”

Koseng:
fixes #308 

This PR is all in one.

- Contains Refactor of #1374 
- Configurable Long Button Release
- Configurable Long Button Press with repeat option
- Functionality also in InputShifters and Multiplexers.

**If you want to test, first backup you configuration.**
- Further UI improvements will be done later.
- I'll write documentation when PR is merged.

TODO:
- [x] need to remove one .bak file
- [x] optimize ms label position
- [x] Disable delay input fields if joystick or mobiflight module is not connected

image

image

### Refactoring of ButtonPanel

Refactor `ButtonPanel` and remove all the long copy and paste switch cases. `ButtonPanel` is shrinked from 520 to 200 lines.

- All input panels now share the same interface `IPanelConfigSync`
- Input actions are mapped to input panels in `InputActionMapping.cs`
- In `ButtonInputConfig` two new methods are introduced:
   - `SetInputActionByName`
   - `GetInputActionByName`

If the refactoring is approved, it can also be applied to the other panels like `AnalogPanel` or `EncoderPanel`.

-> Now introducing a new tab or a new input panel is just a couple lines of code.


DocMoebiuz:
image

The error message doesn't show the name of the config item, compare OnPress vs OnRelease

PR #1384, opened 2023-11-09
Update custom json files

elral:
This pull request adds custom board.json files, custom device.json files and updates already existing device.json files to match with the board.json files as with the Mobiflight-CustomDevices reposority.


DocMoebiuz (approved):
Thanks!

PR #1389, opened 2023-11-10
Automatically keep boards alive when Run is active

### LINKED ISSUE #1200: MAX7219 module goes to sleep
**Describe the bug**
One Max7219 8 digit module connected to a mega 2560 pro (mini) consistently goes to sleep after 15 minutes of inactivity with the Msfs2020 Sim and Mobiflight running,  

**Steps To Reproduce**
1. Start MSFS2020, Mobiflight and load a flight on runway.  
2. Mobiflight in run mode.  
3. Leave unattended for 15 minutes
4. Note led display turns off.

**Expected behavior**
LED display should remain on.

**Actual behavior**
LED display turns off after 15 minutes.  Pressing any button in that panel will awaken the board and the display turns back on.  Pressing buttons in other panels does not awake this display.

**Details**
 - MobiFlight Version: 9.7.1
 - FlightSim: MSFS2020
 - AddOn: irrelevant

**Additional context**
If only happens with one particular mega board in my Transponder panel.  Two other 8-digit displays do not exhibit this behavior connected to a different mega board.

neilenns:
Fixes #1200 

Instead of relying on data updates causing boards to get kept alive (which didn't seem to be working correctly anyway), this PR changes things so any time MobiFlight is in Run mode boards will get kept alive, regardless of whether data is getting sent.

* MobiFlightModule.cs: no longer handles any keep alive times and simply exposes a method to set the power save state of the module to either on or off. On devices without updated firmware this will simply update the firmware's lastCommand time. On devices with updated firmware this will update the firmware's lastCommand time if `false` was provided, otherwise it does nothing.
* MobiFlightCache.cs: exposes a new method to keep all connected modules awake, and manages the last time a keep awake message was sent.
* ExecutionManager.cs: Calls the new method on MobiFlightCache every tick when Run is active.

This change stands alone and does not need a matching firmware update even though it's sending a setPowerSaveMode event that the current firmware doesn't support. This is because unsupported commands received by the firmware [still set the `lastCommand` variable to the current time](https://github.com/MobiFlight/MobiFlight-FirmwareSource/blob/d874cf84b4cb1e503e468ea6690247a17f91e54a/src/CommandMessenger.cpp#L99-L103), which is sufficient to prevent power save mode from activating.

Having said that, I'll still open a companion change to the firmware to respect the setPowerSaveMode event and remove all the existing code for setting the last command value based on some (but not all) events.

Here's a sample of the log messages using the updated firmware, with debug enabled and tweaked sleep times so I'm not sitting around for 15 minutes waiting:

```
Debug	11/10/2023 14:39:43	MobiFlightModule.OnDebugPrint(): MobiFlight RaspiPico.debug: Firmware -> Power saving mode on  
Debug	11/10/2023 14:41:15	MobiFlightModule.SetPowerSaveMode(): Setting power save for MobiFlight RaspiPico (COM13) to False
Debug	11/10/2023 14:41:15	MobiFlightModule.OnDebugPrint(): MobiFlight RaspiPico.debug: Firmware -> Disabling power saving mode via request  
Debug	11/10/2023 14:41:15	MobiFlightModule.OnDebugPrint(): MobiFlight RaspiPico.debug: Firmware -> Power saving mode off  
```


neilenns:
This is a draft pull request. DO NOT MERGE. For draft/testing purposes I've set the keep alive interval to 1 minute. The real release should probably maintain the 5 minutes that's been in the code forever.

MobiFlight/MobiFlightModule.cs
```diff
@@ -725,7 +722,7 @@ void OnDebugPrint(ReceivedCommand arguments)
             String value1 = arguments.ReadStringArg();
             String value2 = arguments.ReadStringArg();
             String value3 = arguments.ReadStringArg();
-            Log.Instance.log($"{this.Name}.debug: Firmware -> {value1} {value2} {value3}.", LogSeverity.Debug);
+            Log.Instance.log($"{this.Name}.debug: Firmware -> {value1} {value2} {value3}", LogSeverity.Debug);
```
neilenns, line 725: Removed the period since it looked really weird in logs when only one value was sent from the firmware.

MobiFlight/MobiFlightCache.cs
```diff
@@ -86,6 +90,27 @@ public void Start()
...
+            // time this method is called this test will fail and the modules will be forced
+            // to set their power save mode to off.
+            if (!force && lastKeepAwake.AddMinutes(KeepAwakeIntervalInMinutes) >= DateTime.UtcNow)
+            {
+                return;
+            }
+
+            foreach (var module in Modules)
+            {
+                module.Value.SetPowerSaveMode(false);
```
DocMoebiuz, line 109: See my comment in the firmware too, based on your naming here `KeepAwake` describes better what this is about.

DocMoebiuz:
> Instead of relying on data updates causing boards to get kept alive (which didn't seem to be working correctly anyway), this PR changes things so any time MobiFlight is in Run mode boards will get kept alive, regardless of whether data is getting sent.
> 
This used to work for a long time, but I still haven't figured out why it broke. All the caching logic still makes sense and yet apparently either:
* not every 5 mins there was a forced data sent
* or the arduino code somehow did not behave correctly

I find it a bit unsatisfactory to now know what the real reason for the current bug is.

This approach however still mean an improvement because @neilenns was saying that we are saving quite some memory with changing the logic on the firmware side of things. And the KeepAwake-message only happens anyway when no other messages are sent, so it doesn't mean an overhead for the serial bus.

@neilenns please help to understand why we were seeing the issue, so that we can make sure we are not doing the same mistake again in the future.

@neilenns please rename the method name or convince me that we should only half implement it and call it all the time with the `false`

DocMoebiuz (changes requested):

MobiFlight/MobiFlightCache.cs
neilenns: See my comment on the firmware repo 😂

neilenns:
@DocMoebiuz I couldn't figure out why the existing code didn't work either. I looked at it for a bit and decided instead of trying to debug it deeply it would be better to just go with a simpler solution overall.

elral:
Memory saving for AVR's is upper 2 digit bytes.

neilenns:
> Memory saving for AVR's is upper 2 digit bytes.

If only I got $1 for every byte I save! 😂

DocMoebiuz:
> @DocMoebiuz I couldn't figure out why the existing code didn't work either. I looked at it for a bit and decided instead of trying to debug it deeply it would be better to just go with a simpler solution overall.

What insights did you have so far? Maybe if we talk about it together then we can pin down the error.

neilenns:
Ok I tweaked this a bit, now with maybe 20% more awesome. I made a firmware change to actually support requesting the modules go to sleep, and the connector now sends a "go to sleep" message to all connected modules when Stop is activated.

The firmware still does keep its own internal 15 minute timeout so even if the connector doesn't send the "go to sleep" message the boards will eventually sleep on their own.

Advantage to this approach: all the method names make sense 🤣

neilenns:
> What insights did you have so far? Maybe if we talk about it together then we can pin down the error.

None, I didn't dig too deeply since I'd already decided it was easier to just go with a simpler keep awake model.

MobiFlight/ExecutionManager.cs
```diff
@@ -992,6 +992,9 @@ void FsuipcCache_ConnectionLost(object sender, EventArgs e)
         void timer_Started(object sender, EventArgs e)
         {
             OnStarted?.Invoke(this, new EventArgs());
+
+            // Force all the modules awake whenver run is activated
+            mobiFlightCache.KeepConnectedModulesAwake(true);
```
DocMoebiuz, line 997: Was there a specific reason that you had decided to take the timer event?

Better options are `MobiFlightModule.Start()` method, `MobiFlightCache.Start()` method for it, or `ExecutionManager.Start()` method.

I am not even sure why are really doing this. What's the sense in it? On Start, we have an empty cache and we will always send messages down to the device if requried, e.g. the current 7-segment values.

MobiFlight/ExecutionManager.cs
```diff
@@ -1005,6 +1008,7 @@ void timer_Stopped(object sender, EventArgs e)
 #endif
             inputCache.Clear();
             inputActionExecutionCache.Clear();
+            mobiFlightCache.ActivateConnectedModulePowerSave();
```
DocMoebiuz, line 1011: Was there a specific reason that you had decided to take the timer event?

Better options are `MobiFlightModule.Stop()` method, `MobiFlightCache.Stop()` method for it, or `ExecutionManager.Stop()` method.

I am not even sure why are really doing this. What's the sense in it? On Stop, we don't care, and the module should take care which it would even without this message.

DocMoebiuz (changes requested):

MobiFlight/ExecutionManager.cs
neilenns: > On Start, we have an empty cache and we will always send messages down to the device if requried, e.g. the current 7-segment values.

That won't be sufficient to wake the devices up with the new firmware. The new firmware only resets its sleep timer when getting an unknown message or a wakeup message. That change saves us some space in the firmware.

> What's the sense in it? 

To be fair, it's not strictly necessary to have this call here. The keep awake message code in each timer tick will wake the devices up the first time the tick is called. I wasn't wild about the subtleness of that though for future people reading this code and thought it would be better to have something explicit on start that forces the wakeup so the flow is more understandable and doesn't reply on "magic" in the tick method.

> Was there a specific reason that you had decided to take the timer event?

Because the keep awake messages happen in the timer tick, so having wakeup and sleep in timer started and timer stopped puts the code together, making it easier to understand for future developers. If I move the code to module or cache it's in an entirely different place.

MobiFlight/ExecutionManager.cs
neilenns: > What's the sense in it? On Stop, we don't care, and the module should take care which it would even without this message.

It's not strictly necessary, but it does make the end user experience pretty nice. In the normal operating case, all your connected modules will wake up as soon as run happens and shut down as soon as MobiFlight stops.

neilenns:
@DocMoebiuz I spent some time in the debugger and I don't think there's anything in the desktop app that would directly cause #1200.

While the old code was a bit more complicated (it depends on each different device type doing a check for keep alive and then resending its data if keep alive is needed), it seems in my debugging to be fine. I couldn't reproduce or find, via code inspection, any way that the keep alive check would return false when it should really return true, and the if statements for each of the device types are straightforward.

Only things I can think of:

1. Mobiflight didn't think the sim was connected. In that situation the keep alive code will never get called because no data is getting sent regularly to the connected modules.
2. There were no outputs in the loaded config. I hit this in my initial debugging, where I had an old test file loaded that only had inputs. With the previous keep alive code you *must* have some sort of output configured otherwise your board will go to sleep.
3. The data did get re-sent to the firmware but for some reason the firmware didn't reset the timer. I can't think of any reason why that would happen though.

MobiFlight/ExecutionManager.cs
neilenns: Funny but true story: I just finished doing more testing with the old code and closed Mobiflight. My radio stack is still has all eight frequencies lit up and I went "wtf, why didn't that turn off?". Oh, right, because with the old code I have to wait 15 minutes for the board to time itself out. LAME!!! 😀

neilenns: I literally just checked out this branch so I could hit the stop button and turn the boards off 🤣

MobiFlight/ExecutionManager.cs
DocMoebiuz: Please move it at least to ExecutionManager.Start() that's the equivalent. It doesn't make sense in this callback function. The only place where the timer is enabled is in ExecutionManager.Start() - that were a future developer should find it.

MobiFlight/ExecutionManager.cs
DocMoebiuz: Same here. The other statements are simply wrong: 

image

They should be moved out to the `ExceutionManager.Stop()`

`timer.enabled=false` is only called in two places:
In `ExecutionManager.Stop()`  or when an exception is raised in the timer_tick - this should be changed to `ExecutionManager.Stop()` to make sure everything is stopped correctly.

DocMoebiuz (changes requested):

DocMoebiuz:
> @DocMoebiuz I spent some time in the debugger and I don't think there's anything in the desktop app that would directly cause #1200.
> 
> While the old code was a bit more complicated (it depends on each different device type doing a check for keep alive and then resending its data if keep alive is needed), it seems in my debugging to be fine. I couldn't reproduce or find, via code inspection, any way that the keep alive check would return false when it should really return true, and the if statements for each of the device types are straightforward.
> 
> Only things I can think of:
> 
> 1. Mobiflight didn't think the sim was connected. In that situation the keep alive code will never get called because no data is getting sent regularly to the connected modules.
> 2. There were no outputs in the loaded config. I hit this in my initial debugging, where I had an old test file loaded that only had inputs. With the previous keep alive code you _must_ have some sort of output configured otherwise your board will go to sleep.
> 3. The data did get re-sent to the firmware but for some reason the firmware didn't reset the timer. I can't think of any reason why that would happen though.

Thanks for taking the time and digging into it. Still weird that we can't solve the root cause. Never satisfying.

neilenns:
Start/Stop moved to ExecutionManager.Start/Stop. Ready for rereview!

DocMoebiuz:
It'd be great to do this https://github.com/MobiFlight/MobiFlight-Connector/issues/1406 right away

neilenns:
> It'd be great to do this #1406 right away

I can look at it later this morning and open a PR for it.

DocMoebiuz (approved):

PR #1394, opened 2023-11-11
Drag and Drop for configuration entries

### LINKED ISSUE #251: Allow user to rearrange order of configs in list
Confirmations:
* User can select an item and define the new position 
* Saving the config, also saves the new position

Koseng:
fixes #251
Drag and Drop for configuration entries.

- [x] Autoscroll implemented
- [x] Changed cursor more prominently on drag
- [x] Highlight current drop target 
- [ ] Decide drop behavior. 
A) Drop the config between the two highlighted rows -  TEST IN OUTPUT CONFIG
B) Drop the config always above the single highlighted target row. TEST IN INPUT CONFIG 

### Behavior A
AnimationA2

### Behavior B
AnimationB

### DIFF
--- a/UI/Panels/InputConfigPanel.Designer.cs
+++ b/UI/Panels/InputConfigPanel.Designer.cs
@@ -30,10 +30,10 @@
         {
             this.components = new System.ComponentModel.Container();
             System.ComponentModel.ComponentResourceManager resources = new System.ComponentModel.ComponentResourceManager(typeof(InputConfigPanel));
-            System.Windows.Forms.DataGridViewCellStyle dataGridViewCellStyle17 = new System.Windows.Forms.DataGridViewCellStyle();
-            System.Windows.Forms.DataGridViewCellStyle dataGridViewCellStyle18 = new System.Windows.Forms.DataGridViewCellStyle();
-            System.Windows.Forms.DataGridViewCellStyle dataGridViewCellStyle19 = new System.Windows.Forms.DataGridViewCellStyle();
-            System.Windows.Forms.DataGridViewCellStyle dataGridViewCellStyle20 = new System.Windows.Forms.DataGridViewCellStyle();
+            System.Windows.Forms.DataGridViewCellStyle dataGridViewCellStyle9 = new System.Windows.Forms.DataGridViewCellStyle();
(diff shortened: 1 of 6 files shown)


JaimeLeon2:
Hello, very nice improvement.  This is a very much anticipated and long awaited feature.  
Just a few comments, if I may:
1.  I find that simply click and drag is way too easy to do accidentally.  One can really subtly and without really noticing, wreak havoc in the order of config lines, just clicking and dragging without knowing.  I think it would be much better to do shift-click and drag.
2. Can you change the cursor to another icon, so that we can visually see the mode has changed?
3. Can you move the selected line (highlighted) as you drag it?  It is not very clear where it will land.  It takes some practice.
4. This is for extra credit: Can you select and move with multiple lines being selected (block move)?
Thank you very much.

Koseng:
> Hello, very nice improvement. This is a very much anticipated and long awaited feature. Just a few comments, if I may:
> 
> 1. I find that simply click and drag is way too easy to do accidentally.  One can really subtly and without really noticing, wreak havoc in the order of config lines, just clicking and dragging without knowing.  I think it would be much better to do shift-click and drag.
> 2. Can you change the cursor to another icon, so that we can visually see the mode has changed?
> 3. Can you move the selected line (highlighted) as you drag it?  It is not very clear where it will land.  It takes some practice.
> 4. This is for extra credit: Can you select and move with multiple lines being selected (block move)?
>    Thank you very much.

Hi,

1. I increased the distance to start the dragging when holding the mouse down a bit. If it still leads to too many accidental drags, we need to think about some under stuff. For example only start drag on certain column.
2. Cursor already was changed but pretty subtle. In default behavior a small rectangle is added to the cursor. Now cursor is changed to "Hand".
3. The current drop target is now highlighted.
4. Maybe at some time in the future.

DocMoebiuz:
Super cool that you took this on. Are you sure the underlying order in the datatable is really changed? I haven't checked yet, but when you save after modifying, then will it be loaded in the same re-ordered sequence?

Koseng:
> Super cool that you took this on. Are you sure the underlying order in the datatable is really changed? I haven't checked yet, but when you save after modifying, then will it be loaded in the same re-ordered sequence?

Yes and yes. 

And please try the latest build in the PR and play around a bit. We need to decide which UX concept to use. Behavior A (currently implemented in Output config) or Behavior B (currently implemented in Input config). We think both ways are fine now and also need your vote.

PR #1402, opened 2023-11-14
Add try/catch around save

### LINKED ISSUE #1401: Exception thrown when saving config file
Reported by user P. Santos in Discord. When attmepting to save the config file an exception is thrown:

```
System.UnauthorizedAccessException: Access to the path is denied.
   at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
   at System.IO.FileStream.WriteCore(Byte[] buffer, Int32 offset, Int32 count)
   at System.IO.FileStream.FlushWrite(Boolean calledFromFinalizer)
   at System.IO.FileStream.Dispose(Boolean disposing)
   at System.IO.Stream.Close()
   at System.IO.StreamWriter.Dispose(Boolean disposing)
   at System.IO.StreamWriter.Close()
   at System.Xml.XmlTextWriter.Close()
   at System.Xml.XmlDocument.Save(String filename)
   at MobiFlight.ConfigFile.SaveFile(DataSet outputConfig, DataSet inputConfig)
   at MobiFlight.UI.MainForm._saveConfig(String fileName)
   at MobiFlight.UI.MainForm.saveToolStripButton_Click(Object sender, EventArgs e)
```

This exception should get caught and an error message shown to the user instead of crashing.

neilenns:
Fixes #1401 

Add a try/catch around saving the config so we don't crash if there's an access exception.

I couldn't find any way to make the exception happen. I tried saving to c:\windows and c:\program files and in both cases nothing happened. No exceptions were thrown.

### DIFF
--- a/UI/MainForm.cs
+++ b/UI/MainForm.cs
@@ -1701,12 +1701,24 @@ namespace MobiFlight.UI
         /// <summary>
         /// saves the current config to filename
         /// </summary>        
-        private void _saveConfig(string fileName)
+        private void SaveConfig(string fileName)
         {
             outputConfigPanel.ApplyBackwardCompatibilitySaving();
 
             ConfigFile configFile = new ConfigFile(fileName);
-            configFile.SaveFile(outputConfigPanel.DataSetConfig, inputConfigPanel.InputDataSetConfig);
+
+            // Issue 1401: Saving the file can result in an UnauthorizedAccessException, so catch any
+            // errors during save and show it in a dialog instead of crashing.
+            try
+            {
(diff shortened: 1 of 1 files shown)


UI/MainForm.cs
```diff
@@ -1701,12 +1701,24 @@ private void _applyBackwardCompatibilityLoading()
         /// <summary>
         /// saves the current config to filename
         /// </summary>        
-        private void _saveConfig(string fileName)
+        private void SaveConfig(string fileName)
```
neilenns, line 1704: Fixes a style message from Visual Studio

DocMoebiuz (approved):
Thanks!

PR #1404, opened 2023-11-15
Treat pull request firmware builds as supporting all features

### LINKED ISSUE #1403: Pull request firmware builds should get recognized and treated as supporting all features
Right now pull request firmware build version numbers (`0.0.prnumber`) are treated like any other firmware version and result in features being blocked (e.g. changing the board name, regenerating the serial number, custom devices, etc.). This makes testing custom firmware a pain.

Since there's no such thing as production firmware builds with versions below `1.0.0` we should update the connector to treat < 1.0.0 as "supports everything".

neilenns:
Fixes #1403


DocMoebiuz (approved):
Nice

PR #1407, opened 2023-11-16
Code cleanup

### LINKED ISSUE #1406: [REFACTOR] Remove methods from timer stop event in Execution Manager
related to #1389 

as part of the pull request we noticed that there are a few methods inside the timer_stopped event handler. This is not so great, because there are more things to do when the timer stops, and those are actually present  in the ExecutionManager.Stop() method already.

Things to do:
1. move all method calls to stop

2. change the exception handling
There are currently only two places where the `timer.Enabled = false` is done, which stops the timer and triggers the callback.
  - in the ExecutionManager `Stop()`
  - in ExecutionManager config execution inside one of the Exception `catch` blocks.

  For the second, we simply should call Stop, instead of stopping the timer.

neilenns:
Fixes #1406 

* Move method calls out of the timer Stopped() event handler
* Call Stop() instead of setting the timer's state via a property


neilenns:
Is this what you were thinking @DocMoebiuz ?

MobiFlight/ExecutionManager.cs
```diff
@@ -1023,7 +1022,7 @@ void timer_Tick(object sender, EventArgs e)
             {
                 Log.Instance.log($"Error on config execution: {ex.Message}", LogSeverity.Error);
                 isExecuting = false;
```
DocMoebiuz, line 1024: can be removed. is done in `Stop()`

DocMoebiuz (changes requested):
Exactly. The one line can also be removed.

MobiFlight/ExecutionManager.cs
neilenns: Fixed

DocMoebiuz (approved):
Cool!

DocMoebiuz (approved):

PR #1411, opened 2023-11-17
Add check for bad CH340 chips

### LINKED ISSUE #1410: Add a warning for counterfeit CH340 chips

neilenns:
Fixes #1410 

Add a test against the garbled device name that is returned by bad CH340 chips and use that to log an error if a bad chip is detected.

### DIFF
--- a/MobiFlight/Monitors/SerialPortMonitor.cs
+++ b/MobiFlight/Monitors/SerialPortMonitor.cs
@@ -8,6 +8,70 @@ namespace MobiFlight.Monitors
 {
     public class SerialPortMonitor : DeviceMonitor
     {
+        /// <summary>
+        /// Uses WMI to retrieve property data from a ManagementObject.
+        /// </summary>
+        /// <param name="managementObject">The object to retrieve the property from</param>
+        /// <param name="propertyName">The name of the property to retrieve</param>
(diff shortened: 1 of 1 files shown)


MobiFlight/Monitors/SerialPortMonitor.cs
```diff
@@ -16,7 +80,7 @@ override protected void Scan()
 
             // Code from https://stackoverflow.com/questions/45165299/wmi-get-list-of-all-serial-com-ports-including-virtual-ports
             ManagementObjectSearcher searcher = new ManagementObjectSearcher("root\\CIMV2", "SELECT * FROM Win32_PnPEntity WHERE ClassGuid=\"{4d36e978-e325-11ce-bfc1-08002be10318}\"");
-            foreach (ManagementObject queryObj in searcher.Get())
+            foreach (ManagementObject queryObj in searcher.Get().Cast<ManagementObject>())
```
neilenns, line 83: Clean up code correctness message from VS.

DocMoebiuz (approved):

PR #1418, opened 2023-11-18
Drag'nDrop improvements including reset sorting

Koseng:
- [x] Short row highlight on drop 
- [x] Bitmap fix when leaving window and dragged row is not visible anymore
- [x] Reset Sorting on third header column click.
- [x] Drag and Drop shall not be possible on sorted list 
- [x] MessageBox when drag and drop is not allowed
- [x] Reset sorting on file open and file new 
- [x] i18n
- [x] ~~unit tests~~ - does not apply here
- [ ] documentation

related to #251 
improves #1405

### DIFF
--- a/UI/Panels/InputConfigPanel.cs
+++ b/UI/Panels/InputConfigPanel.cs
@@ -253,10 +253,23 @@ namespace MobiFlight.UI.Panels
                 row.Selected = false;
             }
         }
-
-        private bool IsSortingActive()
+        
+        /// <summary>
+        /// Is sorting in DataGridView active.
+        /// </summary>        
+        public bool IsSortingActive()
         {
             return inputsDataGridView.SortOrder != SortOrder.None;
+        }
+
+        /// <summary>
+        /// Reset sorting in DataGridView
+        /// </summary>
+        public void ResetSorting()
+        {
+            LastSortingColumnName = string.Empty;
+            LastSortingOrder = SortOrder.None;
+            inputsDataTable.DefaultView.Sort = string.Empty;
         }
 
         private void inputsDataGridView_Sorted(object sender, EventArgs e)
@@ -268,9 +281,7 @@ namespace MobiFlight.UI.Panels
                 // Reset sorting after previous Descending
                 if (LastSortingColumnName == name && LastSortingOrder == SortOrder.Descending)
                 {
-                    LastSortingColumnName = string.Empty;
-                    LastSortingOrder = SortOrder.None;
-                    inputsDataTable.DefaultView.Sort = string.Empty;
+                    ResetSorting();
                 }
                 else
                 {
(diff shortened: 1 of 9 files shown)


UI/Panels/InputConfigPanel.cs
```diff
@@ -698,9 +733,11 @@ private void inputsDataGridView_MouseDown(object sender, MouseEventArgs e)
         private void inputsDataGridView_MouseMove(object sender, MouseEventArgs e)
         {
             if (MouseButtons.Left == (e.Button & MouseButtons.Left))
-            {                
+            {
                 // When mouse leaves the rectangle, start drag and drop
-                if (RectangleMouseDown != Rectangle.Empty && !RectangleMouseDown.Contains(e.X, e.Y))
+                if (RectangleMouseDown != Rectangle.Empty &&
```
DocMoebiuz, line 738: We could change this expression so to do an early return

Additionally, we should provide a message to the user that dragging is not possible because sorting is active... users likely won't understand why suddenly the UI behaves differently.

UI/Panels/InputConfigPanel.cs
```diff
@@ -245,14 +254,40 @@ private void duplicateInputsRowToolStripMenuItem_Click(object sender, EventArgs
...
+        }
+
+        private void inputsDataGridView_Sorted(object sender, EventArgs e)
+        {
+            if (IsSortingActive())
+            {
+                string name = inputsDataGridView.SortedColumn.Name;
+                // It is always (Ascending) -> (Descending) -> (Ascending)
+                // Reset sorting after previous Descending
+                if (LastSortingColumnName == name && LastSortingOrder == SortOrder.Descending)
```
DocMoebiuz, line 269: If we move this into its own method, then we can also call it when opening / creating a different file.

I really think it should be reset when dealing with a new file.

DocMoebiuz (changes requested):
See my comments.

Additionally, when selecting a row, and sorting, I ended up with multiple selected rows.

UI/Panels/InputConfigPanel.cs
```diff
@@ -42,6 +46,8 @@ public InputConfigPanel()
 
         void Init()
         {
+            inputsDataGridView.DataMember = null;
```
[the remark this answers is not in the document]
Koseng: If you data bind to a `DataSet` you also need to tell the control which table to use. That is the property `DataMember`.
If you data bind to a `DataTable` instead, you do not use the `DataMember` Property.

The trick to get the reset of the sorting working, is to directly data bind to the `DataTable`. Then for sorting the `DataTable.DefaultView` is used by the control, which can be accessed. Otherwise some internal Views are used which cannot be accessed.

The code switches from DataSet binding to direct DataTable binding.

UI/Panels/InputConfigPanel.cs
Koseng: Done.

UI/Panels/InputConfigPanel.cs
Koseng: "We could change this expression so to do an early return."
What do you mean? After the if there is no more code. And I think it is evaluated step by step and if the first condition fails the others are not executed any more by .net.

Koseng: Ok in context with the change I make, I understand.

Koseng: Done.

DocMoebiuz (approved):
Mega cool!

PR #1434, opened 2023-11-20
Auto update hubhop presets and display timestamp

### LINKED ISSUE #1152: Feature request: upgrade hubhop events after installation
Since HubHop is now an integral part of how MobiFlight works for both MSFS2020 and X-Plane, I wonder if we should just 
 automatically update the latest events from hubhop when you run MobiFlight for the first time after installation / upgrade?

Because there is currently an issue where your events will get replaced with an older copy depending on the age of the snapshot that was included in a new release. And for a new installation of stable MobiFlight this means the data might be weeks or months old, depending on when the latest stable version of MF was released.

What if we just didnt ship the snapshot data at all, but rather just downloaded it on the first run?

### LINKED ISSUE #1363: Display time stamp for events database
When Mobiflight updates events, there is no way of telling the events database currently available or how old it is.

Display a time stamp representing the snapshot time of the database currently installed.  It could be displayed in Extras - Hubhop update option.  I think this will probably require the time stamp to be stored in Hubhop and updated every time a change in the database is accepted.

No alternative currently exists, and there is no way of telling how old the current database installed is.

**Additional context**
Not knowing the database age is an issue compounded by the fact that Mobiflight installation installs a very old database file.  I think most users are unaware of this and assume that installing Mobiflight also installs an updated set of events.  The reality is the contrary of that.

DocMoebiuz:
- [x] Add a notification after startup that your database might be out-dated
- [x] User can trigger update (re-using the same function as trigger via menu)
- [x] User can cancel the update
- [x] User can disable feature through settings
- [x] Timestamp of last update displayed 
- [x] Community Feedback program integration
- [x] i18n
- [x] ~~unit tests~~ doesn't apply
- [ ] documentation

fixes #1152 
fixes #1363


MobiFlightConnector.csproj
```diff
@@ -1139,7 +1139,7 @@
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </None>
     <None Include="Presets\xplane_hubhop_presets.json">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
DocMoebiuz, line 1142: I changed this for testing, i made the presets files older and I wanted to make sure they will get copied over always.

SimConnectMSFS/WasmModuleUpdater.cs
```diff
@@ -282,6 +282,20 @@ public bool DownloadHubHopPresets()
             return true;
         }
 
+        static public DateTime HubHopPresetTimestamp()
+        {
+            var lastModified = DateTime.MinValue;
+            try
+            {
+                lastModified = System.IO.File.GetLastWriteTimeUtc($@"{WasmEventsHubHopFolder}\{WasmEventsHubHopFileName}");
```
DocMoebiuz, line 290: I took the lastWriteTimeUtc because it is actually safer to go by the modified date than the create date.
Create date is set when copied. If a user copies an old preset file into the directory, it would have the current timestamp.

Not super relevant in normal circumstances.

Presets/msfs2020_hubhop_presets.json
Koseng, line 1: Seems to be an older file version?

Presets/xplane_hubhop_presets.json
Koseng, line 1: File in PR seems to be older.

MobiFlightConnector.csproj
DocMoebiuz: > I changed this for testing, i made the presets files older and I wanted to make sure they will get copied over always.

see my first comment

Presets/msfs2020_hubhop_presets.json
DocMoebiuz: > I changed this for testing, i made the presets files older and I wanted to make sure they will get copied over always.

see my first comment

Presets/xplane_hubhop_presets.json
DocMoebiuz: > I changed this for testing, i made the presets files older and I wanted to make sure they will get copied over always.

see my first comment

SimConnectMSFS/WasmModuleUpdater.cs
Koseng: If it urges you to explain, do a code comment for future generations?

Presets/msfs2020_hubhop_presets.json
Koseng: Just wanted to make sure its intentional...

SimConnectMSFS/WasmModuleUpdater.cs
DocMoebiuz: it is in here.

DocMoebiuz (commented):
Thanks for the feedback.

Koseng (approved):

PR #1435, opened 2023-11-20
Change MessageType to int

### LINKED ISSUE #1433: MessageType (a.k.a. Id) for custom devices should be an int, not a string
The code uses strings, but the JSON file is an int. The code should update to use int.

neilenns:
Fixes #1433

Change the type of the message id in the custom device definition object to `int`, then ripple that change through all the rest of the code.

This does NOT need to ship with 10.0.

This is not a breaking change. The schema already defined this property as an int, it's just that the code internally was calling it a string and that happened to magically work.


elral:
`MobiFlightUnitTests/MobiFlight/MobiFlightCustomDeviceTests.cs`does contain custom device TM1637.

This device was at the beginning intended to add but afterwards deleted at it is now a base device.

DocMoebiuz (approved):

PR #1438, opened 2023-11-21
Startup will finish with only one pico

DocMoebiuz:
- Added CheckIfLookupFinished to UsbDeviceMonitor_PortAvailable

### DIFF
--- a/MobiFlight/MobiFlightCache.cs
+++ b/MobiFlight/MobiFlightCache.cs
@@ -87,7 +87,7 @@ namespace MobiFlight
                     if (0 == AvailableComModules.Count) {
                         isFirstTimeLookup = false;
                         LookupFinished?.Invoke(this, new EventArgs());
-                        Log.Instance.log($"End looking up connected modules. No modules found.", LogSeverity.Debug);
+                        Log.Instance.log($"Finish looking up connected modules. No modules found.", LogSeverity.Info);
                     } 
                 });
         }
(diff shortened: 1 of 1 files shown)


MobiFlight/MobiFlightCache.cs
```diff
@@ -87,7 +87,7 @@ public void Start()
                     if (0 == AvailableComModules.Count) {
                         isFirstTimeLookup = false;
                         LookupFinished?.Invoke(this, new EventArgs());
-                        Log.Instance.log($"End looking up connected modules. No modules found.", LogSeverity.Debug);
+                        Log.Instance.log($"Finish looking up connected modules. No modules found.", LogSeverity.Info);
```
DocMoebiuz, line 90: I changed these events to info because I think they might be valuable enough to be displayed together with the other board detection events.

MobiFlight/MobiFlightCache.cs
```diff
@@ -256,6 +256,8 @@ private void UsbDeviceMonitor_PortAvailable(object sender, PortDetails e)
             OnCompatibleBoardDetected(info);
             var result = new MobiFlightModule(info);
             ModuleConnected?.Invoke(result, new EventArgs());
+
+            await CheckIfLookUpFinished();
```
DocMoebiuz, line 260: This was missing here. See SerialPortMonitor_PortAvailable

neilenns (approved):

PR #1439, opened 2023-11-21
Prevent `IsReady` from blocking too long

### LINKED ISSUE #1437: USB SDCard Reader with no card blocks raspi detection on startup
**Describe the bug**
I have a USB thumb drive SD card reader. When I don't put any SD cards in there, then the start up of MobiFlight takes 20 seconds.
It also doesn't recognize my Pico directly during startup, but only some 20 secs later.

Without the USB SD Card Reader thing, the startup is fast.

**Steps To Reproduce**
1. Plug in your USB Card Reader Thumb Stick
2. Plug in your Pico with no firmware installed
3. Start MobiFlight
5. See error

**Expected behavior**
It should start fast, as always.

**Actual behavior**
It starts very slowly, the progress bar seems to be stuck, and the Pico is not detected after startup.

**Details**
 - MobiFlight Version: 9.7.1

DocMoebiuz:
fixes #1437 

This change makes sure that

- There is a timeout of 100ms for the `DriveInfo.isReady` check
- There can be no parallel execution of a `Scan` for the UsbDeviceMonitor
- A lock was added to prevent concurrent access to `DetectedPorts`
- `FindConnectedUsbDevices` timeout is reduced significantly which will make the firmware reset faster.

### DIFF
--- a/MobiFlight/MobiFlightCache.cs
+++ b/MobiFlight/MobiFlightCache.cs
@@ -276,6 +276,7 @@ namespace MobiFlight
         /// Returns a list of connected USB drives that are supported with MobiFlight and are in flash mode already,
         /// as opposed to being connected as COM port.
         /// </summary>
+        /// <param name="WaitInMilliseconds">Time for the UsbDeviceMonitor to perform a port scan</param>
         /// <returns>The list of connected USB drives supported by MobiFlight.</returns>
         public static List<MobiFlightModuleInfo> FindConnectedUsbDevices(double WaitInMilliseconds = 500)
         {
@@ -287,7 +288,10 @@ namespace MobiFlight
             {
                 usbDeviceMonitor.Start();
                 var task = Task
-                    // we need to wait for the timer to trigger
+                    // we need to give the usbDeviceMonitor the chance
+                    // to do the USB drive scan and return a valid list
+                    // of Detected Ports.
+                    // This might take a few moments to complete.
                     .Delay(TimeSpan.FromMilliseconds(WaitInMilliseconds))
                     .ContinueWith(_ => usbDeviceMonitor.DetectedPorts
                         .ForEach(p =>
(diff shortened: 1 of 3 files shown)


MobiFlight/MobiFlightCache.cs
```diff
@@ -277,33 +277,38 @@ public bool Available()
         /// as opposed to being connected as COM port.
         /// </summary>
         /// <returns>The list of connected USB drives supported by MobiFlight.</returns>
-        public static List<MobiFlightModuleInfo> FindConnectedUsbDevices()
+        public static List<MobiFlightModuleInfo> FindConnectedUsbDevices(double WaitInMilliseconds = 500)
```
neilenns, line 280: There should be some comments in the code explaining why all this was necessary, referencing back to this bug. Future us, looking at this code, will have no idea why there's a task involved at all.

neilenns (changes requested):

MobiFlight/Monitors/UsbDeviceMonitor.cs
```diff
@@ -1,20 +1,50 @@
...
+        /// <summary>
+        /// Returns the IsReady value of the DriveInfo but also
+        /// applies a timeout to not wait for it too long.
+        /// </summary>
+        /// <param name="drive"></param>
+        /// <returns>Actual `IsReady` value, or false if timed out</returns>
+        async Task<bool> DriveIsReady(DriveInfo drive)
+        {
+            // Issue 1074: Failing to check for IsReady caused an IOException on certain machines
+            // when trying to read the volume label when the drive wasn't actually ready.
```
neilenns, line 21:
```suggestion
            // Issue 1074: Failing to check for IsReady caused an IOException on certain machines
            // when trying to read the volume label when the drive wasn't actually ready.
            // Issue 1437: IsReady takes forever to timeout when called on USB card readers that have no
            // cards in them. Wrap the test in a task so we limit the damage to only 100ms.
```

neilenns (changes requested):

neilenns (approved):

PR #1454, opened 2023-11-27
Pico - Limit the number of PWM pins to 16

### LINKED ISSUE #1453: Two Pins get set at the same time on Pico
If two pins with a differnece of 16 of the pin numbers are set to output, both pins are getting changed if one of both pins changes it status.

More explantions can be found in the [issue  283](https://github.com/MobiFlight/MobiFlight-FirmwareSource/issues/283) from the FW repo.

elral:
This PR limits the number of PWM pins to 16.
Only pin 0 to pin 15 can be used as PWM output due to the HW limitation of the RP2040 processor.

A detailed explanation can be found in the [PR 283](https://github.com/MobiFlight/MobiFlight-FirmwareSource/pull/287) within the FW repo.

Fixes #1453


neilenns (approved):

DocMoebiuz (approved):
Thanks!

PR #1457, opened 2023-11-28
Improved encoder reading for Octavi

DocMoebiuz:
we don't poll but use the native callback when reports are received.

### DIFF
--- a/MobiFlight/Joysticks/JoystickManager.cs
+++ b/MobiFlight/Joysticks/JoystickManager.cs
@@ -5,8 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Text;
-using System.Windows.Forms;
+using System.Timers;
 
 namespace MobiFlight
 {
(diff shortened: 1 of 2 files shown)


Koseng:
Before changing the timer in JoystickManager, everything happened in Gui-Thread. So no locking was necessary. Now after the change, I am not sure wether current locking around accessing "joysticks" is enough.

- JoysticksConnected() is called regularly
- PollTimer_Tick(...) is called regularly
- GetJoystickBySerial() might be called regularly.
- Stop() might be called suddenly.

Koseng:
Before that PR JoystickManager and Joystick were single threaded. Everything was called from Gui Thread. Also the Timer was a Winforms Timer in Gui Thread.

Changing the Timer to Systems.Timer changes the whole situation. Now JoystickManager and Joystick gets called from several threads and must be thread safe. Probably thread safety is not guaranteed everywhere.

Methods in JoystickManager and Joystick are called all the time from the Timer (THREAD 1) and from the ExecutionEngine (THREAD 2).

DocMoebiuz:
i actually don't think that these situations are problematic. Problematic is when the collections are altered, like items are added or removed while a different thread is accessing it.

I made sure to introduce a lock around the actions that modify the collection.

PR #1465, opened 2023-12-01
Pico firmware upload works correctly on start up

### LINKED ISSUE #1464: Pico firmware upload doesn't work correctly on start up
**Describe the bug**
Pico firmware upload doesn't work correctly on start up. When it is uploaded in the settings dialog it works ok.

**Steps To Reproduce**
1. Connect a Pico without MobiFlight firmware installed
2. Start MobiFlight
3. Initiate the Auto-firmware upload

**Expected behavior**
It should finish with success message

**Actual behavior**
It finishes with an error message, but the pico is flashed correctly.

**Details**
 - MobiFlight Version: 9.7.1.x

**Additional context**
You can see that there are two modules detected.

Startup

DocMoebiuz:
fixes #1464

- [x] removed `modulesForFlashing.AddRange` which is redundant and just from the old code, but not needed anymore

### DIFF
--- a/UI/MainForm.cs
+++ b/UI/MainForm.cs
@@ -493,8 +493,6 @@ namespace MobiFlight.UI
                 }
             }
 
-            // Connected USB devices that are in bootsel mode get added to the firmware flashing list
-            modulesForFlashing.AddRange(MobiFlightCache.FindConnectedUsbDevices());
 
             if (Properties.Settings.Default.FwAutoUpdateCheck && (modulesForFlashing.Count > 0 || modulesForUpdate.Count > 0))
             {


UI/MainForm.cs
```diff
@@ -493,8 +493,6 @@ void CheckForFirmwareUpdates ()
                 }
             }
 
-            // Connected USB devices that are in bootsel mode get added to the firmware flashing list
```
DocMoebiuz, line 496: not needed anymore because we are scanning continuosly for new USB devices anyways and those that are currently detected are already discovered.

This is just old code which i did not remove by accident.

neilenns (approved):

PR #1468, opened 2023-12-02
Added string SimVars

MiMaKes:
* Added client data area for string SimVars
* Added separate StringSimVars list
* Register on string SimVars
* Assume that a string shall be used when RPN-Code contains "String"

Handling as discussed with @DocMoebiuz 

In conjunction with https://github.com/MobiFlight/MobiFlight-WASM-Module/pull/19

### DIFF
--- a/SimConnectMSFS/SimConnectCache.cs
+++ b/SimConnectMSFS/SimConnectCache.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using HidSharp.Utility;
 using Microsoft.FlightSimulator.SimConnect;
+using System.Text.RegularExpressions;
 
 namespace MobiFlight.SimConnectMSFS
 {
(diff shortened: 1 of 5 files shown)


tigert:
First test: works. This is super cool.

image

SimConnectMSFS/SimConnectCache.cs
```diff
@@ -426,66 +449,135 @@ public void setEventID(string eventID)
...
 
-        private void RegisterSimVar(string SimVarName)
+        private FSUIPCOffsetType RegisterSimVar(string SimVarName)
         {
-            SimVar NewSimVar = new SimVar() { Name = SimVarName, ID = (uint) (SimVars.Count + SIMVAR_DATA_DEFINITION_OFFSET) };
-            SimVars.Add(NewSimVar);
-
-            if (MaxClientDataDefinition >= NewSimVar.ID)
+            // Matches presets like "(A:TITLE,String)" in different variations. These will most likely be of type String an will therefore be treated as String.
+            Match stringPreset = Regex.Match(SimVarName, "\\(.*A.*:.*,.*String.*\\)");
```
Koseng, line 508: Valid presets are also (A:TITLE,string) and (A:TITLE,STRING), which also should be recognized.
So first convert the string .ToUpper() and then check for STRING in the regex or use a regex where case sensitivity does not matter.

tigert:
I gave this a few tests and I noticed a lot of the modifiers do not really handle strings in a consistent manner, should I file bugs about each issue I found or what would be the best way to coordinate the changes needed?

MiMaKes:
> Thank you very much.
> 
> I understand why you did the changes to `GetSimVar` and `RegisterSimVar` and how they return a type now and not the value anymore, and you pass in the out-parameters instead.
> 
> Tbh, I don't like it so much but it works for now. I will merge the PR and we can test it with people and if you like we can make some code improvements later.

With this approach I tried to keep changes in `ExecutionManager.cs` minimal, but yes it probably isn't the most beautiful solution. I would be happy about a discussion about possible improvements :)

PR #1494, opened 2023-12-13
Include installer in release

### LINKED ISSUE #1493: Include the installer in the list of files attached to GitHub releases
Sebastian mentioned this in Discord, currently we aren't attaching the installer to the release. We should.

neilenns:
Fixes #1493


DocMoebiuz:
Since we don't bundle the release content with the installer, there is no real benefit in attaching it as a visible file next to the other artifacts which actually are release specific.

PR #1496, opened 2023-12-14
Interpolation supports float values

### LINKED ISSUE #1481: Unable to save changes to 7 segment display output config
**Describe the bug**
Certain output configs are unable to save after editing 7 segment display config.

**Steps To Reproduce**
1. Open the attached .mmc file
2. Edit the `OH DC Amps` output 
3. Set the module to one you have connected that has 7 segment displays on it
4. Change the number of digits from 3 to 8
5. Hit ok

**Expected behavior**
Dialog closes and changes are applied

**Actual behavior**
Nothing happens. Dialog doesn't close.

**Details**
 - MobiFlight Version: 10.0

**Additional context**
In the output window of the debugger you see this:

```
Exception thrown: 'System.Collections.Generic.KeyNotFoundException' in mscorlib.dll
Exception thrown: 'MobiFlight.ArcazeCommandExecutionException' in MFConnector.exe
```

[Found via Discord](https://discord.com/channels/608690978081210392/1183487663324528661)

[flightsim1.zip](https://github.com/MobiFlight/MobiFlight-Connector/files/13629513/flightsim1.zip)

DocMoebiuz:
fixes #1481 

The UI now supports floats again. The result is still an integer. This behaves like before.

- [x] Add a log entry if form validation failed - in some cases the validation errors are not displayed because they are on a different tab
- [x] The UI Form fields are parsed as `double`

### DIFF
--- a/UI/Dialogs/ConfigWizard.cs
+++ b/UI/Dialogs/ConfigWizard.cs
@@ -397,6 +397,7 @@ namespace MobiFlight.UI.Dialogs
             try {
                 if (!ValidateChildren())
                 {
+                    Log.Instance.log("The dialog cannot be closed. There are invalid values on some tab.", LogSeverity.Error);
                     return;
                 }
             } catch (System.InvalidOperationException ex)
--- a/UI/Panels/Modifier/InterpolationModifier/InterpolationMappingPanel.cs
+++ b/UI/Panels/Modifier/InterpolationModifier/InterpolationMappingPanel.cs
@@ -42,11 +42,11 @@ namespace MobiFlight.UI.Panels.Modifier.InterpolationModifier
 
         public Tuple<double, double> toConfig()
         {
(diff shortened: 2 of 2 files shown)


UI/Dialogs/ConfigWizard.cs
```diff
@@ -397,6 +397,7 @@ private void button1_Click(object sender, EventArgs e)
             try {
                 if (!ValidateChildren())
                 {
+                    Log.Instance.log("The dialog cannot be closed. There are invalid values on some tab.", LogSeverity.Error);
```
neilenns, line 400: Does the dialog just silently not close in this case? Can we pop a dialog up to the user?

neilenns (approved):

UI/Dialogs/ConfigWizard.cs
neilenns: Yeah, I figured, that's why I approved anyway. Thanks for the explanation.

PR #1501, opened 2023-12-14
Fix cross thread exception on update of hubhop presets

### LINKED ISSUE #1500: Cross thread exception on progress after starting update huphob presets older than 7 days
**Describe the bug**
Cross thread exception on progress after starting update huphob presets older than 7 days

**Steps To Reproduce**
Might happen after startup of mobiflight when the reminder to update the older then 7 days huphob presets appears. Probably might also happen when starting the update from the menu.

**Expected behavior**
Progress should be shown and presets updated.

**Actual behavior**
Progress does not show, exception does happen.

**Details**
 - MobiFlight Version: some dev version...

**Additional context**
image

Koseng:
fixes #1500

### DIFF
--- a/UI/Forms/ProgressForm.cs
+++ b/UI/Forms/ProgressForm.cs
@@ -1,20 +1,13 @@
 ﻿using MobiFlight.Base;
 using System;
-using System.Collections.Generic;
-using System.ComponentModel;
-using System.Data;
-using System.Drawing;
-using System.Linq;
-using System.Text;
-using System.Threading.Tasks;
 using System.Windows.Forms;
 
 namespace MobiFlight.UI.Forms
 {
     public partial class ProgressForm : Form
     {
-        delegate void VoidDelegate(); 
-        Timer timer = new Timer();
+        delegate void VoidDelegate();
+        System.Windows.Forms.Timer timer = new Timer();
 
         public ProgressForm()
         {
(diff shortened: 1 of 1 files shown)


UI/Forms/ProgressForm.cs
```diff
@@ -46,8 +42,7 @@ public void OnProgressUpdated(object sender, ProgressUpdateEvent e)
 
             if (e.Current == e.Total)
             {
-                if (this.InvokeRequired)
-                    this.Invoke(new EventHandler<ProgressUpdateEvent>(OnProgressCompleted), new object[] { sender, e });
+                this.Invoke(new EventHandler<ProgressUpdateEvent>(OnProgressCompleted), new object[] { sender, e });
```
DocMoebiuz, line 45: if this cannot be called from a different thread context (we checked above) do:

`OnProgressCompleted?.Invoke(sender, e)`

DocMoebiuz (changes requested):
see my comment

UI/Forms/ProgressForm.cs
Koseng: It's even simpler:
`OnProgressCompleted(sender, e);`

DocMoebiuz (changes requested):
Please use the `?.Invoke` style otherwise this will throw an exception in case no handler is registered

UI/Forms/ProgressForm.cs
DocMoebiuz: That's not a good practice of Invoking an EventHandler - it works when you're absolutely 100% sure that a handler was registered before. I didn't check if this is the case, but let's stick to common practice.

Koseng: There is no Event which can be called in ProgressForm. There only is the method.

DocMoebiuz (approved):
Thanks!

PR #1505, opened 2023-12-16
Better Custom Board Community integration

DocMoebiuz:
This PR improves the way how CustomBoards can be integrated.

- [x] All required files are in one folder which is dropped into the new "Community"-folder
- [x] There are different levels for board integration: 
  - [x] Core, for the official MobiFlight boards
  - [x] Partner, for official MobiFlight partners who are sponsoring the project
  - [x] Community, for all other contributions
- [x] All information about CustomBoards and CustomDevices are loaded dynamically from the Community folder
- [x] Detail information available for Partner boards
  - [x] Logo
  - [x] Links to website, documentation and support page - all optional
  - [x] Option to bundle a default device config which can be uploaded 
- [ ] Unit tests
- [ ] Documentation
- [ ] i18n


Boards/mfboard.schema.json
```diff
@@ -220,6 +220,13 @@
           "title": "Resetfirmwarefile",
           "description": "Firmware filename to reset the board. Deprecated. Use Info.ResetFirmwareFile instead.",
           "type": "string"
+        },
+        "HasDefaultDeviceConfig": {
+          "$id": "#root/Info/HasDefaultConfig",
+          "title": "HasDefaultConfig",
```
neilenns, line 226: Nice!

Community/MobiFlight/boards/mobiflight_GNC255_pico.board.json
```diff
@@ -0,0 +1,205 @@
...
+    "Info": {
+      "CanInstallFirmware": true,
+      "CanResetBoard": true,
+      "FirmwareBaseName": "mobiflight_gnc255_raspberrypico",
+      "FirmwareExtension": "uf2",
+      "FriendlyName": "MobiFlight GNC255 Pico",
+      "LatestFirmwareVersion": "2.5.0",
+      "MobiFlightType": "MobiFlight GNC255 Pico",
+      "ResetFirmwareFile": "reset.raspberry_pico_flash_nuke.uf2",
+      "CustomDeviceTypes": [
```
neilenns, line 26: Maybe call it `CustomDevices` instead?

UI/Panels/Settings/MobiFlightPanel.cs
```diff
@@ -420,6 +506,32 @@ private TreeNode mfModulesTreeView_initNode(MobiFlightModuleInfo moduleInfo, Tre
...
+                moduleNode.SelectedImageKey = moduleNode.ImageKey = imageKey;
+            }
+        }
+
+        /// <summary>
+        /// This generates the imageKey and also adds the image to the imageList
+        /// </summary>
+        /// <param name="board">The Board information</param>
+        /// <returns>the valid imageKey</returns>
+        private string generateImageKeyAndAddImageToImageList(Board board)
```
neilenns, line 525: This probably has a style warning from Visual Studio, it should start with a capital letter.

neilenns (changes requested):
This also needs to update after the json schema validation stuff merges.

MobiFlight/Board.cs
```diff
@@ -151,6 +184,16 @@ public class Info
         /// </summary>
         public List<String> CustomDeviceTypes = new List<string>();
 
+        /// <summary>
+        /// True if the board ships with a default config that can be auto-loaded after flashing.
+        /// </summary>
+        public Boolean HasDefaultDeviceConfig;
```
neilenns, line 190: Does this come from the JSON? Historically we've simply used the presence of a path/value to indicate "has". So in this case, somewhere there would be a `DefaultDeviceConfig` property that either has a path in it or doesn't. It's redundant to specify both the boolean and the path in a .json file.

MobiFlight/Board.cs
```diff
@@ -279,6 +322,36 @@ public class Board
         /// </summary>
         public UsbDriveSettings UsbDriveSettings;
 
+        /// <summary>
+        /// Base path for custom firmware
+        /// </summary>
+        public String BasePath;
```
neilenns, line 328: Will this be specified in the JSON? Or is it calculated? If possible I would try and calculate this based on where the config file is, or say it must be located side-by-side with the config file. Otherwise it's another path for the developer to specify and possibly get wrong.

MobiFlight/Board.cs
```diff
@@ -279,6 +322,36 @@ public class Board
...
+
+        /// <summary>
+        /// The image resource
+        /// </summary>
+        public Image BoardImage;
```
neilenns, line 333: You'll need to have two separate properties here when you merge in the JSON validation. One should be `BoardImagePath` and the other should be `BoardImage`, which is marked to not JSON serialize.

neilenns: Wait, never mind, I see that it's a fixed file name. This will just have to get marked to not serialize when you merge in the JSON schema validation.

MobiFlight/CustomDevices/CustomDevice.cs
```diff
@@ -131,6 +127,11 @@ public class CustomDevice
         /// </summary>
         public List<MessageType> MessageTypes = new List<MessageType>();
 
+        /// <summary>
+        /// Base path for custom firmware
+        /// </summary>
+        public String BasePath;
```
neilenns, line 133: Same comment as above. It'd be nice to calculate this instead of have it come from the JSON, unless I'm misunderstanding?

neilenns (changes requested):
See comments

MobiFlight/CustomDevices/CustomDevice.cs
DocMoebiuz: it's not coming from the JSON - it can be a readonly property

DocMoebiuz:
closing in favor of rebased PR #1507

PR #1507, opened 2023-12-17
UI: Custom Board & Community integration

DocMoebiuz:
This PR improves the way how CustomBoards can be integrated.

- [x] All required files are in one folder which is dropped into the new "Community"-folder
- [x] There are different levels for board integration: 
  - [x] Core, for the official MobiFlight boards
  - [x] Partner, for official MobiFlight partners who are sponsoring the project
  - [x] Community, for all other contributions
- [x] All information about CustomBoards and CustomDevices are loaded dynamically from the Community folder
- [x] Detail information available for Partner boards
  - [x] Logo
  - [x] Links to website, documentation and support page - all optional
  - [x] Option to bundle a default device config which can be uploaded 
- [x] Unit tests - new json schema are tested
- [x] i18n
- [ ] Documentation

### Custom board flash process
custom-board-flash

### Custom board information & options 
partner-integration


MobiFlight/BoardDefinitions.cs
```diff
@@ -84,8 +86,27 @@ public static Board GetBoardByMobiFlightType(String mobiflightType)
...
+                    if (File.Exists(logoPath) && board.Info.Community != null)
+                    {
+                        board.Info.Community.Logo = Image.FromFile(logoPath);
+                    }
+
+                    var boardLogoPath = $@"{boardPath}\board-logo.png";
+                    if (File.Exists(boardLogoPath))
+                    {
+                        board.BoardImage = Image.FromFile(boardLogoPath);
+                    }
```
neilenns, line 108: Thinking about this on my route, I'm not wild about this getting set here instead of being done as getters in the actual objects (like you're already doing for `PartnerLevel`). This stuff is really specific to the Board, not the class that handles loading files.

I took a pass at moving it to getters and doing `BoardImage` is easy since it's right on the main object and `BasePath` is accessible. I got stuck on doing it for the community logo though since that's buried two classes deep from the board and doesn't have access to `BasePath`. Any ideas how that could be done?

MobiFlight/Board.cs
```diff
@@ -279,6 +328,37 @@ public class Board : IMigrateable
...
+        /// The image resource
+        /// </summary>
+        public Image BoardImage;
+
+        /// <summary>
+        /// Returns the correct Partner Level
+        /// </summary>
+        /// <returns></returns>
+        [JsonIgnore]
+        public BoardPartnerLevel PartnerLevel { get {
```
neilenns, line 346: Formatting is off here. `get` and `{` should be on their own lines.

MobiFlight/Board.cs
```diff
@@ -279,6 +328,37 @@ public class Board : IMigrateable
...
+                if (BasePath == "" && Info.MobiFlightType.Contains("MobiFlight"))
+                    return BoardPartnerLevel.Core;
+
+                var partners = new List<string>() { 
+                    "kavSimulations",
+                    "miniCOCKPIT"
+                };
```
neilenns, line 353: I think this should move out to a top-level property on the class at the top of the file to make it easier to find in the future when needing to update the list. Right now it's hiding inside a method.

MobiFlight/Board.cs
neilenns, line 346: You could add unit tests for this property as well, to check that it gets set to one of the three enum values for every file loaded.

neilenns (changes requested):
Added some comments after noodling in the car on my route.

MobiFlight/BoardDefinitions.cs
DocMoebiuz: I put it on purpose there, because i didn't want to make the board class so smart... so far it was more like a simple Data Object filled with the info from the JSON.

Since the infos don't come from the JSON, i thought it would be better to have the Definition loader handle this extra logic and provide it from the outside.

I still like the idea.

MobiFlight/Board.cs
DocMoebiuz: Good idea!

MobiFlight/BoardDefinitions.cs
neilenns: You already have logic in the Board class that is beyond just JSON backed data: `PartnerLevel`. The code is inconsistent now, where some of the calculated properties happen on the fly while others are set by an outside class.

PR #1512, opened 2023-12-19
change SPD/HDG/ALT/VS encoder type

edilclyde:
changed the encoder type for SPD/HDG/ALT/VS 

From `2 detents (00,11)`  > `4 detents per cycle`  to sync with Fenix's rate

Type 3 > type 5


DocMoebiuz:
I want to wait with this change. I independently raised this issue with the miniFCU team, and they intentionally chose to use this behavior.

DocMoebiuz:
It's now included in 0.9.3 version.

Thanks anyways for the PR

PR #1517, opened 2023-12-21
Add settings to disable joystick and midi support

### LINKED ISSUE #1516: Add settings to disable joystick and midi support
- If midi devices are used by other applications, and not mobiflight, it is a good idea to deactivate midi support completely. Current situation in windows is, only one application can have an open communication channel to a midi device.
- If for whatever reasons a problem in context with joystick or midi support appears, there should be the possibility to completely deactivate joystick or midi support until a fix can be provided.

Have a setting in settings dialog to deactivate joystick support and another one to deactivate midi support.

Koseng:
fixes #1516 

- [x] Add a new tab "Peripherals" to the settings dialog
  - [x] Tab contains checkbox for enabling support for joystick & midi devices
  - [x] support is enabled by default
  - [x] detected joysticks and midi devices are displayed in list
- [x] i18n
- [x] documentation
  - [x] for Joysticks, add a section in https://github.com/MobiFlight/MobiFlight-Connector/wiki/Joysticks
  - [x] for midi-devices, add a section in https://github.com/MobiFlight/MobiFlight-Connector/wiki/Midi-Boards
- [x] ~~unit tests~~

v5
image


neilenns:
It feels odd adding an entire new tab just for two checkboxes.

Maybe we can do a slight refactor of the General tab and put them in there? The General tab has a bunch of really small groups (only 1-2 settings in each) that could be consolidated.

Rename the "Run options" section to something like "General options" and move the HubHop presets setting into it (getting rid of the HubHop group) and add these two new checkboxes there as well?

UI/Panels/Settings/JoystickMidiPanel.resx
```diff
@@ -0,0 +1,222 @@
...
+    <value>18, 34</value>
+  </data>
+  <data name="checkBoxJoystickSupport.Size" type="System.Drawing.Size, System.Drawing">
+    <value>203, 24</value>
+  </data>
+  <data name="checkBoxJoystickSupport.TabIndex" type="System.Int32, mscorlib">
+    <value>0</value>
+  </data>
+  <data name="checkBoxJoystickSupport.Text" xml:space="preserve">
+    <value>Disable Joystick-Support</value>
```
neilenns, line 162:
```suggestion
    <value>Disable joystick support</value>
```

UI/Panels/Settings/JoystickMidiPanel.resx
```diff
@@ -0,0 +1,222 @@
...
+    <value>18, 65</value>
+  </data>
+  <data name="checkBoxMidiSupport.Size" type="System.Drawing.Size, System.Drawing">
+    <value>175, 24</value>
+  </data>
+  <data name="checkBoxMidiSupport.TabIndex" type="System.Int32, mscorlib">
+    <value>1</value>
+  </data>
+  <data name="checkBoxMidiSupport.Text" xml:space="preserve">
+    <value>Disable Midi-Support</value>
```
neilenns, line 135:
```suggestion
    <value>Disable MIDI support</value>
```

UI/Dialogs/SettingsDialog.resx
```diff
@@ -501,14 +552,71 @@
...
+    <value>3, 3, 3, 3</value>
+  </data>
+  <data name="tabPage1.Size" type="System.Drawing.Size, System.Drawing">
+    <value>848, 849</value>
+  </data>
+  <data name="tabPage1.TabIndex" type="System.Int32, mscorlib">
+    <value>4</value>
+  </data>
+  <data name="tabPage1.Text" xml:space="preserve">
+    <value>Joystick &amp; Midi</value>
```
neilenns, line 595:
```suggestion
    <value>Joystick &amp; MIDI</value>
```

MIDI is an acronym that's written in all caps (using https://www.midi.org/ as guidance)

neilenns (changes requested):

Koseng:
> It feels odd adding an entire new tab just for two checkboxes.
> 
> Maybe we can do a slight refactor of the General tab and put them in there? The General tab has a bunch of really small groups (only 1-2 settings in each) that could be consolidated.
> 
> Rename the "Run options" section to something like "General options" and move the HubHop presets setting into it (getting rid of the HubHop group) and add these two new checkboxes there as well?

Hahahaha. I exactly did something like that and put the 2 settings in the general tab. Then I showed a Screenshot to Sebastian and he voted strongly to put it in a separate tab. 

To be fair the general tab already is kind of crowded and feels even more crowded with the two additional settings. Also I plan to soon add a simple MIDI exclusion list, like the COM exclusion possibility. And who knows which requirement will suddenly appear in the future, then we do not have any space issues. Therefore I'm know ok with the separate tab, and probably there will be more content in some time....

Thanks for all the comments, will change.

DocMoebiuz:
Please try to add a simple ListView each with just the names of the joysticks and midi boards respectively. Then in the near future we can add a checkbox and enable disable the individual device.

UI/Panels/Settings/JoystickMidiPanel.resx
Koseng: done

UI/Dialogs/SettingsDialog.resx
Koseng: done

UI/Panels/Settings/JoystickMidiPanel.resx
Koseng: done

neilenns (changes requested):
Can the user do anything with the joysticks listed in the two list boxes? If not I suggest setting the list boxes to disabled and naming them "Detected joysticks" and "Detected MIDI boards".

Otherwise it's kinda confusing and looks like I should be able to do something with them when I can't.

Koseng:
> Please try to add a simple ListView each with just the names of the joysticks and midi boards respectively. Then in the near future we can add a checkbox and enable disable the individual device.

Hmmm, accidentally took listbox instead of listview. Will try listview as well.

Koseng:
> Can the user do anything with the joysticks listed in the two list boxes? If not I suggest setting the list boxes to disabled and naming them "Detected joysticks" and "Detected MIDI boards".
> 
> Otherwise it's kinda confusing and looks like I should be able to do something with them when I can't.

I went for "Active Joysticks" instead of "Detected Joysticks" because if joystick support is disabled, or later on when we have the possibility to exclude single joysticks, MobiFlight might "detect" them but ignore them. Then I thought maybe it is better to write "Active" instead of "Detected". 
Probably it also depends on how later on we will do the exclusion. Try to do everything in one list, or two lists (active list and ignored list).
=> I'm open for the naming, if you guys think detected is better I'll rename. We then can also adjust the naming in the feature depending on how the dialog evolves.

I wanted to set the lists to disabled, but then scolling does not work any more. And I'm sure the user with 10 joysticks will come and complain. But I'll also test now ListView instead of ListBox, maybe it behaves differently.

Koseng:
> Can the user do anything with the joysticks listed in the two list boxes? If not I suggest setting the list boxes to disabled and naming them "Detected joysticks" and "Detected MIDI boards".
> 
> Otherwise it's kinda confusing and looks like I should be able to do something with them when I can't.

- I'll stay with ListBox, ListView is so much more complex and not necessary for that case.
- Ok I'll switched to detected
- Had just the idea to change the background-colour, now it looks like readonly :-). And nothing is selectable.

flyingpaul1980:
Hi, I am using the current Beta for MiniFCU support and i cannot find the "Peripherals" page in the setting. I would require to disable Midi support in Mobiflight. Any Ideas what i am doing wrong?

neilenns:
@flyingpaul1980 

> Hi, I am using the current Beta for MiniFCU support and i cannot find the "Peripherals" page in the setting. I would require to disable Midi support in Mobiflight. Any Ideas what i am doing wrong?

This feature isn't released yet, that's why you can't find it in the UI.

For assistance please start a thread in our Discord server in the #mobiflight channel. https://discord.gg/6XSKQmfX

DocMoebiuz (changes requested):
@Koseng, a checkbox is an indicator for "active" so I'd say 

- [ ] enable joystick support
- [ ] enable midi-device support

By default, this setting is active. People have provided the feedback that the positive confirmation is more intuitive (bad example apparently: stepper, auto-zero pin option )

I also propse to simply combine the checkbox with the list into one group box.

- Remove "Activation" group box.
- Have two group boxes: "Joystick settings", "Midi Settings"
- inside the box, each has 
  - the respective checkbox for enabling the support for it, 
  - their respective label that says "Detected joysticks" or "Detected midi devices", including 
  - their respective list of devices.

DocMoebiuz (approved):

JaimeLeon2:
I think individual MIDI device activation in Mobiflight will be a useful feature for the case where a user has two devices connected but only wishes to use one in Mobiflight.

Koseng:
> I think individual MIDI device activation in Mobiflight will be a useful feature for the case where a user has two devices connected but only wishes to use one in Mobiflight.

Yes, exactly that is the plan for the next step. Instead of "Detected MIDI devices" there will be something llike "Active MIDI devices" and the list will get Checkboxes to select indivdual ones.

PR #1520, opened 2023-12-21
Individually include the files via Visual Studio UI

### LINKED ISSUE #1519: Build process blindly copies everything in the Community folder, even if it isn't part of the project
To reproduce:

1. Use Visual Studio's UI to include some custom board files under the Community folder of the MobiFlightConnector project
2. Do not set the `Copy to Output Directory` property
3. Build

Result: The files you added are copied to the output folder
Expected result: They shouldn't be copied since they were not set to copy to output

This happens because there's a wildcard ItemGroup in the project file:

```XML
  <ItemGroup>
    <Content Include="Community\**">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
    </Content>
  </ItemGroup>
```

I lost a good half an hour of my day trying to figure out why files were getting copied to the build folder when I wasn't expecting them to.

There's UI in Visual Studio specifically for managing files, we should be using that instead of magically blanket including the files via an ItemGroup that can only be managed by editing the .csproj file directly.

neilenns:
Fixes #1519 

* Remove the wildcard ItemGroup from the project file
* Explicitly include the various files using the VS UI designed for this purpose

### DIFF
--- a/MobiFlightConnector.csproj
+++ b/MobiFlightConnector.csproj
@@ -1000,6 +1000,90 @@
     </None>
     <None Include="Build\build-firmware.bat" />
     <None Include="Build\release.bat" />
+    <None Include="Community\kavSimulations\boards\kav_mega.board.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+    <None Include="Community\kavSimulations\boards\kav_pico.board.json">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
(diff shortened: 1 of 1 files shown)


DocMoebiuz:
we have to find a different way that doesn't include manual edits of the solution for every hex file etc

neilenns:
I know you did it on purpose but it's wrong. I lost so much time today because it was automatically copying files when I was trying to debug problems in a specific file without the others showing up.

Back when I worked on msbuild we never, ever, advocated customers use wildcards like this because of the future grief it will cause. The short term pain of having to use the VS UI is well worth it compared to the future anguish we will have.

PR #1524, opened 2023-12-22
Remove analog pin labels from Pro Micro board

### LINKED ISSUE #1522: Pro Micro support for pins A6-A10 does not correspond to the board label
**Describe the bug**
The support for pins labeled A6 through A10 on the Pro Micro should be changed to pins 4, 6, 8, 9 and 10

**Steps To Reproduce**
PR #1236 was made changing these labels in the json file.  
They are analog capable pins but the board labeling does not use those names.

**Expected behavior**
Dropdown in Mobiflight should show the board labels used on the board as pins 4, 6, 8, 9, and 10.  Otherwise, there is no way to tell which are those pins, unless you have a board pinout diagram in front of you.

**Actual behavior**
Dropdown menu shows labels A6, A7, A8, A9 and A10 which do not correspond to any pin label on the board silkscreen.

**Details**
 - MobiFlight Version: 10.0.2
 - FlightSim: irrelevant
 - AddOn: irrelevant

**Additional context**
nothing

neilenns:
Fixes #1522

Remove the analog pin labels that shouldn't have been added in #1236. Dropdown now looks like this:

image

### DIFF
--- a/Boards/arduino_micro.board.json
+++ b/Boards/arduino_micro.board.json
@@ -74,8 +74,7 @@
       "isAnalog": true,
       "isI2C": false,
       "isPWM": false,
-      "Pin": 4,
-      "Name": "A6"
+      "Pin": 4
     },
     {
       "isAnalog": false,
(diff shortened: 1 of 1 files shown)


JaimeLeon2:
The fix looks great. Thank you Neil for the quick response.

PR #1530, opened 2023-12-25
Add board type information to mfmc file

### LINKED ISSUE #1521: mfmc files contain board type
with all the different boards and more and more custom boards created, we want to be able to check for compatibility of a mfmc file and the board

Add an xml tag with the MobiFlight Type and check for compatibility on loading

nothing else makes sense.

DocMoebiuz:
fixes #1521 

- [x] On opening - Message popping up to inform when no board type is contained in file.
  image
- [x] On opening - Message popping up to inform that there is a board type mismatch.
  image
- [x] On opening - In both cases user can continue with "Yes" or stop opening with "No"
- [x] On saving - board type is saved as new property
- [ ] Documentation - Add info to https://github.com/MobiFlight/MobiFlight-Connector/wiki/Mobiflight-Connector-Main-Window#mobiflight-modules---tab
- [x] i18n
- [x] unit tests
  - [x] extract methods for open and save logic
  - [x] add unit tests for open and save
  - [x] add missing `equals` implementations in various classes
  - [x] added ToString implementations in various classes
  - [x] refactored ToString method to new style
- [x] remove all unnecessary usings in config namespace

### DIFF
--- a/MobiFlight/Config/Config.cs
+++ b/MobiFlight/Config/Config.cs
@@ -7,7 +7,8 @@ namespace MobiFlight.Config
 {
     public class Config
     {
-        public String ModuleName = "";
+        public string ModuleType = "";
+        public string ModuleName = "";
         public int PowerSavingTime = 60 * 10; // => 10 Minutes Default
 
         [XmlElement(typeof(Button))]
--- a/MobiFlight/Config/Config.cs
+++ b/MobiFlight/Config/Config.cs
@@ -1,6 +1,8 @@
 ﻿using MobiFlight.Config.Compatibility;
 using System;
 using System.Collections.Generic;
+using System.IO;
+using System.Windows.Forms;
 using System.Xml.Serialization;
 
 namespace MobiFlight.Config
@@ -216,6 +218,23 @@ namespace MobiFlight.Config
 
             }
             return this;
+        }
+
+        public static Config LoadFromFile(string fileName)
+        {
+            TextReader textReader = new StreamReader(fileName);
+            XmlSerializer serializer = new XmlSerializer(typeof(Config));
+            var config = (Config)serializer.Deserialize(textReader);
+            textReader.Close();
+            return config;
+        }
+
+        public void SaveToFile(string fileName)
+        {
+            XmlSerializer serializer = new XmlSerializer(typeof(MobiFlight.Config.Config));
+            TextWriter textWriter = new StreamWriter(fileName);
+            serializer.Serialize(textWriter, this);
+            textWriter.Close();
         }
     }
 }
--- a/MobiFlight/Config/Config.cs
+++ b/MobiFlight/Config/Config.cs
@@ -2,7 +2,6 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Windows.Forms;
 using System.Xml.Serialization;
 
 namespace MobiFlight.Config
--- a/MobiFlight/Config/InputMultiplexer.cs
+++ b/MobiFlight/Config/InputMultiplexer.cs
@@ -65,19 +65,27 @@ namespace MobiFlight.Config
             NumBytes    = paramList[6];
             Name        = paramList[7];
 
-            /* TO BE DELETED before PR is accepted:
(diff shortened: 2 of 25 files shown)


neilenns:
I love all the updates to the ToString() methods but why are they in this PR?

UI/Panels/Settings/MobiFlightPanel.cs
```diff
@@ -1074,15 +1070,33 @@ private void openToolStripButton_Click(object sender, EventArgs e)
...
+            var module = (moduleNode.Tag as MobiFlightModule);
+
+            if (newConfig?.ModuleType != module.Type)
+            {
+                // Old configs don't contain a specified type 
+                // In this case newConfig.ModuleType == "" 
+                var message = string.Format(i18n._tr("uiMessageOpenConfigUnspecificTypeText"), module.Type);
+
+                // New configs have a specific type other than ""
+                if ((newConfig?.ModuleType ?? "") != "")
```
neilenns, line 1083: It's a bit of a style thing, but I prefer using `String.IsNullOrEmpty()` when testing strings for empty values instead of "".

MobiFlight/Config/Config.cs
```diff
@@ -216,5 +218,22 @@ public Config FromInternal(String value, bool throwException = false)
...
+            return config;
+        }
+
+        public void SaveToFile(string fileName)
+        {
+            XmlSerializer serializer = new XmlSerializer(typeof(MobiFlight.Config.Config));
+            TextWriter textWriter = new StreamWriter(fileName);
+            serializer.Serialize(textWriter, this);
+            textWriter.Close();
+        }
```
neilenns, line 237: Should there be any exception handling around any of this?

MobiFlight/Config/InputMultiplexer.cs
```diff
@@ -65,19 +65,27 @@ override public bool FromInternal(String value)
...
+                return false;
+            }
+
+            return this.Name == other.Name
+                && this.DataPin == other.DataPin
+                && this.NumBytes == other.NumBytes
+                && this.Selector.Equals(other.Selector)
+                ;
+        }
+
```
neilenns, line 85: Why is this in this PR? It's not related to board type stuff at all.

MobiFlight/Config/InputShiftRegister.cs
```diff
@@ -46,9 +46,25 @@ override public bool FromInternal(String value)
...
+            }
+
+            return this.Name == other.Name
+                && this.LatchPin == other.LatchPin
+                && this.ClockPin == other.ClockPin
+                && this.DataPin == other.DataPin
+                && this.NumModules == other.NumModules
+                ;
+        }
+
```
neilenns, line 64: Why is this in this PR?

neilenns:
Are you sure you did this work on the correct branch? There's a bunch of unrelated changes in it.

neilenns (changes requested):
Will comment more in discord

MobiFlight/Config/Config.cs
DocMoebiuz: there was none so far

UI/Panels/Settings/MobiFlightPanel.cs
DocMoebiuz: good point. they introduced that

DocMoebiuz: It reads better, yet I don't understand why some of these are not simply an extension method.

DocMoebiuz:
image

 What's a better wording?

MobiFlight/Config/Config.cs
neilenns: Odd.

neilenns:
> image
> What's a better wording?

Maybe

> This file does not have a board type associated with it and you are trying to load it on a {board type}. It may be incompatible, are you sure you want to do this?

DocMoebiuz:
The file doesn't contain any board type information. It might be incompatible. Your board type is "....".

I think ideally we could tell the user that it will be saved next time. Just so that they know how to get rid of this message.

DocMoebiuz:
Save the file again to associate it with the current board type.

PR #1536, opened 2023-12-30
Auto width for Module, Input, Output and Type

### LINKED ISSUE #1533: Remember grid view column widths across app launches
It's possible to resize the column widths on the Outputs and Inputs tab, but those widths aren't persisted across app launch. When you get them sized the way you like you'll lose it when you quit Mobiflight then re-launch.

Remember the column widths across app launch.

**Additional context**
[Requested by Glenn in Discord](https://discord.com/channels/608690978081210392/745361245808492584/1189360726964437062)

I asked ChatGPT and this is totally possible to do without too much effort.

Koseng:
Fixes #1533. Or better, fixes the root cause for the wish of #1533.

I'm also annoyed by the truncated or wrapped texts and all the time need to adjust the columns, so everything is visible.

I propose to just auto size Module, Input, Output and Type based on the content. Why not? Try it.

Opinions @neilenns , @DocMoebiuz ?

### DIFF
--- a/UI/Panels/InputConfigPanel.Designer.cs
+++ b/UI/Panels/InputConfigPanel.Designer.cs
@@ -231,7 +231,7 @@
             // 
             // moduleSerial
             // 
-            this.moduleSerial.AutoSizeMode = System.Windows.Forms.DataGridViewAutoSizeColumnMode.Fill;
+            this.moduleSerial.AutoSizeMode = System.Windows.Forms.DataGridViewAutoSizeColumnMode.AllCells;
             this.moduleSerial.DataPropertyName = "moduleSerial";
             dataGridViewCellStyle1.BackColor = System.Drawing.SystemColors.ControlLight;
             this.moduleSerial.DefaultCellStyle = dataGridViewCellStyle1;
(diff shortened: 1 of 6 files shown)


neilenns:
Ooooh. I'll give this a try later today, could be a lot easier than what I was looking at doing (storing the manually set widths in the config files).

DocMoebiuz:
I tried with a bigger config and the resource consumption goes up a lot. UI is not responsive and takes several seconds until the config is loaded. Maybe first load the config, and then set the columns to auto-width. It might re-calcualte it otherwise multiple times while loading the data.

DocMoebiuz:
[aero2023-combined-display.mcc.txt](https://github.com/MobiFlight/MobiFlight-Connector/files/13798207/aero2023-combined-display.mcc.txt)

DocMoebiuz (changes requested):
see my attached config and the comment about performance on opening the file.

Koseng:
Good catch, now performance fix included.

DocMoebiuz:
> Good catch, now performance fix included. Does that fix it for you?

yes. it opens the file like before.

DocMoebiuz (approved):

PR #1538, opened 2023-12-30
Fix erroneous texts in german general settings dialog

### LINKED ISSUE #1537: Erroneous texts in german general settings dialog
**Describe the bug**
Erroneous texts in german general settings dialog

image

**Steps To Reproduce**
1. Set language to german
2. Open general settings dialog

**Expected behavior**
Everything should be shown in readable geman.

**Actual behavior**
Some text in english, some texts not visible.

**Details**
 - MobiFlight Version: 10.0.2.1

Koseng:
fixes #1537 

image

### DIFF
--- a/UI/Panels/Settings/GeneralPanel.de.resx
+++ b/UI/Panels/Settings/GeneralPanel.de.resx
@@ -164,17 +164,11 @@
   <data name="label8.Size" type="System.Drawing.Size, System.Drawing">
     <value>124, 39</value>
   </data>
-  <data name="label8.Text" xml:space="preserve">
-    <value>Sehr schnell</value>
-  </data>
   <data name="label8.TextAlign" type="System.Drawing.ContentAlignment, System.Drawing">
     <value>MiddleRight</value>
   </data>
   <data name="label6.Location" type="System.Drawing.Point, System.Drawing">
     <value>8, 63</value>
-  </data>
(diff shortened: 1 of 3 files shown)


UI/Panels/Settings/GeneralPanel.de.resx
```diff
@@ -164,18 +164,12 @@
   <data name="label8.Size" type="System.Drawing.Size, System.Drawing">
```
DocMoebiuz, line 164: Ideally the localization resource should not contain any size or location values. Why? Because if you move things around in the default language, it will reflect correctly in the localized version too.

I did that wrong in the past. Instead of positioning or re-sizing elements specfically for a language, I started to make labels just a bit wider in the default language and set auto-size property to false. 

Ideally you remove all `.Size`  and `.Location` properties from the `.de.resx`-file

DocMoebiuz: We can keep it the way it is this time, I will leave it up to you.

DocMoebiuz (approved):

DocMoebiuz:
When I open `UI\Panels\Settings\GeneralPanel.de.resx` and select `Other` from the resource type options... and when I remove all rows, then it still looks correct and we don't have any DE-specific positioning/size infos.

It takes 30 secs to do that.

Please do it.

PR #1541, opened 2023-12-31
Manually adjustable config columns after auto resize

### LINKED ISSUE #1540: Manually adjustable config columns after auto resize
I want to be able to manually adjust the width of the config columns after they were auto sized.

I want to be able to manually adjust the width of the config columns after they were auto sized.

**Additional context**
Add on to #1536

Koseng:
Alternative to #1536. Here the columns have relative sizes and the manual resizing behavior is closer to what it was before. But probably not needed.

We wait for feedback to #1536

### DIFF
--- a/UI/Panels/InputConfigPanel.Designer.cs
+++ b/UI/Panels/InputConfigPanel.Designer.cs
@@ -231,7 +231,7 @@
             // 
             // moduleSerial
             // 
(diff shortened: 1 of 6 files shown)


Koseng:
Closing it, because no complaints to the new behavior so far.

PR #1550, opened 2024-01-01
Trim dataref and command paths

### LINKED ISSUE #1424: x-plane dataref strings could trim off for possible trailing spaces
It's quite easy to copy and paste an X-plane dataref or command string and accidentally have a space or enter in the code field. 

This causes the dataref to fail.

Mobiflight should probably check for any trailing whitespace and trim the string to make it less prone to user error.

DocMoebiuz:
fixes #1424 

- [x] add trim to saving and loading the `.Path` value

### DIFF
--- a/UI/Panels/Config/HubHopPresetPanel.cs
+++ b/UI/Panels/Config/HubHopPresetPanel.cs
@@ -233,7 +233,7 @@ namespace MobiFlight.UI.Panels.Config
                 config.SimConnectValue.Value = SimVarNameTextBox.Text.ToLF();
             } else if (FlightSimType == FlightSimType.XPLANE)
             {
-                config.XplaneDataRef.Path = SimVarNameTextBox.Text;
+                config.XplaneDataRef.Path = SimVarNameTextBox.Text.Trim();
             }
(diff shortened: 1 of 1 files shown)


DocMoebiuz:
@tigert - this should fix your issue

UI/Panels/Config/HubHopPresetPanel.cs
```diff
@@ -233,7 +233,7 @@ internal void syncToConfig(OutputConfigItem config)
                 config.SimConnectValue.Value = SimVarNameTextBox.Text.ToLF();
```
Koseng, line 233: Also add .Trim() for MSFS? Won't hurt.

DocMoebiuz: wasn't sure. i will check with Jaime

Koseng (commented):
In general the .Trim() could also make sense on MSFS for manually entered presets.

Lines: 233, 256

DocMoebiuz:
I didn't do the MSFS2020 change, I was worried to break something :D

Issue #1419, opened 2023-11-18
Improve joystick & midi device integration


DocMoebiuz:
**Is your feature request related to a problem? Please describe.**
MobiFilght supports joysticks and that is a great feature. Home cockpit builders have typically multiple joystick / hid devices connected. All produce events that MobiFlight processes, e.g., checks for input configs, etc.

**Describe the solution you'd like**
Better joystick integration means:
* Show a list of joysticks which are supported by Mobiflight, compare to MobiFlight board tree view
* Allow for disabling active support by MobiFlight so that a joystick which is not configured through MobiFlight also is not processed by MobiFlight - compare this to "ignore ports"
* Support dynamic connect and disconnect after startup of MobiFlight, compare to dynamic detection of MobiFlight boards

**Additional context**
Same applies to Midi devices

neilenns:
Also this! https://github.com/MobiFlight/MobiFlight-Connector/issues/1198 :D

PR #1555, opened 2024-01-02
Add possibility to exclude single joysticks or midi devices

### LINKED ISSUE #1554: Add possibility to exclude single joysticks or midi devices
If a midi device is controlled by another application, it must be excluded from mobiflight. 
Also unneeded joysticks can be excluded.

Add possibility to exclude single joysticks or midi devices in peripherals settings dialog.

**Additional context**
Is related to #1419.

Koseng:
fixes #1554 
related to #1419 
- [x] i18n
- [x] ~~unit tests~~
- [ ] documentation

image

### DIFF
--- a/MobiFlight/Joysticks/JoystickManager.cs
+++ b/MobiFlight/Joysticks/JoystickManager.cs
@@ -130,7 +130,7 @@ namespace MobiFlight
             var di = new SharpDX.DirectInput.DirectInput();
             Joysticks?.Clear();
             ExcludedJoysticks?.Clear();
-            string[] excludedJoysticks = Properties.Settings.Default.ExcludedJoysticks.Split('§');            
+            List<string> settingsExcludedJoysticks = JsonConvert.DeserializeObject<List<string>>(Properties.Settings.Default.ExcludedJoysticks);
 
             var devices = di.GetDevices(DeviceClass.GameControl, DeviceEnumerationFlags.AttachedOnly).ToList();
 
@@ -166,7 +166,7 @@ namespace MobiFlight
                 }
 
                 // Check against exclusion list
-                if (excludedJoysticks.Contains(js.Name))
+                if (settingsExcludedJoysticks.Contains(js.Name))
                 {
                     Log.Instance.log($"Ignore attached joystick device: {js.Name}.", LogSeverity.Info);
                     ExcludedJoysticks.Add(js);
--- a/MobiFlight/Joysticks/JoystickManager.cs
+++ b/MobiFlight/Joysticks/JoystickManager.cs
@@ -98,6 +98,10 @@ namespace MobiFlight
         public void Shutdown()
         {
             PollTimer.Stop();
+            foreach (var js in Joysticks)
+            {
+                js.Shutdown();
+            }
             Joysticks.Clear();
             ExcludedJoysticks.Clear();
         }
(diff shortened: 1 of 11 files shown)


MobiFlight/Joysticks/JoystickManager.cs
```diff
@@ -121,7 +128,9 @@ public void SetHandle(IntPtr handle)
         public void Connect()
         {
             var di = new SharpDX.DirectInput.DirectInput();
-            joysticks?.Clear();
+            Joysticks?.Clear();
+            ExcludedJoysticks?.Clear();
+            string[] excludedJoysticks = Properties.Settings.Default.ExcludedJoysticks.Split('§');
```
DocMoebiuz, line 133: we never used § as a delimeter yet. You can actually serialize the data structure like I did in MainForm for the autoload-configs... i think that is way nicer.

see MainForm > RestoreAutoLoadConfig & SaveAutoLoadConfig

UI/Panels/Settings/PeripheralsPanel.cs
```diff
@@ -16,50 +18,81 @@ public PeripheralsPanel()
...
+            bool enableJoysticksChanged = Properties.Settings.Default.EnableJoystickSupport != checkBoxJoystickSupport.Checked;
+            bool enableMidiChanged = Properties.Settings.Default.EnableMidiSupport != checkBoxMidiSupport.Checked;
+
+            Properties.Settings.Default.ExcludedMidiBoards = string.Join("§", excludedBoards);
+            Properties.Settings.Default.ExcludedJoysticks = string.Join("§", excludedJoysticks);
+            Properties.Settings.Default.EnableJoystickSupport = checkBoxJoystickSupport.Checked;
+            Properties.Settings.Default.EnableMidiSupport = checkBoxMidiSupport.Checked;
+
+            // Update joystick management
+            if (!checkBoxJoystickSupport.Checked && enableJoysticksChanged)
```
DocMoebiuz, line 77: Ideally these actions should not happen withing the settingsdialog
We have a settings changed event handler that could handle this. We can do that at a later point... but it makes sense as refactoring.

DocMoebiuz (changes requested):

MobiFlight/Joysticks/JoystickManager.cs
Koseng: yeah that's nicer. Done.

UI/Panels/Settings/PeripheralsPanel.cs
Koseng: Yes, with settings changed event is better. I have put it on my todo list for later.

MobiFlight/Joysticks/JoystickManager.cs
```diff
@@ -97,20 +98,26 @@ public void Startup()
         public void Shutdown()
         {
             PollTimer.Stop();
-            joysticks.Clear();
+            Joysticks.Clear();
```
DocMoebiuz, line 101: I think Shutdown should call Stop for all Joysticks to make sure, that they can do their specific stuff. Like turning off the lights or disconnecting from their stream (Octavi)

DocMoebiuz: But tbh, we have to provide a Shutdown or Disconnect method for HID devices. In the case of Octavi it opens it's own stream. That's why it keeps sending events even after Shutdown of the JoystickManager

DocMoebiuz (changes requested):
See my comments.

DocMoebiuz (approved):
Thanks!

PR #1563, opened 2024-01-04
Firmware update supports concurrent updates with individual base path

### LINKED ISSUE #1562: FirmwareUpdater doesn't support concurrent execution with specific Community Firmware Path
**Describe the bug**
MobiFlightFirmwareUpdater is a static class and has a static property `FirmwarePath`. This was only the same path for all firmware in the past. With the new Community Folder, the firmware path is specific to every project. When there are concurrent firmware updates, the static property doesn't work anymore.

**Steps To Reproduce**
1. Have two community boards with new firmware
2. Start MobiFlight Connector
3. Confirm the message that you want to update all boards

**Expected behavior**
All boards update with their respective firmware.

**Actual behavior**
An error message that the update failed. The error log shows that one firmware file was not found because it was using the basepath of the other firmware.

**Details**
 - MobiFlight Version: 10.0.2.1

DocMoebiuz:
fixes #1562 

- [x] Remove the static `FirmwarePath` property
- [x] Determine the correct base path for every RunAvrDude

### DIFF
--- a/MobiFlight/MobiFlightFirmwareUpdater.cs
+++ b/MobiFlight/MobiFlightFirmwareUpdater.cs
@@ -14,11 +14,6 @@ namespace MobiFlight
         public static String ArduinoIdePath { get; set; }
         public static String AvrPath { get { return "hardware\\tools\\avr"; } }
 
-        /***
-         * C:\\Users\\SEBAST~1\\AppData\\Local\\Temp\\build2060068306446321513.tmp\\cmd_test_mega.cpp.hex
-         **/
-        public static String FirmwarePath { get; set; }
-
         public static bool IsValidArduinoIdePath(string path)
         {
             return Directory.Exists(path + "\\" + AvrPath);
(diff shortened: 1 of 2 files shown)


MobiFlight/MobiFlightFirmwareUpdater.cs
```diff
@@ -132,7 +129,7 @@ public static void RunAvrDude(String port, Board board, String firmwareName)
 
                 var attempts = board.AvrDudeSettings.Attempts != null ? $" -x attempts={board.AvrDudeSettings.Attempts}" : "";
                 string anyCommand =
-                    $@"-C""{Path.Combine(FullAvrDudePath, "etc", "avrdude.conf")}""{verboseLevel}{attempts} -p{board.AvrDudeSettings.Device} -c{board.AvrDudeSettings.Programmer} -P{port} -b{baudRate} -D -Uflash:w:""{FirmwarePath}\{firmwareName}"":i";
+                    $@"-C""{Path.Combine(FullAvrDudePath, "etc", "avrdude.conf")}""{verboseLevel}{attempts} -p{board.AvrDudeSettings.Device} -c{board.AvrDudeSettings.Programmer} -P{port} -b{baudRate} -D -Uflash:w:""{fullFirmwarePath}"":i";
```
DocMoebiuz, line 132: this is actually not nice... I'd prefer a dedicated variable for this.

MobiFlight/MobiFlightFirmwareUpdater.cs
```diff
@@ -124,15 +121,17 @@ public static void RunAvrDude(String port, Board board, String firmwareName)
             //verboseLevel = " -v -v -v -v";
 
             // Process() requires an absolute, rather than relative, path when using UseShellExecute = false
-            String FullAvrDudePath = Path.GetFullPath(Path.Combine(ArduinoIdePath, AvrPath));
+            var FullAvrDudePath = Path.GetFullPath(Path.Combine(ArduinoIdePath, AvrPath));
```
neilenns, line 124: 🥰

neilenns (approved):

PR #1580, opened 2024-01-09
Fix substring modifier implementation

### LINKED ISSUE #1558: Substring modifier returns one less character
**Describe the bug**
New substring modifier returns incorrect number of characters

**Steps To Reproduce**
1. Set a string value 'ABCDEFGH' to an output config
2. Add Substring modifier and select from 0 to 5
3. Display value in 7 segment display
4. Note the string displayed has 5 characters, not 6

image

**Expected behavior**
The modifier default position says 0, so counting from 0 the modifier should return one more than the number of the end position.

**Actual behavior**
The modifier returns one less than the expected number of characters.

**Details**
 - MobiFlight Version: 10.0.2.2
 - FlightSim: irrelevant
 - AddOn: irrelevant

**Additional context**
Maybe the dialog should say starting position: 0 and "Number of character: 5"
Then the behavior would be consistent.  
If I remember correctly, this is how the substring function was in the Fsuipc output config dialog.

DocMoebiuz:
fixes  #1558 

This PR corrects the behavior of substring method. A substring has a Start, and has an End value.
The old behavior: End value was not included. The substring "starts" at Start and "ends **before**" End.
The new behavior: End value is included. The substring "starts" at Start and "ends **on**" End.

- [x] Substring Start and End mark the characters that will be included with the result
- [x] Add unit tests

> [!NOTE]  
> this PR also makes sure that the ConnectorValue is properly cloned and used as the result.


neilenns:
What happens to people who have existing modifiers? Do they get auto-migrated to match this new behaviour? Or does everyone have to go and manually clean them up?

neilenns:
> i know. it's not ideal. but i have no good idea to tell whether it's an old config or not.

All we really need to know is if this specific modifier is old or new. You can add a specific property to just the modifier object to identify that, something like `endson=true`. If it's missing then you know it's the old style and can auto-update it so it doesn't break the config, and then mark the file as dirty. On save, add the `endson=true` (or whatever flag name/value you want).

We had a LOT of problems in discord with the change to reverse digits. We should try and find a way to avoid the same thing happening again with this change :)

PR #1601, opened 2024-01-16
Only prompt for WASM update if it changed

### LINKED ISSUE #1594: Show WASM installation dialog only when update is needed
On every MobiFilght version update, on first start, there is a dialog requesting to install the WASM module. It even happens, when the WASM module has not changed in the new version. This could be improved.

Check the version of the installed WASM module, and only present the Update dialog if the WASM module version that comes with MobiFlight is newer.

neilenns:
Fixes #1594 

* Use the existing `WasmModulesAreDifferent()` method as a condition of prompting on first launch
* Update `WasmModulesAreDifferent()` so it no longer tests the events.txt file. 
* Use `Path.Combine()` to create the path to the wasm module

To test this you'll need to delete your `user.config` file for mobiflight, which is located in `%localappdata%\MobiFlight` and then in one of the gibberish folders with the matching app version.

### DIFF
--- a/SimConnectMSFS/WasmModuleUpdater.cs
+++ b/SimConnectMSFS/WasmModuleUpdater.cs
@@ -155,22 +155,19 @@ namespace MobiFlight.SimConnectMSFS
 
             string installedWASM;
             string mobiflightWASM;
-            string installedEvents;
-            string mobiflightEvents;
 
             if (CommunityFolder == null) return true;
 
-
-            if (!File.Exists(CommunityFolder + $@"\mobiflight-event-module\modules\{WasmModuleName}"))
+            string wasmModulePath = Path.Combine(CommunityFolder, @"mobiflight-event-module\modules\", WasmModuleName);
+            if (!File.Exists(wasmModulePath))
+            { 
                 return true;
+            }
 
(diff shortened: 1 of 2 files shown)


SimConnectMSFS/WasmModuleUpdater.cs
```diff
@@ -155,22 +155,19 @@ public bool WasmModulesAreDifferent()
...
+            if (!File.Exists(wasmModulePath))
+            { 
                 return true;
+            }
 
             installedWASM = CalculateMD5(CommunityFolder + $@"\mobiflight-event-module\modules\{WasmModuleName}");
             mobiflightWASM = CalculateMD5($@".\MSFS2020-module\mobiflight-event-module\modules\{WasmModuleName}");
 
-            installedEvents = CalculateMD5(CommunityFolder + $@"\mobiflight-event-module\modules\{WasmEventsTxtFile}");
-            mobiflightEvents = CalculateMD5($@".\MSFS2020-module\mobiflight-event-module\modules\{WasmEventsTxtFile}");
```
neilenns, line 171: This was causing a crash locally because the file doesn't exist anymore in a normal install. Removed since it's no longer needed.

PR #1623, opened 2024-01-24
Add a copy logs command

neilenns:
Fixes #1662

Add a "Copy logs to clipboard" command to the extras menu. Correctly handles the case of no log file being available, or other exceptions that might come up.

Animation


Base/Log.cs
```diff
@@ -164,6 +164,20 @@ public LogAppenderFile()
                 File.Delete(FileName);
         }
 
+        public void CopyToClipboard()
```
neilenns, line 167: I put this here instead of in MainForm.cs so the knowledge of where the file is on disk is contained within the log class.

UI/MainForm.cs
```diff
@@ -67,10 +68,11 @@ private void InitializeUILanguage()
 
         public event EventHandler<ConfigFile> ConfigLoaded;
 
+        private readonly LogAppenderFile logAppenderFile = new LogAppenderFile();
```
neilenns, line 71: This moved out of InitializeLogging() so it's available in the menu click handler.

UI/MainForm.cs
```diff
@@ -20,6 +20,7 @@
 using System.Threading.Tasks;
 using MobiFlight.InputConfig;
 using Newtonsoft.Json;
+using System.IO;
```
neilenns, line 23: Required for FileLoadException

PR #1624, opened 2024-01-24
Add support for hex values in the joystick json files

### LINKED ISSUE #1621: Joystick schema states VendorID is a number, but T.A320 Pilot is
**Describe the bug**
Trying to add a json for Thrustmaster A320 Pilot stick fails in MF complaining the vendor ID should be a number. However google thinks they are hex values:

- T.A320 Pilot USB VID 044F & PID 0405
- T.A320 Copilot USB VID 044F & PID 0406

[This link](https://www.driverscloud.com/en/drivers/manufacturercat/622-15-1/thrustmaster-usb-devices) shows a lot of vid+pid id's for various usb devices, useful.

**Steps To Reproduce**
1. Add json
2. Start MF
3. Look at log

**Expected behavior**
Hexadecimal PID and VID values should be handled properly.

**Actual behavior**
Fails to load the joystick definition

**Details**
 - MobiFlight Version: 10.1.1

**Additional context**

According to googling, there are two possibilities to fix this: 

- Store the VID and PID as strings and parse them in the code
- Use JSON5 schema that supports hexadecimal numbers, among other things.

**Error in MF Log:**
```01/24/2024 15:48:51(910): JsonBackedObject.LoadDefinitions(): 
Unable to load Joysticks\t.a320_pilot.joystick.json: 
Input string '044F' is not a valid integer. Path 'VendorId', line 4, position 18.```

neilenns:
Fixes #1621 

* Add a new JsonConverter that can handle both integers and strings. If the string starts with `0x` it is treated as a hex value and converted to a number appropriately.
* Update the schema to specify the Joystick VID/PID values are hexadecimal strings starting with `0x`.

### DIFF
--- a/Joysticks/honeycomb_bravo.joystick.json
+++ b/Joysticks/honeycomb_bravo.joystick.json
@@ -1,12 +1,12 @@
 {
   "$schema": "./mfjoystick.schema.json",
   "InstanceName": "Bravo Throttle Quadrant",
-  "VendorId": 10571,
-  "ProductId": 6401,
+  "VendorId": "0x294B",
+  "ProductId": "0x1901",
   "Inputs": [
     {
       "Id": 17,
-      "Type":  "Button",
+      "Type": "Button",
       "Label": "Mode - IAS"
     },
     {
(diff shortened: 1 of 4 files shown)


tigert:
image

Seems to work. PR inbound.

Issue #1626, opened 2024-01-27
On Release not triggering with Octavi IFR1


tfinnie:
**Describe the bug**
The On Release action configured for an Octavi IFR1 button is not triggered on release of the button

**Steps To Reproduce**
1. Create a new Input Configuration
2. Under "Choose input" in the Input Config Wizard choose a button e.g. Button_FMS1_CLR
3. Under "Input settings" configure actions for both "On Press" and "On Release"
4. Enable Info-level logging under Settings
5. Press and release the above button. 

**Expected behavior**
PRESS event should trigger and be shown in the Logging pane as the button is pressed, followed by the RELEASE event when the button is released

**Actual behavior**
Only the PRESS event is shown in the Logging pane

**Details**
 - MobiFlight Version: 10.1.1

**Additional context**
Refer to [this discussion on Discord](https://discord.com/channels/608690978081210392/1199945180489457714)

PR #1627, opened 2024-01-27
Add handling of button release events for Octavi devices

tfinnie:
Fixes Issue #1626


MobiFlight/Joysticks/Octavi/Octavi.cs
```diff
@@ -121,9 +121,9 @@ private void InputReceiver_Received(object sender, System.EventArgs e)
             {
                 OctaviReport OReport = new OctaviReport();
                 OReport.parseReport(inputReportBuffer);
-                List<int> btnz = octaviHandler.toButton(OReport);
-                foreach (int i in btnz) {
-                    TriggerButtonPress(i);
+                List<(int, MobiFlightButton.InputEvent)> btnz = octaviHandler.toButton(OReport);
```
neilenns, line 124:
```suggestion
                var buttons = octaviHandler.toButton(OReport);
```

We've been slowly moving towards `var` when possible. And I realize it wasn't your original code, but we can spell the whole word `buttons` out, there's no need to abbreviate it.

MobiFlight/Joysticks/Octavi/Octavi.cs
```diff
@@ -121,9 +121,9 @@ private void InputReceiver_Received(object sender, System.EventArgs e)
...
+                foreach (var btn in btnz) {
+                    TriggerButtonPress(btn.Item1, btn.Item2);
```
neilenns, line 126:
```suggestion
                foreach (var button in buttons) {
                    TriggerButtonPress(button.Item1, button.Item2);
```

neilenns (changes requested):

tfinnie:
While working on this code I've noticed some other behaviour which doesn't seem correct to me:

1. The AP context is toggling shift mode when pressing the encoder button and I don't think it should be. This is giving rise to the following button events: **Button_AP^_TOG**, **Button_AP^_OI**, **Button_AP^_OD**, **Button_AP^_II**, **Button_AP^_ID**. Can these be removed and the shift mode removed also? It is also causing the **Button_AP_CRSR** event to only fire every second time the encoder button is pressed.
2. COM2, NAV1, NAV2, and XPDR are raising **Button_COM2_CRSR**, **Button_NAV1_CRSR**, **Button_NAV2_CRSR** and **Button_XPDR_CRSR** events respectively (COM1 is not doing this). I don't think these should exist since pushing the encoder button is for toggling shift mode in these contexts.

MobiFlight/Joysticks/Octavi/OctaviHandler.cs
```diff
@@ -393,47 +320,73 @@ public OctaviHandler()
...
+            foreach (OctaviReport.OctaviButtons button in Enum.GetValues(typeof(OctaviReport.OctaviButtons)))
             {
-                if ((pressed & HIDEvent) != 0)
+                if (pressed.HasFlag(button) || released.HasFlag(button))
                 {
-                    int ButtonPress = extContentState * 16 + (int)HIDEventAssignments[HIDEvent]; // find "full matrix" event index
-                    ButtonPress = OctaviButtonMatrix.ElementAt(ButtonPress).Value; // translate to existing Octavi "devices" in MF
-                    if (ButtonPress >= 0) ButtonPresses.Add(ButtonPress); // if not "unassigned" (-1), then press the button!
+                    int lookupIndex = extendedContextState * 16 + HidEventAssignments[button]; // find "full matrix" event index
+                    int buttonIndex = OctaviButtonMatrix.ElementAt(lookupIndex).Value; // translate to existing Octavi "devices" in MF
```
tfinnie, line 380: OctaviButtonMatrix is a Dictionary which has no guaranteed order. Using ElementAt is not really the correct way to do this.

tfinnie:
I would also like to suggest a renaming of the AP buttons like so, to better represent their actual functions and the Octavi hardware:

- **Button_AP_CDI** -> **Button_AP_AP**
- **Button_AP_OBS** -> **Button_AP_HDG**
- **Button_AP_MSG** -> **Button_AP_NAV**
- **Button_AP_FPL** -> **Button_AP_APR**
- **Button_AP_VNAV** -> **Button_AP_ALT**
- **Button_AP_PROC** -> **Button_AP_VS**

This would obviously break exisiting configuration though. Not sure how big of a deal that is?

tfinnie:
I've created Issue #1642 for this. I think I should hold off on creating a PR until this one is finalised otherwise it will create merge issues.

PR #1670, opened 2024-02-11
Provide different mfmc files for same custom board

### LINKED ISSUE #1669: Provide different mfmc files for same custom board.
For a plug n play experience where you have the same hardware device multiple times in the cockpit, like EFIS, Radios, etc. - one for CPT and one for FO, it would be nice to let the user select which type they want.

At the moment we have only one option for the bundled mfmc file for a custom board, so we want to have an option to select from. CPT vs FO.

Alternatives:
* Use the exact same mfmc, then let the user rename the board.
* Use a profile for CPT and FO with the respective board names, but device names are the same.
* Merge the configs.

**Additional context**
This is a request by @coreflighttech

DocMoebiuz:
fixes #1669 

1668-multiple-mfmc-files

### DIFF
--- a/MobiFlight/Board.cs
+++ b/MobiFlight/Board.cs
@@ -2,6 +2,8 @@
 using System;
 using System.Collections.Generic;
 using System.Drawing;
+using System.IO;
+using System.Linq;
 
 namespace MobiFlight
 {
@@ -424,19 +426,39 @@ namespace MobiFlight
         /// Get the name for the Default Config
         /// </summary>
         /// <returns>The </returns>
-        public string GetDefaultDeviceConfigFilePath()
+        protected string GetDefaultDeviceConfigFilePath()
         {
-            return $@"{BasePath}\config\{Info.FirmwareBaseName}.mfmc";
+            return Path.Combine(BasePath, "config", $"{Info.FirmwareBaseName}.mfmc") ;
         }
 
-        /// <summary>
-        /// Get the full path for the specific device config file
-        /// This is needed when there are multiple device config files for a single board
-        /// </summary>
-        /// <returns>The full path to the device config</returns>
-        public string GetDeviceConfigFilePath(string deviceConfigFile)
+        public IEnumerable<DeviceConfigFile> GetExistingDeviceConfigFiles()
         {
-            return $@"{BasePath}\config\{deviceConfigFile}";
+            // this is the fallback that we used before we had the DeviceConfigs property
+            if (Info.DeviceConfigs == null)
+            {
+                var DefaultDeviceConfigFile = GetDefaultDeviceConfigFilePath();
+                if (!File.Exists(DefaultDeviceConfigFile))
+                    return new List<DeviceConfigFile>();
+                
+                return new List<DeviceConfigFile>
+                {
+                    new DeviceConfigFile
+                    {
+                        Name = "Default",
+                        Description = "Default device configuration.",
+                        File = DefaultDeviceConfigFile
+                    }
+                };
+            }
+                
+            return Info.DeviceConfigs
+                    .Select(file => new DeviceConfigFile
+                    {
+                        Name = file.Name,
+                        Description = file.Description,
+                        File = Path.Combine(BasePath, "config", file.File)
+                    })
+                    .Where(file => File.Exists(file.File));
         }
     }
 }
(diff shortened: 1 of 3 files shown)


MobiFlight/Board.cs
```diff
@@ -417,5 +428,15 @@ public string GetDefaultDeviceConfigFilePath()
...
         }
+
+        /// <summary>
+        /// Get the full path for the specific device config file
+        /// This is needed when there are multiple device config files for a single board
+        /// </summary>
+        /// <returns>The full path to the device config</returns>
+        public string GetDeviceConfigFilePath(string deviceConfigFile)
+        {
+            return $@"{BasePath}\config\{deviceConfigFile}";
```
neilenns, line 439: `Path.Combine()`!!! 😆

Boards/mfboard.schema.json
```diff
@@ -264,6 +264,39 @@
...
             }
           }
+        },
+        "DeviceConfigs": {
+          "$id": "#root/Info/DeviceConfigs",
+          "title": "DeviceConfigs",
+          "description": "A list of default device configs for this board.",
+          "type": "array",
+          "default": [],
+          "items": {
```
neilenns, line 274: Name, Description, and File are all required, yes? They should be listed as required in the schema.

MobiFlight/Board.cs
```diff
@@ -417,5 +428,15 @@ public string GetDefaultDeviceConfigFilePath()
         {
             return $@"{BasePath}\config\{Info.FirmwareBaseName}.mfmc";
```
neilenns, line 429: Might as well fix this too while you're at it :)

UI/Panels/Device/MFModulePanel.cs
```diff
@@ -93,20 +88,53 @@ private void DisplayDetails(Board board)
...
                 buttonSupport.Enabled = false;
 
 
-            
+
             var defaultDeviceConfigFile = board.GetDefaultDeviceConfigFilePath();
-            if (defaultDeviceConfigFile != null && File.Exists(defaultDeviceConfigFile))
+            var defaultDeviceConfigFileExists = defaultDeviceConfigFile != null && File.Exists(defaultDeviceConfigFile);
+            var specificDeviceConfigs = board.Info.DeviceConfigs?.Where(p => File.Exists(board.GetDeviceConfigFilePath(p.File)));
+            var atLeastOneSpecificConfigExists = specificDeviceConfigs.Count() > 0;
```
neilenns, line 102: Feels like this logic should be in the class itself not here. This code should just be able to call `board.Info.DeviceConfigs` and only get back ones that actually exist instead of having to worry about it. Same for calculating an `atLeastOneSpecificConfigExists` property, make it a getter-based property on the class.

UI/Panels/Device/MFModulePanel.cs
```diff
@@ -93,20 +88,53 @@ private void DisplayDetails(Board board)
...
+                        var profile = specificDeviceConfigs.First();
+                        var filePath = board.GetDeviceConfigFilePath(profile.File);
+                        UploadDefaultConfigRequested?.Invoke(this, filePath);
+                    }   
+
+                    // since we have more options, we present a context menu
+                    else
+                    {
+                        var menu = new ContextMenuStrip();
+                        foreach (var profile in board.Info.DeviceConfigs)
```
neilenns, line 125: Why is this using `board.Info.DeviceConfigs` instead of `specificDeviceConfigs` which already filtered out configs that don't have paths?

frontend/.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
```
neilenns, line 5: This file doesn't belong in this PR

frontend/.vscode/tasks.json
```diff
@@ -0,0 +1,12 @@
+{
```
neilenns, line 1: This file doesn't belong in this PR either.

neilenns (changes requested):
Buncha feedback before Tay Tay.

PR #1680, opened 2024-02-12
Auto-run works on plane change.

### LINKED ISSUE #1488: Honour AutoRun setting when auto loading a config based on aircraft type

The new auto-config loading for aircraft is sweet, unfortunately it doesn't support AutoRun. To reproduce:

1. Map a config file to an airplane (C172 for example), make sure to set it up for auto-load config
2. Load a different config file (CJ4 for example)
3. Enable AutoRun
4. Close Mobiflight
5. Spawn with the C172
6. Run Mobiflight

I'd love to have the config both load *and* start MobiFlight whenever I change planes in MSFS2020. That way I can just leave MobiFlight minimized to the system tray, open whatever plane I feel like flying, and fly.

Right clicking on the icon in the system tray and hitting Run, which is lame, because the whole point to auto-loading configs is so you don't have to interact with MobiFlight.

**Additional context**

This should probably only happen for auto-load, not when a user opens a config file from the file menu.

DocMoebiuz:
fixes #1488

Auto-run is postponed to after switching an aircraft, and it is executed on EVERY detected aircraft change.

Note: there is still a problem with the auto-retrigger hapenning too early. it should only happen when the flight is loaded.

### DIFF
--- a/MobiFlight/ExecutionManager.cs
+++ b/MobiFlight/ExecutionManager.cs
@@ -290,12 +290,17 @@ namespace MobiFlight
 
             simConnectCache.Start();
             xplaneCache.Start();
+
+            // the timer has to be enabled before the 
+            // on start actions are executed
+            // otherwise the input events will not be executed.
+            timer.Enabled = true;
+            
+            // Now we can execute the on start actions
             OnStartActions();
 
             // Force all the modules awake whenver run is activated
             mobiFlightCache.KeepConnectedModulesAwake(true);
-
-            timer.Enabled = true;
         }
 
         public void Stop()


MobiFlight/ExecutionManager.cs
```diff
@@ -1362,7 +1362,7 @@ void mobiFlightCache_OnButtonPressed(object sender, InputEventArgs e)
             {
                 eventAction = MobiFlightEncoder.InputEventIdToString(e.Value);
             }
-            if (e.Type == DeviceType.InputShiftRegister)
+            else if (e.Type == DeviceType.InputShiftRegister)
```
neilenns, line 1365: Unrelated change?

neilenns (approved):

PR #1688, opened 2024-02-19
Octavi shift modes controlled by .json file

felixc7:
This PR containts changes to enable users to define which Octavi IFR-1 context modes have a shift mode and which don't. This is done by defining inputs in the octavi.joystick.json. Those context modes that are listed in the json will have a shift mode, the others won't. In future versions, this definition should be moved to the profile .mcc.

### DIFF
--- a/MobiFlight/Joysticks/Octavi/OctaviHandler.cs
+++ b/MobiFlight/Joysticks/Octavi/OctaviHandler.cs
@@ -1,5 +1,8 @@
-﻿using System;
+﻿using Antlr.Runtime.Tree;
+using System;
 using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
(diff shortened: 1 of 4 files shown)


MobiFlight/Joysticks/Octavi/OctaviHandler.cs
```diff
@@ -17,14 +17,14 @@ internal class OctaviHandler
         private enum OctaviEncoder { OUTER_INC, OUTER_DEC, INNER_INC, INNER_DEC }
 
         public IEnumerable<string> JoystickButtonNames { get; private set; }
-        public List<context> contexts;
-        public struct context
+        public List<Context> contexts;
```
felixc7, line 20: Changed as per @DocMoebiuz ' suggestion

PR #1713, opened 2024-02-23
Prevent Octavi Exception on Connect

### LINKED ISSUE #1697: Octavi Connect is called twice and inputReceiver throws exception
**Describe the bug**
In the Redesign branch I noticed that there is an Exception for the Octavi because Connect is called and Update is also called, which subsequently calls Connect

**Steps To Reproduce**
1. Connect Octavi
2. Start Mobiflight

**Expected behavior**

**Actual behavior**
A clear and concise description of what you actually happens.

**Details**
 - MobiFlight Version: [Redesign branch]

**Additional context**
image

DocMoebiuz:
fixes #1697 

**Notes**
* The connect logic differs from the other built-in joysticks. Connect is called when Update() or SendData() is called which happens periodically. Especially in the rebrush2024 it happens a lot that Update() and SendData() are called concurrently from different threads which throws an Exception in the Connect method.
* With the lock and the additional checks that should not be a problem anymore.
* Once we remove the DirectInput we can use a common virtual Connect method for all classes

### DIFF
--- a/MobiFlight/Joysticks/Octavi/Octavi.cs
+++ b/MobiFlight/Joysticks/Octavi/Octavi.cs
@@ -20,20 +20,35 @@ namespace MobiFlight.Joysticks.Octavi
 
         }
 
+        /// <summary>
+        /// Method is not called by regular Joystick Manager
+        /// This method is called implicitly when Update() is called
+        /// </summary>
         public void Connect()
         {
-            if (Device == null)
+            // Prevent reentry and parallel execution by multiple threads
+            lock (this)
             {
-                Device = DeviceList.Local.GetHidDeviceOrNull(vendorID: VendorId, productID: ProductId);
-                if (Device == null) return;
-            }
+                if (Device == null)
+                {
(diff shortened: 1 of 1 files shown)


MobiFlight/Joysticks/Octavi/Octavi.cs
```diff
@@ -20,20 +20,35 @@ public Octavi(SharpDX.DirectInput.Joystick joystick, JoystickDefinition definiti
...
 
+        /// <summary>
+        /// Method is not called by regular Joystick Manager
+        /// This method is called implicitly when Update() is called
+        /// </summary>
         public void Connect()
         {
-            if (Device == null)
+            // Prevent reentry and parallel execution by multiple threads
+            lock (this)
```
Koseng, line 30: Usually lock(this) is not recommended. Best practice is to use a separate dedicated new private object for a lock. Because it is out of control if someone else might use that object reference also for some locking.

PR #1718, opened 2024-02-25
Fix issue with loading of OnHold and OnLongRelease config in OUTPUT

### LINKED ISSUE #1689: Issue with On Hold and On Long Release tabs in OUTPUT / Display / Input Action
Hello 
It seems there is a problem with the On Hold and On Long Release tabs in the
OUTPUT configs/ Display tab / Display type : Input Action.

Any change in these tabs are lost as soon as you click OK.
No issue with the "On Press" and "On Release" tabs on the other hand.

To reproduce the issue :

1. Create a new OUTPUT in Mobiflight. Let's call it TEST in the description zone.
2. Edit this OUTPUT with the button on the right.
3. In the Display tab, choose " Input Action ".
4. Keep the default On Press / On Release radio button.
5. In the On Press Tab, choose for example "Mobiflight - Variable".
6. Let's keep the default name "MyVar" and set value "1".
7. Do the same thing with the "On Release" tab, this time with set value "0".
8. Click OK, you're back on Mobiflight main page.
9. Edit again the OUTPUT named TEST we've just created.
10. In the Display Tab, we see that the changes we've created have been saved correctly.
11. Now, open the On Hold Tab or the On Release Tab, as the result will be just the same.
12. Choose any action type, let's try again for example  "Mobiflight - Variable".
13. Let's keep the default name "MyVar" and this time set value "2".
14. Click OK to return to the main Mobiflight page.
15. Edit again the OUTPUT named TEST you will be able to notice that any change in the On Hold Tab or the On Release Tab
we have made has been lost.

You also get the same error with other types of Input actions in these "On Hold" and "On Long Release" tabs.

**Details**
 - MobiFlight Version : 10.1.4
 - FlightSim : MSFS2020

Koseng:
fixes #1689
Since `syncFromConfig` handles also null input and the whole config is cloned on opening the wizard, the previous code was unecessary and missed the new button config parameters.

I don't understand why new AnalogInputConfig and ButtonInputConfig objects were created. First I wanted to clone the config instead of assigning single parameters, but then I thought why cloning at all?

### DIFF
--- a/UI/Panels/OutputWizard/DisplayPanel.cs
+++ b/UI/Panels/OutputWizard/DisplayPanel.cs
@@ -167,14 +167,8 @@ namespace MobiFlight.UI.Panels.OutputWizard
             }
             else
             {
-                var analogInputConfig = new AnalogInputConfig();
-                analogInputConfig.onChange = config.AnalogInputConfig?.onChange;
-                analogPanel1.syncFromConfig(analogInputConfig);
-
-                var buttonInputConfig = new ButtonInputConfig();
-                buttonInputConfig.onPress = config.ButtonInputConfig?.onPress;
-                buttonInputConfig.onRelease = config.ButtonInputConfig?.onRelease;
-                buttonPanel1.syncFromConfig(buttonInputConfig);
+                analogPanel1.syncFromConfig(config.AnalogInputConfig);
+                buttonPanel1.syncFromConfig(config.ButtonInputConfig);
 
                 InputTypeButtonRadioButton.Checked = config.AnalogInputConfig?.onChange == null;
                 InputTypeAnalogRadioButton.Checked = config.AnalogInputConfig?.onChange != null;


Koseng:
> > fixes #1689 Since `syncFromConfig` handles also null input and the whole config is cloned on opening the wizard, the previous code was unecessary and missed the new button config parameters.
> > I don't understand why new AnalogInputConfig and ButtonInputConfig objects were created. First I wanted to clone the config instead of assigning single parameters, but then I thought why cloning at all?
> 
> we clone the configs ALWAYS inside the config wizards to make sure that any change made inside the WIzard does not affect the original item. So when you click cancel then we can assure that the original item was not changed. Also it helps us to determine whether anything was changed and if we need to mark the config file as "modified" to enable/disable the save-button in the menu bar accordingly.

Yeah, that is clear. What I did not understand was, since the behavior is as described, why there was an **additional** copy of the ButtonConfig in the previous code. So first the clone, then an additional copy, which I removed.

Issue #1720, opened 2024-02-28
APC Key 25 midi controller - I/O configuration in .midiboard.json


ZionFon:
New configuration I/O for APC Key 25 midi controller.
LimitedKnobs, LED buttons..

PR #1722, opened 2024-02-28
APC Key 25 midi controller - I/O configuration in .midiboard.json

ZionFon:


Koseng:
Hi, thanks.
Why no Input Config for all the buttons? 
If the buttons are in the input config as well, they show in the dropdown list.

ZionFon:
Midi buttons have labels configured by default. As 'Note 1_39', 'Note 1_40', etc.
`var inputEventArgs = GetDefaultInputEventArgs(MidiMessageType.Note, channel, id);  `
I set up pots that were treated as buttons (OnPress, OnRelease) and now they have OnChange.

ZionFon:
> If the buttons are in the input config as well, they show in the dropdown list.
aa.. ok... 
I will add.

Koseng:
Related to #1720 
@DocMoebiuz now looks good to me.

PR #1768, opened 2024-04-23
Changed "pin" type to uint to extend max pin number

Ragot-Pierre:
I'm not used to git and pull-requests so if I'm doing something wrong, please tell me :)

MobiFlight is a great piece of software but for my project I am limited by the 255 maximum pin per module.
I forked the repository, looked at the code and changed each "pin" type definition to _uint_ instead of _byte_ to get rid of this limitation.

I understand that is something only few (if any) people encounter but I think this change would not break anything for people who don't need it.
Not a C# pro so I might have missed something but everything seems to work fine on my end with both local and Git builds.


elral:
Hmhm, within the firmware the pins are also defined as byte (uint8\_t), so limited to 255.

How do you get more than 255 pins per board?

Ragot-Pierre:
> Hmhm, within the firmware the pins are also defined as byte (uint8_t), so limited to 255.
> 
> How do you get more than 255 pins per board?

With a custom firmware (and hardware, too). I made the A32X/A33X/A34X MCDU with an RP2040 but it's not a simple button matrix board :
1 core handles USB, the 74 button matrix, the 8 announciators LEDs, buttons / frame backlight, and even the screen controls (not tested yet, but the hardware is here).
The other core orchestrates 2 custom data bus going to other panels that will be "Plug & Play" (without going back to the config) and can be daisy-chained. 
The MCDU will act as a hub, piping all the data through 1 USB cable, even with a full cockpit.

PR #1781, opened 2024-05-25
Crash when attempting to connect to board via serial

### LINKED ISSUE #1778: Crash when attempting to connect to board via serial
[Reported by a user in Discord](https://discord.com/channels/608690978081210392/1243952273395286106). Don't know the cause, but we should handle the exception and not crash.

Exception is System.InvalidOperationException from CommandMessenger.Transport.Serial.SerialTransport.Connect(). This is called in MobiFlightModule.cs:Connect() with no try/catch.

neilenns:
Fixes #1778

Add a try/catch around the connect code. This is a draft PR to get the build for a user in Discord to try.


neilenns:
The specific devices is getting connected with an empty COM port name for some reason.

Issue #1789, opened 2024-06-16
VKB Controller LED and Encoder support


cbrauers:
**Is your feature request related to a problem? Please describe.**
As discussed on Discord, a number of users would like to control the LEDs on their VKB devices with Mobiflight. Since VKB's LED API is more complex than the generic LED API introduced for Honeycomb devices, this requires a special implementation. A VKB-specific implementation also allows interfacing directly with rotary encoders, since DirectInput modes tend to slow down the encoders a lot.

**Describe the solution you'd like**
Introduce handling of VKB LEDs and encoders into MobiFlight.

**Describe alternatives you've considered**
I have written a C++ tool to interface with MSFS and VKB devices before, but with HIDSharp already a part of MobiFlight, it is more effective to implement the communication from scratch with the experiences gained during its development.

**Additional context**
I have been working on [an implementation](https://github.com/cbrauers/MobiFlight-Connector) over the past few weeks, currently doing some final polishing before creating a pull request. You can find [my branch](https://github.com/cbrauers/MobiFlight-Connector) (along with two [companion](https://github.com/cbrauers/VKB-Mobiflight-Definer) [tools](https://github.com/cbrauers/EncoderVisualizer)) on my GitHub profile.
With the modularity and customizability of VKB devices, it is not sufficient to include a handful of pre-made definitions. With just module combinations sold on the store or featured in official guides, we are already in the ballpark of about 40 different definitions, and user customization can increase that arbitrarily. I have therefore written a [tool](https://github.com/cbrauers/VKB-Mobiflight-Definer) to quickly create own definitions. Currently just essentially copying and pasting blocks together without in-depth JSON awareness, but it works for combinations that are not too heavily manipulated. The second [tool](https://github.com/cbrauers/EncoderVisualizer) is a testing tool for the encoder interface.
Protocol documentation from VKB is a bit sparse, but I have provided the most relevant pieces here:
[LEDs Protocol](https://github.com/user-attachments/files/15859047/LEDs_Protocol.pdf)
[Encoder capabilities](http://forum.vkb-sim.pro/viewtopic.php?f=28&t=7843#p61557)

Koseng:
Implemented with #1792

Wibbbs:
Sorry to revive a closed issue, but I would LOVE to be able to use the LEDs on my VKB joystick as outputs from mobiflight. Is there any documentation on how to set this up?

cbrauers:
[...]

Issue #1786, opened 2024-05-31
WASM module may crash in conjunction with joystick encoders and an "on Hold" binding


cbrauers:
**Describe the bug**
When HID devices (e.g. Winwing FCU) provide encoders, those are rendered as buttons. Those buttons have "on release", "on hold" and "on long release" bindings. Since those encoders do not send release events, "on hold" actions may cause unexpected behavior, including the MSFS2020 WSM binding crashing

**Steps To Reproduce**
1. Connect a HID device with encoders, e.g. a WinWing FCU
2. Assign an action to "on hold" on one of the encoders
3. Assign the same action to another input
4. In the sim, turn the encoder
5. Use the other input to execute that same action

**Expected behavior**
It is either not possible to assign "on hold" actions to encoders, or such actions are ignored

**Actual behavior**
The on hold action is apparently sent and never released, causing the WASM plugin to crash if a press from another source is received

**Details**
 - MobiFlight Version: 10.2.2.2
 - FlightSim: MSFS 2020
 - AddOn: Any plane, peripheral that acts as HID joystick with encoders supported by MobiFlight

**Additional context**
Found by accident (incomplete update of older MCC file) during development of VKB plugin. Later confirmed on WinWing FCU.

DocMoebiuz:
@cbrauers the behavior you are describing about the encoders, this is new to me. The encoders that I have seen on Joysticks have a OnPress -> OnRelease sequence. But they would not stay in the OnPress when at detent.

Maybe i am not fully understanding it.

We also implemented https://github.com/MobiFlight/MobiFlight-Connector/issues/2890 which might have solved this problem.

PR #1792, opened 2024-06-18
Add VKBDevice type of HID controller

cbrauers:
Solves #1789

- Implements VKB LED protocol (firmware 2.08.5+), allowing for control of single-color, bi-color and RGB LEDs. In accordance with current limitation on HID LEDs, only on and off are supported for each channel. The byte identifier is used for the LED ID, the bit identifier repurposed for the color channel.
- Implements VKB encoder protocol (firmware 2.17.9+, still experimental on VKB's end), solving slowness arising from the use of directinput buttons as the default form of output. Encoders suffer from the same limitations as Octavi and Winwing FCU encoders, specifically being affected by #1786.
- Changes identification of VKB device names (VID 0x231D):
  - Extraneous spaces (VKB quirk) are stripped
  - HID API is used to query names instead of DirectInput - more robust with configuration changes; DirectInput may cache outdated names 
- Includes a large array of definition files for common VKB controllers:
  - Stock configs of Gladiator NXT (including EVO), STECS and GNX module combos
  - GNX module configs featured in VKB instructional videos
  - Standalone GNX modules
  - a small selection of custom combos
- Tools for quickly making own definition files (VKB devices are very modular, so even that only covers the common use cases) are made available separately and are currently not planned to be distributed alongside MobiFlight.

### DIFF
--- a/MobiFlight/Joysticks/VKB/VKBDevice.cs
+++ b/MobiFlight/Joysticks/VKB/VKBDevice.cs
@@ -1,0 +1,192 @@
+﻿using HidSharp;
+using HidSharp.Reports;
+using HidSharp.Reports.Input;
+using System;
+using System.Collections.Generic;
+
+namespace MobiFlight.Joysticks.VKB
+{
+    internal class VKBDevice : Joystick
+    {
+        private HidStream Stream;
+        private readonly HidDevice Device;
+        private readonly JoystickDefinition Definition;
+        private readonly VKBLedContainer Lights = new VKBLedContainer();
+        private HidDeviceInputReceiver InputReceiver;
+        private readonly byte[] InputReportBuffer = new byte[64];
+        private readonly SortedList<byte, VKBEncoder> Encoders = new SortedList<byte, VKBEncoder>();
+        int lastSeqNo = -1;
+
+        public VKBDevice(SharpDX.DirectInput.Joystick joystick, JoystickDefinition definition) : base(joystick, definition)
+        {
+            Definition = definition;
+            if (Device == null)
+            {
+                Device = GetMatchingHidDevice(joystick);
+                if (Device == null) return;
+            }
+        }
+        public override void Connect(IntPtr handle)
+        {
+            base.Connect(handle);
+            if (Stream == null)
+            {
+                Stream = Device.Open();
+                Stream.ReadTimeout = System.Threading.Timeout.Infinite;
+            }
+
+            if (InputReceiver == null)
+            {
+                InputReceiver = new ReportDescriptor(VKBHidReport.Descriptor).CreateHidDeviceInputReceiver();
+                // We use our own descriptor since the one we get from Windows/HIDSharp has been altered and encoder data is missing
+                InputReceiver.Received += OnHidReportReceived;
+                InputReceiver.Start(Stream);
+            }
+        }
+        protected override void SendData(byte[] data)
+        {
+            // VKBLedContainer has its own handling that replaces RequiresOutputUpdate, so we just send.
+            // Don't try and send data if no outputs are defined.
+            if (Definition?.Outputs == null || Definition?.Outputs.Count == 0)
+            {
+                return;
+            }
+            Stream?.SetFeature(data);
+
+        }
+        protected override void EnumerateDevices()
+        {
+            base.EnumerateDevices();
+            SortedList<byte, JoystickDevice> EncoderDecList = new SortedList<byte, JoystickDevice>();
+            SortedList<byte, JoystickDevice> EncoderIncList = new SortedList<byte, JoystickDevice>();
+            if (Definition == null || Definition.Inputs == null) return;
+            foreach (var input in Definition.Inputs)
+            {
+                // The 1xxx range is limited to encoders. They are also not fed from DirectInput.
+                if (input.Id >= 1000 && input.Id < 2000 && input.Type == JoystickDeviceType.Button)
+                {
+                    byte encoderIndex = GetEncoderIndex(input);
+                    VKBEncoder.EncoderAction encoderAction = GetEncoderAction(input);
+                    if (encoderAction == VKBEncoder.EncoderAction.DEC)
+                    {
+                        EncoderDecList.Add(encoderIndex, new JoystickDevice { Name = $"Button {1000 + 10 * encoderIndex + (int)encoderAction}", Label = input.Label, Type = DeviceType.Button, JoystickDeviceType = input.Type });
+                    }
+                    if (encoderAction == VKBEncoder.EncoderAction.INC)
+                    {
+                        EncoderIncList.Add(encoderIndex, new JoystickDevice { Name = $"Button {1000 + 10 * encoderIndex + (int)encoderAction}", Label = input.Label, Type = DeviceType.Button, JoystickDeviceType = input.Type });
+                    }
+                    Buttons.FindAll(but => but.Label == input.Label).ForEach(but => but.Label += " (Legacy DirectInput)"); 
+                    // Rename the original buttons to keep indexing intact and ensure compatibility with devices not configured to use encoder channels.
+                }
+            }
+            foreach (var encdec in EncoderDecList)
+            {
+                if (EncoderIncList.ContainsKey(encdec.Key))
+                {
+                    Encoders.Add(encdec.Key, new VKBEncoder(EncoderIncList[encdec.Key], encdec.Value)); // If both increment and decrement were in the definition, we create an encoder object from the definition file
+                    Buttons.Add(Encoders[encdec.Key].DeviceDec); // And register their virtual buttons in the joystick's button list.
+                    Buttons.Add(Encoders[encdec.Key].DeviceInc);
+                }
+            }
+        }
+
+        private static byte GetEncoderIndex(JoystickInput input)
+        {
+            return (byte)((input.Id - 1000) / 10);
+        }
+        private static VKBEncoder.EncoderAction GetEncoderAction(JoystickInput input)
+        {
+            return (VKBEncoder.EncoderAction)(input.Id % 10);
+        }
+        protected override void EnumerateOutputDevices()
+        {
+            Definition?.Outputs?.ForEach(output => Lights.AddChannel(output));
+            base.EnumerateOutputDevices();
+        }
+
+        public override void SetOutputDeviceState(string name, byte state)
+        {
+            Lights.UpdateState(name, state);
+        }
+
+
+        public override void UpdateOutputDeviceStates()
+        {
+            var data = Lights.CreateMessage();
+            if (data[7] == 0) return; // Only send message if there are non-zero LEDs to be updated
+            try
+            {
+                SendData(data);
+            }
+            catch (System.IO.IOException)
+            {
+                base.OnDeviceRemoved();
+            }
+        }
+        private void OnHidReportReceived(object sender, System.EventArgs e)
+        {
+            var inputReceiver = sender as HidDeviceInputReceiver;
+            while (inputReceiver.TryRead(InputReportBuffer, 0, out _))
+            {
+                byte ReportId = InputReportBuffer[0];
+                if (ReportId != 0x08) // 0x08 = Monitoring channel / virtual bus
+                    continue;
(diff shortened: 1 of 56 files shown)


MobiFlight/Joysticks/VKB/VKBDevice.cs
```diff
@@ -0,0 +1,192 @@
...
+        private readonly SortedList<byte, VKBEncoder> Encoders = new SortedList<byte, VKBEncoder>();
+        int lastSeqNo = -1;
+
+        public VKBDevice(SharpDX.DirectInput.Joystick joystick, JoystickDefinition definition) : base(joystick, definition)
+        {
+            Definition = definition;
+            if (Device == null)
+            {
+                Device = GetMatchingHidDevice(joystick);
+                if (Device == null) return;
```
neilenns, line 26: Is this necessary? There's no code that follows so this really doesn't do anything.

If it does stick around, please put the return in { } to stay consistent with how we handle if statements elsewhere.

MobiFlight/Joysticks/VKB/VKBDevice.cs
```diff
@@ -0,0 +1,192 @@
...
+
+            if (InputReceiver == null)
+            {
+                InputReceiver = new ReportDescriptor(VKBHidReport.Descriptor).CreateHidDeviceInputReceiver();
+                // We use our own descriptor since the one we get from Windows/HIDSharp has been altered and encoder data is missing
+                InputReceiver.Received += OnHidReportReceived;
+                InputReceiver.Start(Stream);
+            }
+        }
+        protected override void SendData(byte[] data)
```
neilenns, line 46: Minor nit but there should be blank lines between each function definition to stay consistent with our style in other files.

MobiFlight/Joysticks/VKB/VKBDevice.cs
```diff
@@ -0,0 +1,192 @@
...
+                return;
+            }
+            Stream?.SetFeature(data);
+
+        }
+        protected override void EnumerateDevices()
+        {
+            base.EnumerateDevices();
+            SortedList<byte, JoystickDevice> EncoderDecList = new SortedList<byte, JoystickDevice>();
+            SortedList<byte, JoystickDevice> EncoderIncList = new SortedList<byte, JoystickDevice>();
```
neilenns, line 61: This should be `var` EncoderDecList/EncoderIncList to stay consistent with our (ongoing!) style update to use var where possible. Also applies to all other variable definitions in this PR, where possible use `var`.

MobiFlight/Joysticks/VKB/VKBDevice.cs
```diff
@@ -0,0 +1,192 @@
...
+            foreach (var input in Definition.Inputs)
+            {
+                // The 1xxx range is limited to encoders. They are also not fed from DirectInput.
+                if (input.Id >= 1000 && input.Id < 2000 && input.Type == JoystickDeviceType.Button)
+                {
+                    byte encoderIndex = GetEncoderIndex(input);
+                    VKBEncoder.EncoderAction encoderAction = GetEncoderAction(input);
+                    if (encoderAction == VKBEncoder.EncoderAction.DEC)
+                    {
+                        EncoderDecList.Add(encoderIndex, new JoystickDevice { Name = $"Button {1000 + 10 * encoderIndex + (int)encoderAction}", Label = input.Label, Type = DeviceType.Button, JoystickDeviceType = input.Type });
```
neilenns, line 72: Can you add a comment that shows what the formatted display name looks like? It's a little tough to tell from the string interpolation.

MobiFlight/Joysticks/VKB/VKBDevice.cs
```diff
@@ -0,0 +1,192 @@
...
+                    }
+                    if (encoderAction == VKBEncoder.EncoderAction.INC)
+                    {
+                        EncoderIncList.Add(encoderIndex, new JoystickDevice { Name = $"Button {1000 + 10 * encoderIndex + (int)encoderAction}", Label = input.Label, Type = DeviceType.Button, JoystickDeviceType = input.Type });
+                    }
+                    Buttons.FindAll(but => but.Label == input.Label).ForEach(but => but.Label += " (Legacy DirectInput)");
```
neilenns, line 78: Will users know what "(Legacy DirectInput)" means? I assume that will eventually get shown in the UI. Is it necessary?

MobiFlight/Joysticks/VKB/VKBDevice.cs
```diff
@@ -0,0 +1,192 @@
...
+                    }
+                    Buttons.FindAll(but => but.Label == input.Label).ForEach(but => but.Label += " (Legacy DirectInput)"); 
+                    // Rename the original buttons to keep indexing intact and ensure compatibility with devices not configured to use encoder channels.
+                }
+            }
+            foreach (var encdec in EncoderDecList)
+            {
+                if (EncoderIncList.ContainsKey(encdec.Key))
+                {
+                    Encoders.Add(encdec.Key, new VKBEncoder(EncoderIncList[encdec.Key], encdec.Value)); // If both increment and decrement were in the definition, we create an encoder object from the definition file
```
neilenns, line 86: Are there situations where enc and dec are *not* in the definition? What happens in that case, the encoder is just ignored entirely? Might be good to add a comment explaining.

MobiFlight/Joysticks/VKB/VKBDevice.cs
```diff
@@ -0,0 +1,192 @@
...
+            Definition?.Outputs?.ForEach(output => Lights.AddChannel(output));
+            base.EnumerateOutputDevices();
+        }
+
+        public override void SetOutputDeviceState(string name, byte state)
+        {
+            Lights.UpdateState(name, state);
+        }
+
+
```
neilenns, line 112: Nitpicky whitespace consistency ;)

MobiFlight/Joysticks/VKB/VKBDevice.cs
```diff
@@ -0,0 +1,192 @@
...
+            }
+        }
+        private void ParseEncoderReport(byte[] Report)
+        {
+            byte sequenceNo = Report[2];
+            if (((lastSeqNo + 1) & 0xFF) != sequenceNo)
+            {
+                Log.Instance.log("Some VKB encoder messages may have been missed", LogSeverity.Debug);
+                // Can easily happen if many updates are sent in quick succession (like when an encoder is spun fast)
+                // Not a problem since API reports absolute state, so the updates are just received on the next received message.
```
neilenns, line 147: This comment should go above the if statement and explain that the if statement is checking to see if the sequency number increased by more than one from the last received message.

MobiFlight/Joysticks/VKB/VKBDevice.cs
```diff
@@ -0,0 +1,192 @@
...
+            }
+            lastSeqNo = sequenceNo;
+            byte encoderCount = Report[3];
+            int maxEncoders = (Report.Length - 4) / 2;
+            if (encoderCount > maxEncoders)
+            {
+                Log.Instance.log($"Log message reports {encoderCount} encoders, but only has space for {maxEncoders}. Some encoders were ignored.", LogSeverity.Warn);
+                encoderCount = (byte)maxEncoders;
+                // Should not occur in most real-life scenarios, but it is theoretically possible to construct a device with more encoders than the report can handle.
```
neilenns, line 156: Same deal, the comment should go before the if statement to explain why the if is taking place.

MobiFlight/Joysticks/VKB/VKBEncoder.cs
```diff
@@ -0,0 +1,79 @@
...
+            INC = 5
+        }
+        private bool firstStart = false; // Ensures initial values get pulled from the device when encoder object is created from definition file.
+        public JoystickDevice DeviceInc = null; // Virtual button for increment events.
+        public JoystickDevice DeviceDec = null; // Virtual button for decrement events.
+        private ushort value = 0; // Stores current state of encoder
+        public VKBEncoder(byte index, ushort? initialValue = null)
+        {
+            // Constructor if virtual button devices are not created otherwise.
+            if (initialValue == null) firstStart = true;
```
neilenns, line 20: Please put the `firstStart = true` in `{}`. This applies for all if/else/etc statements in all files touched by the PR. Thanks!

MobiFlight/Joysticks/VKB/VKBLedContainer.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace MobiFlight.Joysticks.VKB
+{
+    // This holds all LED channels of a single device along with a label list for speedy retrieval.
+    internal class VKBLedContainer
+    {
+        private readonly SortedList<byte, VKBLed> Leds = new SortedList<byte, VKBLed>();
+        private readonly Dictionary<String, (byte, byte)> Labels = new Dictionary<string, (byte, byte)>();
```
neilenns, line 10: `var`

MobiFlight/Joysticks/VKB/VKBLedContainer.cs
```diff
@@ -0,0 +1,60 @@
...
+        public void AddChannel(JoystickOutput output)
+        {
+            if (!Leds.ContainsKey(output.Byte))
+                Leds.Add(output.Byte, new VKBLed(output.Byte));
```
neilenns, line 14: `{}`

MobiFlight/Joysticks/VKB/VKBLedContainer.cs
```diff
@@ -0,0 +1,60 @@
...
+            Leds[output.Byte].AddChannel(output);
+            Labels.Add(output.Label, (output.Byte, output.Bit));
+        }
+        public void UpdateState (string Label, byte State)
+        {
+            // Find the LED and its color channel and update it.
+            String[] updateLabels = Label.Split('|'); // Multiple output pin support
```
neilenns, line 21: `var`

neilenns (changes requested):
Added a few comments here and there, mostly style related. I'm really not sure why Visual Studio isn't flagging them in the error list as messages since many of them (e.g. using `var`) should be flagged automatically.

To my untrained joystick eye the changes look nice and the way you made things like LEDs work without having to make dramatic changes to existing Joystick code is cool.

I didn't comment on this everywhere but you can make more use of whitespace between sections of code. Even within functions things are all smooshed together. The rest of the codebase makes good use of whitespace between logical chunks of code.

MobiFlight/Joysticks/VKB/VKBDevice.cs
cbrauers: Apparently any code that used it got refactored out before I made my first commit in my fork. Will clean the not-reakky-early return up.

MobiFlight/Joysticks/VKB/VKBDevice.cs
cbrauers: I have it in the other place encoders can be created (when they are found on the device but not in the definition), but I can reproduce it here.

MobiFlight/Joysticks/VKB/VKBDevice.cs
cbrauers: Not intentionally, but I imagine it could happen if a user needs to modify a definition for a customized stick and they make a mistake in the process. The encoder assumes both buttons are present. Will add comment and may look into additional safeguards in such situations.

MobiFlight/Joysticks/VKB/VKBDevice.cs
cbrauers: It is necessary to put *something* there. Removing it entirely would require a new copy of the button handling code that can handle gaps in the button list, plus it would be bad for users who have reconfigured their device to reuse those button slots for different purposes, or for those whose devices are not set up to use the encoder API (old firmware or wrong firmware settings).
Basically, the purpose is to have something that tells users "try the other one first, but if that does not work this option still might". I am open to suggestions how this could be presented in a way that is more accessible to users.

MobiFlight/Joysticks/VKB/VKBLedContainer.cs
cbrauers: Not possible on members.

Issue #1808, opened 2024-08-03
Buttons do not get sorted according to definition order


cbrauers:
**Describe the bug**
The functions to sort buttons and axes according to their order in the definition file do not actually return the buttons or axes in that order - The buttons function returns the original array based on DirectInput indexing, the axis function sorts the original array instead of a copy.

**Steps To Reproduce**
Connect a device where the defined button order is different from the DirectInput button order. I used a VKB FSM.GA with encoder API support, but a Honeycomb Bravo should do.
Create an Input config, select the device, and note the order of the dropdown.

**Expected behavior**
Items in the dropdown are in definition file order

**Actual behavior**
Items in the dropdown are in DirectInput order

**Details**
 - MobiFlight Version: 10.3 (with custom build for VKB encoders - should be reproducible on vanilla 10.3 with Honeycomb)

**Additional context**
If the code is changed to return sorted arrays, other issues can pop up:

- The sort function used, List.Sort, uses an unstable algorithm. If any DirectInput buttons on the device are not in the definition file, their order will be undefined instead of falling back to DirectInput order.
- DirectInput buttons which are not in the definition file get a sorting index of -1, and thus appear before inputs within the definition file in the dropdown.
- The functions that look up buttons in the definition file are currently not virtual / overridable, so sorting would not work for custom devices which have their own naming scheme.

PR #1811, opened 2024-08-04
Refactor joystick handling

cbrauers:
- Modularize EnumerateDevices into smaller overridable chunks
- Fix some issues with button and axis sorting and enable it

### DIFF
--- a/MobiFlight/Joysticks/Joystick.cs
+++ b/MobiFlight/Joysticks/Joystick.cs
@@ -96,11 +96,6 @@ namespace MobiFlight
             {
                 this.DIJoystick.GetObjectInfoById(device.ObjectId);
 
-                int offset = device.Offset;
-                int usage = device.Usage;
-                ObjectAspect aspect = device.Aspect;
-                String name = device.Name;
-
                 bool IsAxis = (device.ObjectId.Flags & DeviceObjectTypeFlags.AbsoluteAxis) > 0;
                 bool IsButton = (device.ObjectId.Flags & DeviceObjectTypeFlags.Button) > 0;
                 bool IsPOV = (device.ObjectId.Flags & DeviceObjectTypeFlags.PointOfViewController) > 0;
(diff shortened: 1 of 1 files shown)


MobiFlight/Joysticks/Joystick.cs
```diff
@@ -96,54 +96,22 @@ protected virtual void EnumerateDevices()
...
-
-                    } catch (ArgumentOutOfRangeException)
-                    {
-                        Log.Instance.log($"Axis can't be mapped: {DIJoystick.Information.InstanceName} Aspect: {aspect} Offset: {offset} Usage: {usage} Axis: {name} Label: {axisLabel}.", LogSeverity.Error);
-                        continue;
-                    }
-                    Axes.Add(new JoystickDevice() { Name = axisName, Label = axisLabel, Type = DeviceType.AnalogInput, JoystickDeviceType = JoystickDeviceType.Axis});
-                    Log.Instance.log($"Added {DIJoystick.Information.InstanceName} Aspect {aspect} + Offset: {offset} Usage: {usage} Axis: {name} Label: {axisLabel}.", LogSeverity.Debug);
-
+                    RegisterAxis(device);
```
cbrauers, line 106: This method has three sub-methods abstracted for more specific overrides in derived classes and to keep function length down - derived classes do not need to copy the full size of the method if all they want to change is a line or two.

MobiFlight/Joysticks/Joystick.cs
```diff
@@ -216,32 +246,33 @@ public List<ListItem<IBaseDevice>> GetAvailableDevicesAsListItems()
 
         private List<JoystickDevice> GetButtonsSorted()
         {
-            var buttons = Buttons.ToArray().ToList();
-            buttons.Sort(SortByPositionInDefintion);
-            return Buttons;
+            return Buttons.OrderBy(button => GetIndexForJoystickDevice(button)).ToList();
```
cbrauers, line 249: This addresses two nuisances (cf. issue #1808):

- The mix of member "Buttons" and local "buttons" led to the sorted list being completely ignored
- List.Sort() uses an unstable sorting algorithm, effectively applying an undefined permutation to buttons without a sorting index.

The Lambda could also be a function pointer, writing it this way mainly makes it more explicit that it is acting on individual buttons.

MobiFlight/Joysticks/Joystick.cs
```diff
@@ -216,32 +246,33 @@ public List<ListItem<IBaseDevice>> GetAvailableDevicesAsListItems()
...
         }
 
         private List<JoystickDevice> GetAxisSorted()
         {
-            var axes = Axes.ToArray().ToList();
-            Axes.Sort(SortByPositionInDefintion);
-
-            return Axes;
+            return Axes.OrderBy(axis => GetIndexForJoystickDevice(axis)).ToList();
```
cbrauers, line 254: Same as the buttons, with the added fun that the sort function was run on the member axes, not the local axes.

MobiFlight/Joysticks/Joystick.cs
```diff
@@ -216,32 +246,33 @@ public List<ListItem<IBaseDevice>> GetAvailableDevicesAsListItems()
...
         }
 
-        public int GetIndexForKey(string key)
+        public virtual int GetIndexForKey(string key)
         {
-            return Definition?.Inputs?.FindIndex(input => input.Name == key) ?? 0;
+            // -2 = definition not found / no inputs in definition
+            // -1 = index not found
+            // 0 or greater: valid index
+            return Definition?.Inputs?.FindIndex(input => input.Name == key) ?? -2;
```
cbrauers, line 262: Documented some outputs and changed the return value in case of null pointers from a valid index to something the caller can distinguish.
Added virtual keyword because this is something that derived classes may want to override: the naming scheme of derived classes is often not based on DirectInput numbers, whereas input.Name is always based on the "Button <ID>" scheme.

MobiFlight/Joysticks/Joystick.cs
```diff
@@ -216,32 +246,33 @@ public List<ListItem<IBaseDevice>> GetAvailableDevicesAsListItems()
...
         }
 
-        int SortByPositionInDefintion(JoystickDevice b1, JoystickDevice b2)
+        public int GetIndexForJoystickDevice(JoystickDevice button)
```
cbrauers, line 265: Enumerable.OrderBy, unlike List.Sort, allows us to define what property to sort by. Therefore, the approach to integrate the index retrieval into the function is somewhat different.

MobiFlight/Joysticks/Joystick.cs
```diff
@@ -216,32 +246,33 @@ public List<ListItem<IBaseDevice>> GetAvailableDevicesAsListItems()
...
         {
-            if (GetIndexForKey(b1.Name) == GetIndexForKey(b2.Name)) return 0;
-            if (GetIndexForKey(b1.Name) > GetIndexForKey(b2.Name)) return 1;
-            return -1;
+            var index = GetIndexForKey(button.Name);
+            // Ensure items in definition appear first
+            if(index < 0)
+            {
+                index = Int32.MaxValue; // force undefined stuff to the end of the list
```
cbrauers, line 271: The inputs most relevant to MobiFlight users are probably spelled out in the definition. Entries not in the definition are likely to be less relevant, so let us move them after the named ones.

MobiFlight/Joysticks/Joystick.cs
```diff
@@ -152,6 +120,69 @@ protected virtual void EnumerateDevices()
...
+            POV.Add(new JoystickDevice() { Name = $"{PovPrefix} {name}R", Label = $"{name} (→)", Type = DeviceType.Button, JoystickDeviceType = JoystickDeviceType.POV });
+            POV.Add(new JoystickDevice() { Name = $"{PovPrefix} {name}DR", Label = $"{name} (↘)", Type = DeviceType.Button, JoystickDeviceType = JoystickDeviceType.POV });
+            POV.Add(new JoystickDevice() { Name = $"{PovPrefix} {name}D", Label = $"{name} (↓)", Type = DeviceType.Button, JoystickDeviceType = JoystickDeviceType.POV });
+            POV.Add(new JoystickDevice() { Name = $"{PovPrefix} {name}DL", Label = $"{name} (↙)", Type = DeviceType.Button, JoystickDeviceType = JoystickDeviceType.POV });
+            POV.Add(new JoystickDevice() { Name = $"{PovPrefix} {name}L", Label = $"{name} (←)", Type = DeviceType.Button, JoystickDeviceType = JoystickDeviceType.POV });
+            POV.Add(new JoystickDevice() { Name = $"{PovPrefix} {name}UL", Label = $"{name} (↖)", Type = DeviceType.Button, JoystickDeviceType = JoystickDeviceType.POV });
+            LogInputAdded(device, name);
+        }
+
+        protected void LogInputAdded(DeviceObjectInstance device, string label)
```
cbrauers, line 168: Trading debug information (calling function name) for DRYness here. If having the log point to dedicated logging methods is unwanted, I have a different compromise in mind as an alternative option (generate the string in a method and call Log.Instance.log with that).

PR #1816, opened 2024-08-15
Fix crash on changing a customized presets

### LINKED ISSUE #1814: Inserting a custom variable in the Button Input OnRelease causes an Exception
**Describe the bug**
When I try to create an input with custom functions due to >> Button >> Input >> OnPress and OnRelease Mobiflight closes with the following exception:
**System.ArgumentOutOfRangeException: InvalidArgument=The value '-1' is invalid for 'index'.
Parameter name: index**

If you use the Hubhop database to set up the functions everything works fine.

**Steps To Reproduce**
1. Go to '>> Input >> OnPress and OnRelease'
2. Write custom functions for both states
3. Apply
4. See exception

**Expected behavior**
Added the custom functions that generated the exception into Hubhop, updated Hubhop presets in MF and rewrote the configuration: now the exception is not generated.

**Details**
 - MobiFlight Version: [10.3.0.1]
 - FlightSim: [MSFS2020]

image - 2024-08-14T092117 217

 - AddOn: [not relevant]

Koseng:
fixes #1814 

The display entry to show "Customized Preset or Custom Code" was appended to the end of the preset list. Since now only the first 1500 items are shown, that entry was cut as well, which caused issues.

Now the display entry "Customized Preset or Custom Code" is insert as first item of the preset list.

### DIFF
--- a/UI/Panels/Config/HubHopPresetPanel.cs
+++ b/UI/Panels/Config/HubHopPresetPanel.cs
@@ -393,7 +393,7 @@ namespace MobiFlight.UI.Panels.Config
                     code = Code,
                     description = "This is a customized preset or custom code."
                 };
-                FilteredPresetList.Items.Add(CustomPreset);
+                FilteredPresetList.Items.Insert(0, CustomPreset);
                 FilteredPresetListChanged();
                 // We have found the original preset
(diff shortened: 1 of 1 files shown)


Koseng:
Fix was successfully tested by discord user.

PR #1829, opened 2024-10-01
support for multiple configuration lines on same display

niddu85:
with this small change is possible to define multiple configuration lines on Mobiflight Connector in order to manage different parts of the display on different configuration lines.
It's enough to define an escape character in the display configuration to skip that character from being shown on the display

Config line 1
image

Config line 2
image

Result
image

This PR should be used together with 

https://github.com/MobiFlight/MobiFlight-FirmwareSource/pull/336


niddu85:
Hi maintainers,
Is this feature interesting to be merged or shall I close the PR?
I'm using it since I created it with no issues.

B.R.

PR #1839, opened 2024-10-14
Default `TimeoutForFirmwareUpdate` to `Timeout`

### LINKED ISSUE #1838: `TimeoutForFirmwareUpdate` defaults to 15000, but should default to `Timeout` instead
Found by Ralf. When flashing large firmware files to the Mega, Mobiflight will attempt to query via GetInfo before the flashing is complete, even if you set `Timeout` to a higher value.

This is because by default the `TimeoutForFirmwareUpdate` property is omitted from the mega.board.json file and you have to know to set it.

Instead, we should just default `TimeoutForFirmwareUpdate` to the value of `Timeout` if no specific one is specified in the board.json file.

neilenns:
Fixes #1838

* Use `Timeout` for the value of `TimeoutForFirmwareUpdate` if it wasn't specified in the board.json file
* Fall back to `15000` if neither is specified

### DIFF
--- a/MobiFlight/Board.cs
+++ b/MobiFlight/Board.cs
@@ -139,7 +139,7 @@ namespace MobiFlight
         /// <summary>
         /// Number of milliseconds to wait for the firmware update to complete before attempting to call GetInfo on the board.
         /// </summary>
-        public int TimeoutForFirmwareUpdate = 15000;
+        public int TimeoutForFirmwareUpdate;
     }
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


elral:
I tested the latest built with a large firmware where the time for flashing  takes ~20sec.
Updating and uploading on a fresh board is working without having the setting `TimeoutForFirmwareUpdate` in the board.json file. Only `Timeout` is set to 25000 ms.

MobiFlight/Board.cs
```diff
@@ -414,6 +414,23 @@ public void Migrate()
...
+
+            // Issue 1838:
+            // Handle undefined TimeoutForFirmwareUpdate by setting it to the value for Timeout.
+            if (Connection.TimeoutForFirmwareUpdate == 0)
+            {
+                // This test is required to ensure AvrDudeSettings actually exists. It may not, for example when
+                // loading a board.json file for something like a Pico.
+                if (AvrDudeSettings != null)
+                {
+                    Connection.TimeoutForFirmwareUpdate = AvrDudeSettings.Timeout != 0 ? AvrDudeSettings.Timeout : 1500;
```
neilenns, line 426:
```suggestion
                    Connection.TimeoutForFirmwareUpdate = AvrDudeSettings.Timeout != 0 ? AvrDudeSettings.Timeout : 15000;
```

PR #1862, opened 2024-11-23
Add a swap button for encoder pins [Feature suggestion]

fredimachado:
Hey all!

I'm not sure it only happens to me, but I usually have to swap the encoder pins, at it was a bit of a pain for me to manually assign to free pins and then fix it.

So I decided to add a Swap button to make this hassle free.

Here's my implementation in case you think it's a good feature to add to the project.
I had to move the right pin combo box and label a bit to the right so everything is aligned.
Let me know if I need to change or fix anything.

Thanks a lot for this App, it's amazing!

image


neilenns:
Love it!!!

PR #1868, opened 2024-11-28
Don't try and install the WASM module for MSFS versions that aren't installed

### LINKED ISSUE #1867: Don't attempt to install the WASM if a community folder isn't found
With the addition of MSFS20204 support in #1825 we now have a situation where someone might only have one of the two sim versions installed. This is normal and expected. However, when installing the wasm module, it leads to the following in the logs:

```
11/26/2024 22:27:08(471): WasmModuleUpdater.AutoDetectCommunityFolder(): Attempting to auto-detect community folder location for MSFS 2024
11/26/2024 22:27:08(471): WasmModuleUpdater.ExtractCommunityFolderPath(): No UserCfg found at C:\Users\cgbre\AppData\Roaming\Microsoft Flight Simulator 2024\UserCfg.opt
11/26/2024 22:27:08(472): WasmModuleUpdater.ExtractCommunityFolderPath(): No UserCfg found at C:\Users\cgbre\AppData\Local\Packages\Microsoft.Limitless_8wekyb3d8bbwe\LocalCache\UserCfg.opt
11/26/2024 22:27:08(480): WasmModuleUpdater.InstallWasmModule(): WASM module cannot be installed. Community folder '' not found.
```

We've already had two users in Discord who, for some reason, run with logs enabled and see the error and think it's the cause of their issue (it wasn't, wasm was running fine).

We shouldn't be logging an error at all here. If no UserCfg is ever found we should just skip attempting to install the WASM at all for that version of MSFS.

neilenns:
Fixes #1867

### DIFF
--- a/SimConnectMSFS/WasmModuleUpdater.cs
+++ b/SimConnectMSFS/WasmModuleUpdater.cs
@@ -188,7 +188,7 @@ namespace MobiFlight.SimConnectMSFS
             string installedWASM;
             string mobiflightWASM;
 
-            if (communityFolder == null) return true;
+            if (String.IsNullOrEmpty(communityFolder)) return true;
 
             string wasmModulePath = Path.Combine(communityFolder, @"mobiflight-event-module\modules\", WasmModuleName);
             if (!File.Exists(wasmModulePath))
--- a/UI/MainForm.cs
+++ b/UI/MainForm.cs
@@ -2143,13 +2143,13 @@ namespace MobiFlight.UI
                 }
 
                 // Try updating the 2020 install
-                if (Is2020Different)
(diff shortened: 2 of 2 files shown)


SimConnectMSFS/WasmModuleUpdater.cs
```diff
@@ -140,7 +140,6 @@ public bool InstallWasmModule(string communityFolder)
             // Remove the old Wasm File
             DeleteOldWasmFile(communityFolder);
 
-            Log.Instance.log("WASM module has been installed successfully.", LogSeverity.Info);
```
neilenns, line 143: The caller now logs this based on the return value so different log messages can be sent based on the simulator version.

neilenns:
@DocMoebiuz I unwound the if statements, thanks for the suggestion. So much better.

PR #1870, opened 2024-11-28
Check FW capabilities against CoreVersion

elral:
With the introduction of community devices a CoreVersion was added where the actual FW is based on. This CoreVersion must be used to check the FW capabilities and not the releasd version of the FW.

1. Check FW capabiliteies against `CoreVersion` and not `Version`
2. Add `CoreVersion` to the class to remember it.
3. If no `CoreVersion` is provided and fallback check fails, set `CoreVersion` to `Version`. This reflects all standard FW below 2.5.0 whithout `CoreVersion`. In this case they are always the same.

Remark: I am not sure if `FallbackCoreVersion(string version, Board board)` is really required with this change. 
1. Partner Boards are based on FW versions > 2.5.0 and report a `CoreVersion` which is used -> OK
2. Partner Boards have a FW version > 2.5.0 and do not provide a `CoreVersion`, then `CoreVersion` is set to `Version` -> OK
3. Partner Boards have a FW version <= 2.5.0 and do not provide a `CoreVersion` **but** they are based on `CoreVersion` > 2.5.0. In this case it will fail, but as far as I know that is never the case.

### DIFF
--- a/MobiFlight/MobiFlightModule.cs
+++ b/MobiFlight/MobiFlightModule.cs
@@ -242,6 +242,7 @@ namespace MobiFlight
         {
             Name = moduleInfo.Name ?? moduleInfo.Board.Info.FriendlyName;
             Version = moduleInfo.Version;
+            CoreVersion = moduleInfo.CoreVersion;
             Serial = moduleInfo.Serial;
             _comPort = moduleInfo.Port;
             Board = moduleInfo.Board;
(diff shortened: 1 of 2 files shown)


MobiFlight/MobiFlightModule.cs
```diff
@@ -910,16 +911,29 @@ public IModuleInfo GetInfo()
...
+                    // we also introduced CoreVersion
+                    // if a CoreVersion was not provided 
+                    // we determine a fallback version
+                    devInfo.CoreVersion = FallbackCoreVersion(devInfo.Version, devInfo.Board);
+                    // if no guess is possible, set CoreVersion to Version
+                    // this reflectes all standard boards with FW version below 2.5.0
+                    if (String.IsNullOrEmpty(devInfo.CoreVersion))
+                    {
+                        devInfo.CoreVersion = devInfo.Version;
+                    }
```
neilenns, line 930: Can this be moved inside FallbackCoreVersion, so it returns Version instead of null?

elral: Yeah, I was also thinking about this. I left it here as I am still thinking that the fallback function is not required. If I am wrong, I will move it.

PR #1873, opened 2024-11-30
Prevent WASM install when MSFS is running

### LINKED ISSUE #1872: Remind users to restart the sim after installing WASM
We've had several cases in Discord lately where users install the WASM module while the sim is running then wonder why things don't work.

We should have a dialog that shows after WASM install that says "You must restart your simulator" if wasm is installed while the sim is running. It shouldn't be an auto-timeout dialog either, it should stay there until the user clicks ok.

Shouldn't be too hard to do, we already have a flag for the sim being connected.

neilenns:
Fixes #1872 

Dialog looks like this:

image

It only appears if MSFS is the detected running sim, and will not show if the sim is closed.

### DIFF
--- a/ProjectMessages/ProjectMessages.Designer.cs
+++ b/ProjectMessages/ProjectMessages.Designer.cs
@@ -1428,7 +1428,7 @@ namespace MobiFlight.ProjectMessages {
         /// <summary>
         ///   Looks up a localized string similar to The WASM module cannot be installed while Microsoft Flight Simulator is running.
         ///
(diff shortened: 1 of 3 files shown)


neilenns:
@DocMoebiuz German string added

neilenns:
No more or less than it was before 😂 They have to go to the extras menu.

neilenns:
Can you give me the German version?

neilenns:
@DocMoebiuz updated

neilenns:
@DocMoebiuz image updated

PR #1875, opened 2024-12-04
Use an array of InstanceNames

neilenns:
Draft PR of a change to use an array of `InstanceNames` to identify joysticks instead of a single `InstanceName`, based on a report in Discord: https://discord.com/channels/608690978081210392/1313961757672472687


neilenns:
Fun exercise but confirmed not needed :)

Issue #1876, opened 2024-12-07
Repeated commands are lost with the WASM module


Nezz:
**Describe the bug**
I noticed an issue with the Winwing FCU (and EFIS) knobs, but it might affect other devices too that send lots of repeated commands.

When the Winwing FCU's knobs are turned quickly (see additional context), some commands don't get executed by the WASM module.

**Steps To Reproduce**
1. Configure the heading knob to execute the following:
`1 (A:AUTOPILOT HEADING LOCK DIR:1, degrees) 1 - dnor (>K:2:HEADING_BUG_SET)`
2. Start a flight with the 172, turn on avionics
3. Turn the knob quickly

**Expected behavior**
The heading changes by the amount of commands sent (see [this code](https://github.com/MobiFlight/MobiFlight-Connector/blob/fa270406f0a4729c7ae31c54782d8036b5ff9a70/SimConnectMSFS/WasmModuleClient.cs#L72))

**Actual behavior**
`MF.SimVars.Set` is sent to the WASM module the correct times, but it does not change the heading correctly. The slower the sim runs, the less the heading will change.

**Details**
 - MobiFlight Version: 10.4.0.1 and lastest master (fa27040)
 - FlightSim: MSFS 2024
 - AddOn: Tested with default G1000 or G3000 aircraft, possibly others

**Additional context**

Mobiflight exposes the knobs as buttons, and it simulates a button press for each increment:
https://github.com/MobiFlight/MobiFlight-Connector/blob/fa270406f0a4729c7ae31c54782d8036b5ff9a70/MobiFlight/Joysticks/WinwingFcu/WinwingFcu.cs#L156-L160

This can result in dozens of events sent to the sim at the same time. I debugged this issue and it seems that Mobiflight Connector sends the commands to the WASM module correct, so the issue is either with simconnect not liking this many events, or the WASM module not handling them.

PR #1877, opened 2024-12-07
Add support for Winwing FCU knobs with acceleration

Nezz:
The Winwing FCU currently exposes its knobs using buttons, but I encountered an issue with this approach:
https://github.com/MobiFlight/MobiFlight-Connector/issues/1876

To work around this, I added the knobs as axes because that's how they are implemented by Winwing under the hood. The value of the axis is changed depending on how strong acceleration is set in SimAppPro.

Note that even the lowest acceleration setting has some acceleration, so if you don't want any at all then you should keep using the buttons.

### DIFF
--- a/MobiFlight/Joysticks/Joystick.cs
+++ b/MobiFlight/Joysticks/Joystick.cs
@@ -22,8 +22,8 @@ namespace MobiFlight
         public event ButtonEventHandler OnButtonPressed; 
         public event EventHandler OnDisconnected;
 
-        protected List<JoystickDevice> Buttons = new List<JoystickDevice>();
-        private readonly List<JoystickDevice> Axes = new List<JoystickDevice>();
+        protected readonly List<JoystickDevice> Buttons = new List<JoystickDevice>();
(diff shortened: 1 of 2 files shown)


Koseng:
I'll look into it.
@Nezz I do not understand, What does it help to map the encoders/knobs to an axis/analog type? What we would like is to map them to an encoder type in mobiflight (which is currently not supported by joysticks, we would need to add it).

How do you use that in an actual mobiflight config? Do you have an example on how to use that?
Thanks!

Nezz:
>What does it help to map the encoders/knobs to an axis/analog type? What we would like is to map them to an encoder type in mobiflight

Initialy I implemented this by adding the encoder type, but that has:
- Left
- Left Fast
- Right
- Right Fast

Whereas Winwing gives us information like "1 left" and "8 right", which cannot be mapped directly to an encoder event.

Here is how I use it in MSFS 2024 with Garmin G1000/G3000 equipped planes:
https://gist.github.com/Nezz/c378369e1d714b4cc84fd878209cc43e

Nezz:
Do you want me to change it to use encoders? It should be easy.

PR #1886, opened 2024-12-23
Sending deviceID for outputs instead of pin number

### LINKED ISSUE #1885: Command for outputs should have deviceID and not pin number
For all devices except outputs the connector send a command where the device is identified by the device type (e.g. LED segment, outputshifter) and the deviceID. DeviceID means it is the 1st, 2nd, 3rd and so on device of the same type.

For outputs the command is slightly different. Instead of sending the deviceID, the pin number gets send.
For harmonization and for future usage/extensions it would be great to change this also to send the deviceID.

elral:
Instead of sending a pin number for outputs the deviceID is send like for all other devices. This is for harmaonization and possible future extensions.
PR https://github.com/MobiFlight/MobiFlight-FirmwareSource/pull/346 must be merged at the same time.
Fixes #1885


neilenns:
Is this *not* a breaking change correct? Everything continue to work properly if users don't upgrade their firmware and continue to use older firmware with newer desktop app?

elral:
Yes, correct.
Tested it in all combinations (OK, not so much).

elral:
A new PR will be issued which includes the proposed changes.

The branch is already set up from @DocMoebiuz , see [branch](https://github.com/MobiFlight/MobiFlight-Connector/tree/1885-output-uses-device-index)

The changes are already tested successful.

elral:
A new PR will be issued which includes the proposed changes.

The branch is already set up from @DocMoebiuz , see [branch](https://github.com/MobiFlight/MobiFlight-Connector/tree/1885-output-uses-device-index)

The changes are already tested successful.

PR #1888, opened 2024-12-26
Add a joystick dropdown to the status bar

neilenns:
Adds a joystick dropdown to the status bar listing the connected USB HID devices.

image

I put it in front of Modules because there's no need for a status icon and it looked weird if it was after Modules.


neilenns:
> Why remove the status icon? I think people are used to looking for that. I would also have to check the status icon... does it consider missing devices in current config... i am not sure.

The only thing that controlled the status icon was having a non-zero number of modules connected. It doesn't convey any information beyond "at least one module is connected" and is a constant source of posts in Discord from people who only use joysticks wondering why there's a warning showing.

Even if I did change it to hide if at least one joystick was connected, then it would likely always be hidden.

The new UI directly shows you what's connected in each category. The icon just leads to confusion, so it's gone :)

neilenns:
@DocMoebiuz It's back.

PR #1917, opened 2025-01-15
Don't throw error on WASM install when MSFS2024 isn't installed

### LINKED ISSUE #1916: WASM install reports failure if only 2020 is installed
Yes, this bug still exists despite prior fixes.

neilenns:
Fixes #1916 

* Return `true` if the community folder isn't found instead of `false`
* Get rid of unnecessary parameter since the check happens elsewhere
* Make sure logs happen when one, but not both, are skipped due to up-to-date

### DIFF
--- a/UI/MainForm.cs
+++ b/UI/MainForm.cs
@@ -2253,12 +2253,12 @@ namespace MobiFlight.UI
 
                 if (Is2020Different)
                 {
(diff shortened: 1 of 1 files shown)


neilenns:
The problem is `HandleWasmInstall` returns `false` if the community folder isn't there, which means the `Update2024Successful` becomes false and we throw the error.

Making it return `true` in this case is sufficient and makes sure we don't throw the error incorrectly.

PR #1918, opened 2025-01-15
Use TryParse and log

neilenns:
Test PR to add logging so we can get a private build for Stefano to try.

### DIFF
--- a/UI/Panels/Device/MFCustomDevicePanelPin.cs
+++ b/UI/Panels/Device/MFCustomDevicePanelPin.cs
@@ -4,7 +4,6 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Windows.Forms;
-using static System.Windows.Forms.VisualStyles.VisualStyleElement;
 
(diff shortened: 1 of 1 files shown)


Krakend70:
It does not work for me... still the same message... @neilenns

neilenns:
> It does not work for me... still the same message... @neilenns

It should work fine, several other users have confirmed. Make sure you are running the MFConnector.exe from the downloaded zip file. If you have additional issues with it please open a thread in #mobiflight in the Discord

Krakend70:
Just to make sure: You do run the installer and when the user interface
opens, it should work? Or you need to run the MFConnector.exe instead ? In
this latter case, you run the MFConnector.exe alone or after the installer
? Thanks in advance for your help !

Le jeu. 30 janv. 2025 à 18:05, Neil Enns ***@***.***> a
écrit :

neilenns:
Don't run the installer. Just run `MFConnector.exe` from the extracted folder. If you have additional issues with it please open a thread in #mobiflight in the Discord

PR #1935, opened 2025-02-07
MobiFlight 11 - New UI

DocMoebiuz:
This PR serves as a preview for the current MobiFlight 2025 UI rebrush efforts. 
This will soon be merged into `main`

### DIFF
--- a/UI/MainForm.cs
+++ b/UI/MainForm.cs
@@ -1726,14 +1726,18 @@ namespace MobiFlight.UI
 
             if (!merge)
             {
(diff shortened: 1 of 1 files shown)


Koseng:
When we start versions with new UI, maybe we should take the chance and drop some old stuff in the same timeframe.
- Drop Arcaze support
- Drop 32bit support, use newer simconnect.dll
- Drop loading event presets. Still hundreds of events are loaded in SimConnectCache. That is still active.
- ...

PR #2042, opened 2025-03-23
DRAFT: Add support for ProSim for the A322

tristanmcpherson:


Koseng:
Great!
Since the Prosim Userbase is probably smaller than the regular user base, I thought not force everyone to the "pythonnet" package. Everyone will be annoyed again by the popup message. I thought about making a special check for that, only in case the prosim scripts are started. So a new script dependent check and message.

tristanmcpherson:
Sounds good to me. I can look to add that

Koseng:
But on the other hand, that package also might be valuable in the future and we are still in beta. So for now just leave it like it is.

Scripts/Winwing/prosim_a320_winwing_cdu.py
```diff
@@ -0,0 +1,384 @@
...
+            return True
+        except Exception as e:
+            logging.error(f"ProSim setup failed for {self.cdu_name}: {e}")
+            return False
+
+    async def process_prosim(self) -> None:
+        while True:
+            dataref = self.prosim_client.get_dataref_value(self.cdu_dataref_name)
+            if dataref:
+                asyncio.run_coroutine_threadsafe(self.mobiflight.send(create_mobi_json(dataref)), self.event_loop)
```
Koseng, line 331: Since we are polling, I would propose to store the current value and only send to mobiflight if the value has changed. There is also a check before sending to the winwing device, but not sending at all to mobiflight would be the best. I assume each 100ms the same value is returned.

tristanmcpherson: Yep exactly, if you check the 737 implementation, I have exactly that, where I store the last value sent and only send if there is a diff

tristanmcpherson: The issue with the callback solution is that pythonnet runs it on the C# side, and it has problems interoping with asyncio. I was never able to get the callback solution to work nicely, so I think polling with checking should be good enough most likely

Koseng:
@tristanmcpherson will you adjust that PR to the GraphQL based version? And then please adjust the title to add the 737.

tristanmcpherson:
Can do!

PR #2061, opened 2025-03-28
Fix crash with MSFS2024 SU2

### LINKED ISSUE #2060: Crash when attempting to build community folder path
**Describe the bug**
Reported in Discord. We can crash if the Path.Combine fails. 

```
************** Exception Text **************
System.ArgumentException: Illegal characters in path.
   at System.IO.Path.CheckInvalidPathChars(String path, Boolean checkAdditional)
   at System.IO.Path.Combine(String path1, String path2)
   at MobiFlight.SimConnectMSFS.WasmModuleUpdater.ExtractCommunityFolderFromUserCfg(String UserCfg)
   at MobiFlight.SimConnectMSFS.WasmModuleUpdater.ExtractCommunityFolderPath(String[] basePaths)
   at MobiFlight.SimConnectMSFS.WasmModuleUpdater.AutoDetectCommunityFolder()
```

We should catch the exception and just proceed like we couldn't find the folder.

neilenns:
Fixes #2060 

It looks like MSFS2024 SU2 adds a new property to the UserCfg.opt file: `InstalledPackagesPathNextBoot`. Unfortunately, this causes a crash when we try and build the path to the community folder, since we are doing a substring search for `InstalledPackagesPath`, and we wind up including `NextBoot` as part of the path we're building.

Fix by searching for the string with a space after it, and also breaking out of the loop after it is found.

### DIFF
--- a/SimConnectMSFS/WasmModuleUpdater.cs
+++ b/SimConnectMSFS/WasmModuleUpdater.cs
@@ -65,6 +65,7 @@ namespace MobiFlight.SimConnectMSFS
                 if (line.Contains("InstalledPackagesPath "))
                 {
                     InstalledPackagesPath = line;
+                    break;
                 }
             }
 
--- a/SimConnectMSFS/WasmModuleUpdater.cs
+++ b/SimConnectMSFS/WasmModuleUpdater.cs
@@ -62,7 +62,7 @@ namespace MobiFlight.SimConnectMSFS
 
             while ((line = file.ReadLine()) != null)
             {
-                if (line.Contains("InstalledPackagesPath"))
+                if (line.Contains("InstalledPackagesPath "))
                 {
                     InstalledPackagesPath = line;
                 }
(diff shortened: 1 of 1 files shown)


SimConnectMSFS/WasmModuleUpdater.cs
```diff
@@ -62,9 +62,12 @@ private String ExtractCommunityFolderFromUserCfg(String UserCfg)
 
             while ((line = file.ReadLine()) != null)
             {
-                if (line.Contains("InstalledPackagesPath"))
+                // Issue #2061: The space at the end is intentional, to ensure it only matches the whole string InstalledPackagesPath
+                // and not the InstalledPackagesPathNextBoot property added in MSFS2024 SU2.
+                if (line.Contains("InstalledPackagesPath "))
                 {
                     InstalledPackagesPath = line;
+                    break;
```
neilenns, line 70: This is the actual fix. Everything else is just try/catch exception logging.

Issue #2056, opened 2025-03-26
Performance impact results in High WebView CPU usage.


DocMoebiuz:
**Describe the bug**
Users have reported high CPU usage with the new UI.
They shared some configs.

**Steps To Reproduce**
1. Open the config
2. Run the config

**Expected behavior**
Low CPU usage

**Actual behavior**
High CPU usage

**Details**
 - MobiFlight Version: 10.5.2.5

PR #2067, opened 2025-03-30
Performance improvements & refactoring

DocMoebiuz:
- [x] Reduce the data volume sent during updates - status, raw, final
- [x] Memoize all cells to prevent re-renders
- [x] Throttle UI updates to 5HZ
- [x] Suspend UI updates when minimized
- [x] Refactor `ExecuteConfig` logic, create dedicated class
- [x] Fix Blink.Equals implementation
- [x] Disable console output for production build
- [x] Add unit tests

partially #2056


tristanmcpherson:
There seems to be some sort of issue with the virtualizer, I can't get it to display all of my items, only shows abotu 20 items, and scrolling does not trigger the virtualizer to return more items

PR #2116, opened 2025-04-12
Memer921 patch 2

memer921:
Corrected ProductId


Koseng:
@memer921 thanks for that, but we are currently already preparing a bigger PR for the PAP3. But please leave the PR open for now, we will compare the things.

 @DocMoebiuz do not merge for now.

Koseng:
Full support is now in:
#2135 and #2131

PR #2133, opened 2025-04-21
Fixed issue in Maddog wiwning script, now working with dual CDU

dmarras68:
Winwing script now working with CAPTAIN + FO, thanks to @Koseng for testing and spotting the issues.

### DIFF
--- a/Scripts/Winwing/maddogx_winwing_cdu.py
+++ b/Scripts/Winwing/maddogx_winwing_cdu.py
@@ -93,7 +93,7 @@ MDX_CDU_0_ID: int = 85
 MDX_CDU_0_DEFINITION: int = 802
 MDX_CDU_1_NAME: str = "MaddogX CDU2 Data Atrb"
 MDX_CDU_1_ID: int = 86
-MDX_CDU_1_DEFINITION: int = 802
+MDX_CDU_1_DEFINITION: int = 801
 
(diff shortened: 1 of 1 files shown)


Koseng:
I was worried that changing to PERIOD_VISUAL_FRAME has a bigger performance impact. In a short test I could not see a big impact.

Nervertheless for other planes it works with DATA_PERIOD_ON_SET, and we plan to move the common simconnect code to a common library. If that code is different it is not that easy to use the library.

We will merge it now, but could you check whether you could do something in the future, maybe on the plane side that it works with DATA_PERIOD_ON_SET? That is really strange.

@DocMoebiuz ready to be merged.

dmarras68:
> I was worried that changing to PERIOD_VISUAL_FRAME has a bigger performance impact. In a short test I could not see a big impact.
> 
> Nervertheless for other planes it works with DATA_PERIOD_ON_SET, and we plan to move the common simconnect code to a common library. If that code is different it is not that easy to use the library.
> 
> We will merge it now, but could you check whether you could do something in the future, maybe on the plane side that it works with DATA_PERIOD_ON_SET? That is really strange.
> 
> @DocMoebiuz ready to be merged.

@Koseng yes, I guess I can change plane code to have it working with DATA_PERIOD_ON_SET even when MF is launched after the aircraft is running

Issue #1691, opened 2024-02-19
[Redesign] Introduce Project


DocMoebiuz:
As a user I can create a project which contains some additional meta information and allows to group mutliple configs which improves user friendliness.

Confirmations:
* A new project can be created, saved and loaded (/) #1949 
* A project has 
  * a title, (/) - #2411
  * an option to define the simulator you want to use it with (/) #2411 
  * a list of aircraft that it is linked to #2819
  * control for autoRun enabled/disabled / auto-link / auto-load 
* one or many configuration file sections (/) - #2139 
* configuration files can
  * be links to the actual config files - #2532
  * embed a config file as copy (/) - #2139 
* the file format is json (/) #1949 
* there is a json schema in place to validate the json file

joeherwig:
for the meantime until this gets implemented, I wrote a small powershell script that you can add with filewatcher to be executed, to auto-merge the different configs.

## merge-configs.ps1
It's just a powershell-script to merge different configs based on a json-definition where you can add relative paths to the ps1 file.

### The config format
```
{
  "assets":[
    "A20N/A20N_Overhead_Electrics.mcc",
    "A20N/A20N_Overhead_Lights.mcc",
    "A20N/A20N_Overhead_Fuel.mcc",
    "A20N/A20N_Pedestal.mcc",
    "Generic/Primary_control_axis.mcc",
    "Generic/Trim_controls.mcc",
    "Generic/GNC_255.mcc"
  ]
}
```

#### The filename
is used for the merged file. So If you want to have a `A32nx_merged.mcc` as result, just name the JSON as `A32nx_merged.json` and place it at the same level as the merge-configs.ps1

#### Execute it
either manually via `.\merge-configs.ps1` or for instance with a file-watcher or in VSCode via "Run on Safe" script runner to build on each `*.mcc` or `*_merged.json` save.

```
# Function to combine and modify XML files
function Merge-MobiflightConfigs() {
  param(  
    [Parameter(
        Position=0, 
        Mandatory=$true, 
        ValueFromPipeline=$true,
        ValueFromPipelineByPropertyName=$true)
    ]
    [Alias('FullName')]
    [String[]]$FilePath
    ) 

    Write-Host $FilePath
    foreach($path in $FilePath)
    {
        # Select the configs files
        $inputDirectory = $PSScriptRoot
        $assetConfig = Get-Content -Path $path -Raw | ConvertFrom-Json
    
        # Create a new XML document
[...]

PR #2139, opened 2025-04-27
Multiple "config files" in project

DocMoebiuz:
image

- [x] Existing configs / projects can be loaded
- [x] You can add a new Config File
- [x] You can include an existing Config file
- [x] You can save an individual config file
- [x] You can rename config files
- [x] Switching between config files in one project is possible
- [x] All config items of all config files in the current project are evaluated
- [x] Orphaned serial assignment is working across all files
- [x] i18n
- [x] unit tests
- [x] UI tests

relates to #1691

### DIFF
--- a/frontend/public/locales/de/translation.json
+++ b/frontend/public/locales/de/translation.json
@@ -1,4 +1,9 @@
 {
+  "General" : {
+    "Action" : {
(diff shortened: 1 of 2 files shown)


Koseng:
Also always open the config wizard when opening a saved file with one output config entry.

PR #2146, opened 2025-05-04
Input log shows correctly with multiple configs

DocMoebiuz:
The log messages show correctly like before.
For each included config file the log messages show.

- [x] using same log levels as before
- [x] unit tests added


Matthisfr:
Same trouble on this beta for me

PR #2147, opened 2025-05-05
Null reference fix for null InputAction expression in Replace method

Reyuu:
This is a pretty self-explanatory quick fix.

**Problem:**
When the expression in Replace method of the abstract class InputAction is null we run into NullReferenceError. This leads to devices not being detected correctly down the line:
image

**Solution:**
Detect if the expression is null. If it is, return early.


Reyuu:
#2149 fixed the issue's underlying cause. Closing.

PR #2149, opened 2025-05-05
Fix missing input events

### LINKED ISSUE #2148: Beta-Bug: Inputs are not working correctly
**Describe the bug**
several users have reported that input events are not executed correctly in MSFS

**Steps To Reproduce**
1. Load config that was working before which contains config references
2. Run
3. Use button

**Expected behavior**
Button input event should still be processed and executed in MSFS

**Actual behavior**
Processing happens, but the event doesn't do anything.

**Details**
 - MobiFlight Version: 10.5.3.4
 - FlightSim: MSFS2020

A config was shared on discord with the issue.

DocMoebiuz:
fixes #2148


cbrauers:
Looks like it fixes the issue as I am observing it.

PR #2162, opened 2025-05-17
Fix minimum Python version check

Nezz:
Mobiflight requires Python 3.10. However, starting the MCDU with the FBW A320 using Python 3.10 fails:
```
ImportError: cannot import name 'StrEnum' from 'enum'
```

That import is only present in 3.11, so I'd suggest bumping the minimum version.

Docs change: https://github.com/neilenns-projects/mobiflight-docs/pull/318


Koseng:
@DocMoebiuz  can be merged.

PR #2192, opened 2025-06-07
Ignore FSUIPC aircraft name change events when directly connected to sim

### LINKED ISSUE #2191: Aircraft Name race condition when FSUIPC installed
**Describe the bug**

There's a race condition currently in MobiFlight which can be experienced when using X-Plane with the XPUIPC plugin installed (the FSUIPC API in X-Plane) whereby MobiFlight will connect via X-Plane Direct and XPUIPC. Both connectors trigger the `AircraftChanged` event which results in the aircraft name being set as one of two possible values.

**Steps To Reproduce**
1. Enable logging in Mobiflight
2. Install XPUIPC in X-Plane
3. Start Mobiflight and observe the logging window
4. Two events should appear with different aircraft naming

**Expected behavior**
I'm unsure as to whether it's by design that FSUIPC and SimConnect or XPUIPC and X-Plane Direct will both connect at the same time. I can understand there are scenarios where users might want to map to offsets or datarefs.

In such a scenario, I'd imagine there should be an order of priority in terms of which aircraft name is picked based on the detected simulator.

If the simulator is X-Plane then use the aircraft name produced by the Direct connection
If the simulator is MSFS then use the SimConnect reported aircraft name
Otherwise, if P3D/FSUIPC-capable then use FSUIPC

**Actual behavior**
 Mobiflight detects two aircraft name changes via both connection sources. For MSFS there is consistency since FSUIPC wraps SimConnect
Image

For X-Plane, the names will be different as XPUIPC is a translation layer rather than a wrapper.

This isn't too bad as the first part can be matched
Image

For the ToLiss A320, XPUIPC returns an empty string
Image

For the ToLiss A319, XPUIPC simply returns "A319"
Image
Here's a scenario where XPUIPC won and Mobiflight used its name
Image

The X-Plane direct connect name has won in most scenarios but this is the race-condition. Either of the 2 names might end up being picked in Mobiflight.

**Details**
 - MobiFlight Version: 10.5.3.7
 - FlightSim: X-Plane
 - AddOn: ToLiss airbus series

**Additional context**
This is an issue under the new ScriptRunner concept which relies on pattern matching for an aircraft since in some cases XPUIPC will return a more generic aircraft name or a different name entirely (requiring 2 ScriptMapping entries). For example, under X-Plane Direct connect the ToLiss A319 will return "toliss a319 hi def" while XPUIPC returns "a319". Adding "a319" to the ScriptMappings may result in edge cases where Python scripts run for the wrong addons across different simulators.

Unfortunately, while XPUIPC is a very old plugin, some development teams (of flight trackers and career apps) insist on using it to interface with X-Plane and MSFS across a single API. It is refreshing that the MobiFlight team make a direct connection to X-Plane.

MobiFlight also connects to SimConnect and FSUIPC on MSFS but the aircraft names are consistent since FSUIPC echoes the SimConnect response.

ColinM9991:
This resolves #2191 by ignoring FSUIPC Aircraft Changed events when either X-Plane or SimConnect are directly connected. I've gone with the path of least resistance on this one as, based on Discord searches, I can see that users are using both FSUIPC and SimConnect or XPUIPC and X-Plane direct for synchronizing their profiles against both sims, so both connections would be required.

Rather than redesigning parts of the events pub/sub, it's easier to ignore the events given how infrequently they should trigger.

I've also removed a superfluous `if (true) {}` condition from `AutoConnectTimer_TickAsync`

### DIFF
--- a/MobiFlight/ExecutionManager.cs
+++ b/MobiFlight/ExecutionManager.cs
@@ -934,53 +934,51 @@ namespace MobiFlight
 #endif
 
             // Check only for available sims if not in Offline mode.
-            if (true)
+            if (SimAvailable())
             {
-
-                if (SimAvailable())
+                if (LastDetectedSim != FlightSim.FlightSimType)
                 {
-                    if (LastDetectedSim != FlightSim.FlightSimType)
-                    {
-                        LastDetectedSim = FlightSim.FlightSimType;
-                        OnSimAvailable?.Invoke(FlightSim.FlightSimType, null);
-                    }
-
-                    if (!fsuipcCache.IsConnected())
-                    {
-                        if (!simConnectCache.IsConnected() && !xplaneCache.IsConnected())
-                        {
-                            // we don't want to spam the log
-                            // in case we have an active connection
-                            // through a different type
-                            Log.Instance.log("Trying auto connect to sim via FSUIPC", LogSeverity.Debug);
-                        }
-
-                        fsuipcCache.Connect();
-                    }
-#if SIMCONNECT
-                    if (FlightSim.FlightSimType == FlightSimType.MSFS2020 && !simConnectCache.IsConnected())
-                    {
-                        Log.Instance.log("Trying auto connect to sim via SimConnect (WASM)", LogSeverity.Debug);
-                        simConnectCache.Connect();
-                    }
-#endif
-                    if (FlightSim.FlightSimType == FlightSimType.XPLANE && !xplaneCache.IsConnected())
-                    {
-                        Log.Instance.log("Trying auto connect to sim via XPlane", LogSeverity.Debug);
-                        xplaneCache.Connect();
-                    }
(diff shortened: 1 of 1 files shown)


ColinM9991:
> Please add unit tests for this so that we have regression testing for the future.

@DocMoebiuz tests have now been added to cover the happy-path scenarios and both FSUIPC conflicts with X-Plane and SimConnect independently. As we discussed on Discord, I've added another constructor to inject the interfaces. There's a default constructor chained to the new constructor for the benefit `MainForm`. All relevant code has been updated and some new properties/methods added to the `SimConnectCacheInterface` where dependents were previously using the `SimConnectCache` class directly.

ColinM9991:
Another point worth raising is that I've removed the `#if SIMCONNECT #endif` preprocessor directives as these appeared to be an overhang from the initial SimConnect implementation in 2020. All compilation options seemed to have this enabled

PR #2197, opened 2025-06-11
Add Uno VID/PID

### LINKED ISSUE #2196: Add missing VID/PID
VID_0843&PID_5740

[It's a knockoff Uno with a counterfeit CH340G.](https://discord.com/channels/608690978081210392/1382319620819783772)

neilenns:
Fixes #2196


neilenns:
We support VID/PIDs for other counterfeit CH340Gs

elral:
Yeah, but this chip is a count of strange. It uses a VID/PID which is not registered.

It looks like this chip needs also the "old" CH340 driver, see also this [discussion](https://www.eevblog.com/forum/microcontrollers/yet-another-chinese-usb-serial-ic-on-arduino/?all)

Furthermore these boards are using an ATmega328P**B** chip which has two more pins which are also breaked out and not an ATmega328P. If we want to support these both pins a new board.json file would be required which would mean setting up a new board.

neilenns:
Ok, let's not do this

PR #2200, opened 2025-06-11
Use the WinWing CDU screen for error messages and user info

### LINKED ISSUE #2199: Use WinWing CDU screen to show errors or feedback to user
When the WinWing CDU is not working as expected, it is hard to see the reason. It would be great if the CDU screen would also output error messages or information to the user what is going on. For example which python script is started for execution.

Show errors and status info on the CDU screen itself.

This is in addition to the logging.

Koseng:
fixes #2199 
Use the WinWing CDU screen for error messages and user info.
It is implemented in a generic way, that game controllers, which have the means to show an output to the user, can display status and error messages or codes. Could be used by game controllers which offer a full text display or just some numerical output.

IMG_7051
IMG_7052
IMG_7053


MobiFlight/Joysticks/Winwing/WinwingCdu.cs
```diff
@@ -93,5 +97,121 @@ public override void Shutdown()
...
+        /// display.</remarks>
+        /// <param name="lines">A collection of strings representing the lines of text to be displayed. Each line is centered horizontally
+        /// and the entire set of lines is centered vertically within the display.</param>
+        private void SendToDisplay(IList<string> lines)
+        {
+            // Center vertically          
+            int emptyLineCount = (14 - Math.Min(lines.Count, 14)) / 2;
+            var linesToDisplay = Enumerable.Repeat(string.Empty, emptyLineCount).Concat(lines).ToList();          
+
+            var cduData = new CduChar[336]; // 14 * 24
```
ColinM9991: I appreciate I'm just a user and not a member of the review group so please feel free to tell me to get lost if you disagree.

Regarding the Columns, Rows and Cell numbers, is it worth using constants for these to add consistent naming and meaning behind the values?

For example:
```csharp
public static class CduConstants
{
    public const int COLUMNS = 24;
    public const int ROWS= 14;
    public const int CELLS = COLUMNS * ROWS;
}
```

That then provides reusable values and less need for comments

```
int emptyLineCount = (CduConstants.ROWS - Math.Min(lines.Count, CduConstants.ROWS)) / 2;
var cduData = new CduChar[CduConstants.CELLS];
```

Koseng: Hi, thanks. Will do. Every feedback is appreciated.

I'm also always critizing magic constants and the need for comments in my professional code reviews. Here in the spare time work, sometimes it is getting late in the evening and I want to get it out for review Otherwise it might take another week. So sometimes also not perfect towards my standards...

MobiFlight/Joysticks/Winwing/WinwingCdu.cs
```diff
@@ -93,5 +99,121 @@ public override void Shutdown()
...
+                        cduData[currentIndex] = new CduChar
+                        {
+                            Character = paddedLine[j]
+                        };                        
+                    }
+                    currentIndex++;
+                }
+            }
+            // Convert CduChar array to json.
+            var builder = new StringBuilder("{ \"Target\": \"Display\", \"Data\": [");
```
tristanmcpherson, line 201: What are your thoughts on just creating an object and then using either STJ or Newtonsoft to convert it?

ColinM9991: +1 for STJ if exploring that avenue. It's deeply integrated with `Span<T>` which makes it far more performance friendly compared to Newtonsoft. The original Newtonsoft developer actually worked on STJ under Microsoft.

Koseng: Only there is no STJ in .NET Framework 4.8 ;-). Some point in the future, the switch to current .NET versions will happen...

PR #2204, opened 2025-06-16
[X-Plane 12] Add CDU support for Zibo 737

ColinM9991:
Adds support for the Zibo 737 CDUs (pilot and co-pilot).

My approach is as follows:
1. Identify the available WINWING devices via Mobiflight
2. Create tasks to process each device display independently
3. Fetch the datarefs for the corresponding devices in X-Plane
4. Subscribe to the dataref updates
5. Push event to the device-specific queue.

The websocket connections are performed using `async for`. If a connection is closed then the message is put back on the queue, the while loop is exited and then `async for` reinitializes the connection again. The message is then picked up from the queue and dispatched to the device.

The benefit of this is that if the connection is lost for any moment, the end user will receive only a second or several second delay in updating the display while the connection is created again in the background.

I'm open to feedback and suggestions on the preferred approach for logging and fault tolerance.

### DIFF
--- a/Scripts/Winwing/zibo_737_800x.py
+++ b/Scripts/Winwing/zibo_737_800x.py
@@ -7,16 +7,16 @@ In order to support multiple CDU devices seamlessly, a dynamic approach is taken
 A device is considered "supported" if it exists in the aircraft. Some aircraft have 3 CDUs while others have 2.
 Each enum member is assigned a value that represents the X-Plane dataref identifier. Example: fmc1 of laminar/B738/fmc1/Line04_I.
 
-Upon script start, Mobiflight is probed (get_available_devices()) to detect the devices connected to the PC. Any device that returns a successful response is then tracked.
+Upon script start, MobiFlight is probed (get_available_devices()) to detect the devices connected to the PC. Any device that returns a successful response is then tracked.
 
 Two tasks are started independently for each avialable CDU device.
 1. handle_dataref_updates -> Listens to X-Plane's WebSocket server for dataref updates for that specific CDU and pushes an event to a queue
-2. handle_device_update   -> Listens to the queue and dispatches updates to Mobiflight to update that CDU
+2. handle_device_update   -> Listens to the queue and dispatches updates to MobiFlight to update that CDU
 
 Tasks are started independently for each CDU device to ensure each device can update quickly, particularly when players might be performing shared cockpit flights.
 
-Upon a failed connection while dispatching updates to Mobiflight, the handle_device_update function use `async for` with the websockets client. The failed message is put back in the queue, the loop continues to the next iteration which then reconnects again.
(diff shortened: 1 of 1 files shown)


ColinM9991:
I'll update this very shortly with font support now that #2208 is merged

Koseng:
Hi,
thanks a lot for you contribution, very nice code! And thanks for being patient, sometimes it takes time....

Please:

- Add some info logging, if successfully connected to a CDU. Also add some logging that the script for the zibo was started. Currently we have no other feedback in the UI, so the only way to see something is going on, is the logging window. Also I plan to add passing the log level of MobiFlight (info, debug) as a parameter to the python scipts. So then we can think about doing a more extensive debug log level in case of issues.

- Add the description of your approach as a comment header to the python script. It is very valuable information and otherwise will be lost.

- Comment the purpose of the elapsed time handling in handle_device_update in the code.

- I noticed when setting the log level in the websockets package to debug, that  the "\u00b0" Syntax causes internal exceptions. Therefore avoid the "\u" syntax and directly copy the respective character to the code like now done here:
https://github.com/MobiFlight/MobiFlight-Connector/blob/3aae54bfa7eba7371e493d91a278a0522d344efa/Scripts/Winwing/fenix_winwing_cdu.py#L12

ColinM9991:
@Koseng 

> * Add some info logging, if successfully connected to a CDU. Also add some logging that the script for the zibo was started. Currently we have no other feedback in the UI, so the only way to see something is going on, is the logging window. Also I plan to add passing the log level of MobiFlight (info, debug) as a parameter to the python scipts. So then we can think about doing a more extensive debug log level in case of issues.

Logging is on my list - I was hoping to seek some guidance from you on this based on the ScriptRunner system. Is `print()` sufficient?

> Comment the purpose of the elapsed time handling in handle_device_update in the code.

Will do. This is to prevent the websocket instance from being spammed as some users with weaker CPUs reported delays when spamming pages, but I'll add that as a comment.

> I noticed when setting the log level in the websockets package to debug, that the "\u00b0" Syntax causes internal exceptions. Therefore avoid the "\u" syntax and directly copy the respective character to the code like now done here:

This might be one that requires a change to the documentation too as I originally followed the [guidance outlined here](https://docs.mobiflight.com/game-controllers/winwing/winwing-cdu/dev-instructions/#special-chars)

Koseng:
Please do the logging approach as in the PMDG scripts.

ColinM9991:
> Please do the logging approach as in the PMDG scripts.

@Koseng job done. I've added logging with some info (for debugging), warnings (to help users troubleshoot if a device isn't working) and errors (connection issues).

Hopefully the header is sufficient. I've tried to cover all angles (that I could remember) without being _too_ verbose. Both other comments also addressed.

I'll hold off for now on making changes in the other PR so we can use this one as a guinea-pig, then I'll carry the changes over to the other scripts.

PR #2226, opened 2025-07-01
Fix Lowercase lettering in PMDG CDU

Qowy:
Sometimes the pmdg CDU has lowercase lettering without the lowercase flag set but using actual lowercase letters. The included Font that PMDG uses seems to just use smaller uppercase letters for lowercase so there is no difference, However Mobiflight displays these as actual lowercase letters. Examples include units during weight and balance and file names during State loading.

Alternative solution would be to change the included Font.
Additionally this probably also applies to the 737 but I cannot test this.

Sorry for the stray whitespace I used the online Editor..
But since this applies to additional Files maybe just make your own commit :D

### DIFF
--- a/Scripts/Winwing/pmdg_777_winwing_cdu.py
+++ b/Scripts/Winwing/pmdg_777_winwing_cdu.py
@@ -138,8 +138,10 @@ def create_mobi_json(data: bytes) -> str:
                 message["Data"][dst_idx] = []
                 continue
                 
-            try:
+            try:                
                 symbol: str = chr(data[src_idx])
+                is_lowercase: bool = symbol.islower()
+                symbol = symbol.upper()
                 color: int = data[src_idx + 1]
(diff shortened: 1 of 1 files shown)


Qowy:
perf init ingame:
image

Panel state ingame
image

perf init current (note the kg)
image

panel state current
image

perf init "fixed"
image

panel state fixed
image


I noticed that the boxes are now filled out. the substuition is probably not working correctly since '\u00EA' now will be '\u00CA' so changing to:

`elif symbol == '\u00CA': symbol = "\u2610"  # box` and all is well again
image

Koseng:
@Qowy ?

Qowy:
Sorry I do not have the 737 but I would expect the same Issue.

Yes I have not seen any arrows anywhere. But if a unicode point has no toUpper mapping (which special characters should not have) nothing should happen to them

Koseng:
Ready to be merged.

PR #2246, opened 2025-07-07
Add CDU scripts for FlightFactor and ToLiss

ColinM9991:
This adds Python scripts to support the following aircraft:

- ToLiss A319
- ToLiss A320
- ToLiss A321
- FlightFactor 757
- FlightFactor 767
- FlightFactor 777v2

This doesn't yet support the ToLiss A340 and A330. I plan to tweak the existing script to dynamically detect the center CDU for 3-way CDU support. This too could be carried over to the FlightFactor script, which will remove the need for one script, since that API is the same across all FF aircraft.

### DIFF
--- a/MobiFlightConnector.csproj
+++ b/MobiFlightConnector.csproj
@@ -223,6 +223,9 @@
     <Content Include="frontend\dist\**">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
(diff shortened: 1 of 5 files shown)


Koseng:
Thanks a lot, I know that was a lot of work!

Same remarks as in #2204. Also some remarks in the header about the general approach would be nice.

Koseng:
Ready to be merged.

PR #2257, opened 2025-07-16
[X-Plane] Map size correctly in 777v2 and use Boeing font

ColinM9991:
FlightFactor's 777v2 size dataref begins uses a different integer layout for representing text size. This means that all text is currently being displayed with a small font.
* 1 = Large
* 2 = Small

As such, this update subtracts 1 from the size integer to translate it to
* 0 = Large
* 1 = Small

### DIFF
--- a/Scripts/Winwing/flightfactor_777v2.py
+++ b/Scripts/Winwing/flightfactor_777v2.py
@@ -94,7 +94,7 @@ def get_size(size: int) -> int:
     # FlightFactor's 777v2 size starts at 1
     # 1 = large
     # 2 = small
-    return int(size == 2)
+    return 1 if size == 2 else 0
 
 
 def fetch_dataref_mapping(device: CduDevice):


Scripts/Winwing/flightfactor_777v2.py
```diff
@@ -123,7 +134,11 @@ def generate_display_json(device: CduDevice, values: dict[str, str]):
             if char == " ":
                 continue
 
-            display_data[index] = [get_char(char), get_color(color, effect), size]
+            display_data[index] = [
```
DocMoebiuz, line 136: Does it handle somewhere the case that this could potentially be out of bounds, eg in case for some reason the size calculation was called with 0?

ColinM9991: At the moment it works on the assumption that it's a standard value. I'll update this in the morning to default it to a large font if it's out of bounds, since small font is typically the overriden behaviour

Scripts/Winwing/flightfactor_777v2.py
```diff
@@ -87,6 +90,14 @@ def get_color(color: int, effect: int) -> str:
     return COLOR_MAP.get(color, "w")
 
 
+def get_size(size: int) -> int:
+    # FlightFactor's 777v2 size starts at 1
+    # 1 = large
+    # 2 = small
+    # Subtract 1 from dataref to normalize size format
+    return size - 1
```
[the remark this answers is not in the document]
ColinM9991: `max` wouldn't work for out of bounds values. New logic will set small font accordingly or default to large font

Scripts/Winwing/flightfactor_777v2.py
```diff
@@ -227,10 +242,12 @@ async def get_available_devices() -> list[CduDevice]:
...
         device_endpoint = device.get_endpoint()
         try:
-            async with websockets.connect(device_endpoint) as _:
+            async with websockets.connect(device_endpoint) as socket:
                 logging.info(
                     "Discovered CDU device %s at endpoint %s", device, device_endpoint
                 )
+
+                await socket.send(FONT_REQUEST)
                 available_devices.append(device)
```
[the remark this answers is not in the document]
ColinM9991: Added logging for font request and device is not added as a candidate device if the font request fails

Scripts/Winwing/flightfactor_777v2.py
```diff
@@ -87,6 +90,13 @@ def get_color(color: int, effect: int) -> str:
     return COLOR_MAP.get(color, "w")
 
 
+def get_size(size: int) -> int:
+    # FlightFactor's 777v2 size starts at 1
+    # 1 = large
+    # 2 = small
+    return int(size == 2)
```
DocMoebiuz: i know it is a detail - i just find this hard to read. Tertiary expression would make it more obvious dont you think

DocMoebiuz:
```suggestion
    return (size == 2) ? 1 : 0
```

ColinM9991: Python ternary syntax is slightly different with
`1 if size == 2 else 0`

I've made that change now

Scripts/Winwing/flightfactor_777v2.py
ColinM9991: Resolved with newer implementation for `get_size` which returns 1 if small font is required, otherwise defaulting to 0

Issue #2270, opened 2025-07-25
Automatically recognize joystick devices based on device type alone, avoid orphaned serials dialog


tigert:
**Is your feature request related to a problem? Please describe.**
When sharing pre-made configuration files for joystick devices, for example the WinWing panels, you get the orphaned serials dialog for a new config you download every time, even if you have just one panel of the type on your setup, because MobiFlight checks the serial number of all devices for a match when events are received or sent.

**Describe the solution you'd like**
In case of known joystick devices, when checking a config for orphaned events, we could ignore the serial number if there is only one compatible device connected. 

We should resort to the serial number in case you have many and thus the choice is ambiguous.

DocMoebiuz:
fixed by #2535

PR #2276, opened 2025-07-29
Automatically resolve simple orphaned serials

Nezz:
This was suggested in #2270.

The orphaned serials will be resolved automatically if there are only unique device names (e.g. one MCDU, one FCU) in a profile. If this happens successfully the dialog will not be displayed.

- [x] All device types are supported
  - [x] MobiFlight Controllers
  - [x] Joystick Controllers
  - [x] Midi Controllers
- [x] Automatic replacement
  - [x] For every device that is referenced in the profile but not connected, and
  - [x] if another device of same type and same name is found, it will be used 
- [x] No auto replacement if multiple devices are connected that match type and name
- [ ] User message presented in case:
  - [ ] automatic profile update happened successfully, now changes have to be saved (best to new file)
  - [ ] automatic profile was not possible because user has to manually select
  - [x] profile update was not possible because no matching device connected
- [ ] Unit tests are added

Context: I'm one of those people who only uses MobiFlight with off-the-shelf hardware. This makes things easier for people who download profiles from the web and don't make build their own hardware.

### DIFF
--- a/UI/MainForm.cs
+++ b/UI/MainForm.cs
@@ -485,7 +485,7 @@ namespace MobiFlight.UI
         private async Task CleanRecentFilesAsync()
         {
             var recentSnapshot = Properties.Settings.Default.RecentFiles.Cast<string>().ToList();
-            
+
             var missingFiles = await Task.Run(() => CheckForMissingFiles(recentSnapshot)).ConfigureAwait(false);
 
             if (missingFiles.Count == 0) return;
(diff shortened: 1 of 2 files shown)


UI/MainForm.cs
```diff
@@ -2016,22 +2016,33 @@ private void _checkForOrphanedSerials(bool showNotNecessaryMessage)
...
-                        }
-
-                        MessageExchange.Instance.Publish(execManager.Project);
+                        UpdateProject();
                     }
                 }
+                else if (opd.HasChanged())
+                {
+                    // If there are no orphaned serials, serials can still be changed automatically
+                    UpdateProject();
```
Nezz, line 2187: We could add a `TimeoutMessageDialog` here, but I think these "serials" are an implementation detail we don't need to expose to users.

DocMoebiuz: I kind of disagree - if we manipulate a config/profile then we should at least tell the user about it.

Please see my comment on the original issue #2270

If it is really straight forward, well then everything is fine and MobiFlight should just tell the user:

* We have optimized this profile to use all your connected devices * 

Also what about saving the changes? This would pop-up again next time.

DocMoebiuz: What if we can't resolve it? Two Yokes with the same name?

Nezz: >I kind of disagree - if we manipulate a config/profile then we should at least tell the user about it.

It's just not actionable to the user. They probably just downloaded a profile and want it to work. Note that I didn't touch any of the edge-cases here where assignments could be destructive, only the trivial ones.

>Also what about saving the changes? This would pop-up again next time.

There is no popup if everything resolves automatically. Only if there are two devices with the same name or a device that isn't connected.

>What if we can't resolve it? Two Yokes with the same name?

If we can't resolve it then the popup will be displayed as usual.

PR #2280, opened 2025-08-04
ProSim: Keep Alive WebSocket

tristanmcpherson:
Users are seeing issues where inactivity in inputs results in a 5 to 10 second delay when sending a new input after about 5 minutes or so. Implemented a heartbeat timer to keep the websocket alive

### DIFF
--- a/ProSim/ProSimCache.cs
+++ b/ProSim/ProSimCache.cs
@@ -3,6 +3,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


tristanmcpherson:
> Out of curiosity, did you try `_connection.SendPingAsync`? Could that be of any help?

Ping/Pong are only supported by the ws protocol `graphql-transport-ws` which is an older protocol. Ours connects with `graphql-ws` so it is not supported unfortunately.

tristanmcpherson:
This has been reported to work by several ProSim users. I think it's safe to merge if we have no concerns about the code

PR #2281, opened 2025-08-04
Fix React performance issues

Nezz:
Context: https://discord.com/channels/608690978081210392/745361245808492584/1399891466410528848

The new React frontend had performance issues, in particular when using test mode. These were caused by the whole list re-rendering whenever something changed. There was a bit of memoization done already, but it wasn't nearly enough to fix the performance. Fortunately [React Compiler](https://react.dev/learn/react-compiler) was built to do it automatically. To enable it I had to update to React 19, but now test mode is super fast:

https://github.com/user-attachments/assets/50a3b576-db69-4cb1-a218-4330541abec7

Caveat: TanStack Table does not follow the current [Rules of React](https://react.dev/reference/rules/components-and-hooks-must-be-pure#props), so full memoization is not possible currently.


frontend/package.json
DocMoebiuz, line 1: We should also upgrade vite to 

`"vite": "^7.0.6"`

which is used by vite in their official react 19 ts template

DocMoebiuz: https://stackblitz.com/edit/vitejs-vite-w3rrckid?file=package.json&terminal=dev

DocMoebiuz:
```
    "dependencies": {
    "react": "^19.1.1",
    "react-dom": "^19.1.1"
  },
  "devDependencies": {
    "@eslint/js": "^9.32.0",
    "@types/react": "^19.1.9",
    "@types/react-dom": "^19.1.7",
    "@vitejs/plugin-react": "^4.7.0",
    "eslint": "^9.32.0",
    "eslint-plugin-react-hooks": "^5.2.0",
    "eslint-plugin-react-refresh": "^0.4.20",
    "globals": "^16.3.0",
    "typescript": "~5.8.3",
    "typescript-eslint": "^8.39.0",
    "vite": "^7.0.6"
  }
```

DocMoebiuz: "babel-plugin-react-compiler": "^19.1.0-rc.2",
this one is mentioned by the react compiler docs. That should be fine!

DocMoebiuz: "eslint-plugin-react-hooks": "^6.0.0-rc.1",

is this one required? RCs better only if necessary. Unless, it is MobiFlight of course :D

frontend/src/components/tables/config-item-table/config-item-table.tsx
DocMoebiuz, line 1: If we use the "use no memo" directive, should the useCallback hook not actually be used?

It makes sense to remove the hooks if the compiler can do the work... but I understood the directive will prevent the compiler to do its magic.

frontend/src/components/tables/config-item-table/data-table-faceted-filter.tsx
```diff
@@ -37,6 +37,10 @@ export function DataTableFacetedFilter<TData, TValue>({
   title,
   options,
 }: DataTableFacetedFilterProps<TData, TValue>) {
+  // TODO: Fix this to work with React Compiler, changes done to table columns does not trigger a re-render
+  // eslint-disable-next-line react-hooks/react-compiler
+  "use no memo"
+
```
DocMoebiuz, line 43: why do we need this here if there are no components or functions that were memoized?

frontend/src/components/tables/config-item-table/data-table-toolbar.tsx
```diff
@@ -45,6 +45,10 @@ export function DataTableToolbar<TData>({
   onToggleSelected,
   onClearSelected,
 }: DataTableToolbarProps<TData>) {
+  // TODO: Fix this to work with React Compiler, changes done to table columns does not trigger a re-render
+  // eslint-disable-next-line react-hooks/react-compiler
+  "use no memo"
+
```
DocMoebiuz, line 51: same question like in the other file. no memoized components or functions, yet we have to flag this with this directive?

frontend/src/components/tables/config-item-table/items/ConfigItemTableActiveCell.tsx
```diff
@@ -30,16 +29,16 @@ const ConfigItemTableActiveCell = React.memo(({ row }: ConfigItemTableActiveCell
         <Switch
           checked={row.getValue("Active") as boolean}
           onClick={() => {
-            item.Active = !item.Active
+            const updatedItem = { ...item, Active: !item.Active }
```
DocMoebiuz, line 32: good catch!

frontend/package.json
Nezz: >We should also upgrade vite

I tried upgrading to Vite 7.0 and 7.1, but the Playwright tests are failing:
https://github.com/MobiFlight/MobiFlight-Connector/actions/runs/16801814934/job/47584767662?pr=2281

>"eslint-plugin-react-hooks": "^6.0.0-rc.1",
>
>is this one required?

That's not required but highly recommended. It contains linter checks for React Compiler. For example it caught this one:
https://github.com/MobiFlight/MobiFlight-Connector/pull/2281#discussion_r2258305474

frontend/src/components/tables/config-item-table/config-item-table.tsx
Nezz: `useCallback` only has an effect in components that are memoized.

frontend/src/components/tables/config-item-table/data-table-faceted-filter.tsx
Nezz: This component would get memoized automatically by React Compiler, but it ends up never updating without this opt-out.

Under the hood the problem is that the dependencies of the filter are:
- Column (changes, but does not follow Rules of React, therefore is not compatible with auto-memoization)
- Title (doesn't change)
- Options (doesn't change)

frontend/src/components/tables/config-item-table/data-table-toolbar.tsx
Nezz: Same situation as the other one, if you remove this line you'll see that the toolbar doesn't update. The underlying issue is that table self-mutates.

frontend/src/components/tables/config-item-table/config-item-table.tsx
DocMoebiuz: I am not sure if we understood each other.

If we use the "use no memo" directive, no react compiler optimizations are applied.

So why would we remove the useCallback hook which memoizes the callback function based on the parameters, basically  what the compiler would be applying if we hadn't used the "use no memo" directive

We are passing the callback to the DnDcontext which will force a re-render of its sub-components. So useCallback will prevent the re-render as long as the dependency parameters haven't changed.

Nezz: >So useCallback will prevent the re-render as long as the dependency parameters haven't changed.

I did not find any instances where `useCallback` prevented re-renders as some of the dependencies always change, but I just re-added it. In terms of performance I did not see any measurable difference between having them and not having them.

Without useCallback:
without

With useCallback:
with

PR #2296, opened 2025-08-19
Resolve Python versioning validation issue in ScriptRunner

ColinM9991:
This resolves an issue with the version validation logic whereby an incremented major version would cause the check to fail when it compares the minor versions since the minor version of the newer package is less than the minimum required minor version

Example:
Required: 3.8 - Installed: 4.0
4 > 3: true
8 > 0: false
Return false

To resolve this, I've used the `Version` struct that's built into .NET since the overloaded operators will handle the comparison logic.

In order to test this, I opted for a data driven approach which required installing MSTest to take advantage of `DynamicData`. Since both MSTest and the GAC based Unit Testing are developed by Microsoft, the APIs are _very_ similar (with one exception) so it's all backwards compatible. I opted for the MSTest approach as it opens more opportunities for performing data driven unit tests.

The test coverage covers a variety of scenarios:
1. No packages installed
2. Missing various packages
3. Lower major and minor versions
4. Higher major, minor and revision numbers
5. Happy path - all required versions installed

The MSTest Adapter has also been installed so that tests are discoverable and can be run in local development IDEs. I don't expect any manual intervention from other developers but it's worth ensuring the tests still run locally as it may differ on a per-IDE basis (I use Visual Studio 2022 for MobiFlight)

### DIFF
--- a/MobiFlightUnitTests/MobiFlight/BrowserMessages/Incoming/Converter/ConfigItemConverterTests.cs
+++ b/MobiFlightUnitTests/MobiFlight/BrowserMessages/Incoming/Converter/ConfigItemConverterTests.cs
@@ -42,7 +42,7 @@ namespace MobiFlight.BrowserMessages.Incoming.Converter.Tests
         public void ReadJson_UnsupportedType_ThrowsNotSupportedException()
         {
             var json = "{\"Type\":\"UnsupportedConfigItem\"}";
-            Assert.Throws< NotSupportedException>(() => JsonConvert.DeserializeObject<ConfigItem>(json));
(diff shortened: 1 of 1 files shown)


Koseng:
Thanks, very nicely done. From my point of view it can be merged. @DocMoebiuz regarding MSTest, you need to make the call.

MobiFlightUnitTests/MobiFlight/Scripts/ScriptRunnerTests.cs
DocMoebiuz, line 1: That's pretty cool!

PR #2303, opened 2025-08-27
Fix issue setting font and sending display data at the same time

Koseng:
In rare cases transferring the font to the CDU interferes with setting new display data. Setting the font is asyncronous, the websocket does not wait for the operation to complete.

As a first solution a wait of 1 second is introduced to give enough time for transferring the font to the CDU.

For a perfect solution at some point in the future, a two way websocket communication needs to be introduced.

### DIFF
--- a/Scripts/Winwing/fslabs_winwing_cdu.py
+++ b/Scripts/Winwing/fslabs_winwing_cdu.py
@@ -100,7 +100,7 @@ async def run_mobiflight_websocket_client():
                 fontName = "AirbusThales"
                 await mobi_websocket_connection.send(f'{{ "Target": "Font", "Data": "{fontName}" }}')
                 logging.info(f"Setting font: {fontName}")
-
+                await asyncio.sleep(1) # wait a second for font to be set
             await asyncio.sleep(0.05)
 
         except Exception as ex:
--- a/Scripts/Winwing/zibo_737_800x.py
+++ b/Scripts/Winwing/zibo_737_800x.py
@@ -263,6 +263,7 @@ async def get_available_devices() -> list[CduDevice]:
                 )
                 available_devices.append(device)
                 await socket.send(FONT_REQUEST)
+                await asyncio.sleep(1) # wait a second for font to be set
         except websockets.WebSocketException:
             logging.warning(
                 "Attempted to probe CDU device %s at endpoint %s but device wasn't available",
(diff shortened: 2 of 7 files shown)


Scripts/Winwing/fslabs_winwing_cdu.py
```diff
@@ -100,7 +100,7 @@ async def run_mobiflight_websocket_client():
                 fontName = "AirbusThales"
                 await mobi_websocket_connection.send(f'{{ "Target": "Font", "Data": "{fontName}" }}')
                 logging.info(f"Setting font: {fontName}")
-
+                await asyncio.sleep(1) # wait a second for font to be set
             await asyncio.sleep(0.05)
```
ColinM9991, line 104: Duplicate sleep. This could be increased to 1 second and the preceding sleep removed

Scripts/Winwing/zibo_737_800x.py
```diff
@@ -263,6 +263,7 @@ async def get_available_devices() -> list[CduDevice]:
                 )
                 available_devices.append(device)
                 await socket.send(FONT_REQUEST)
+                await asyncio.sleep(1) # wait a second for font to be set
```
ColinM9991, line 266: It doesn't harm to have in the X-Plane scripts but I'm not sure if it's needed since the approach is slightly different. In some MSFS scripts the functions subscribe to updates immediately after sending the font request. For X-Plane, the font request is sent during the initial stages where the CDU devices are being probed for availability. The actual connections between X-Plane and MobiFlight are opened afterwards. This should in theory leave enough time between the last font request and the first received message that the font request won't override the first message

HeikoS-Trumpf: The question is, are you sure time will be enough? We could wait and see whether some issue occurs in the future or not. Or we could just keep it and be more on the safe side. My tendency would be to keep the delay.

Koseng: The question is, are you sure time will be enough? We could wait and see whether some issue occurs in the future or not. Or we could just keep it and be more on the safe side. My tendency would be to keep the delay.

ColinM9991: I don't mind either way. The impact of a delay is very minimal especially since this is in the initialization

Koseng: ok, then let's keep it for now.

PR #2316, opened 2025-09-11
Added Winwing script for the Rotate MD11 in XPlane 12

Brutarul:
Added CDU1,2,3 script for the Rotate MD-11.

### DIFF
--- a/Scripts/Winwing/rotate_md11.py
+++ b/Scripts/Winwing/rotate_md11.py
@@ -1,3 +1,24 @@
+"""
+Adds support for the Rotate MD-11 in X-Plane
+
+Many X-Plane aircraft have similar formats for datarefs and the means of retrieving, translating and sending updates is mostly the same.
+
+In order to support multiple CDU devices seamlessly, a dynamic approach is taken whereby an enum class is defined that contains the supported devices.
+A device is considered "supported" if it exists in the aircraft. Some aircraft have 3 CDUs while others have 2.
+Each enum member is assigned a value that represents the X-Plane dataref identifier. Example: cdu_0 for Rotate/aircraft/controls/cdu_0/mcdu_line.
+
+Upon script start, MobiFlight is probed (get_available_devices()) to detect the devices connected to the PC. Any device that returns a successful response is then tracked.
+
+Two tasks are started independently for each available CDU device.
+1. handle_dataref_updates -> Listens to X-Plane's WebSocket server for dataref updates for that specific CDU and pushes an event to a queue
+2. handle_device_update   -> Listens to the queue and dispatches updates to MobiFlight to update that CDU
+
+Tasks are started independently for each CDU device to ensure each device can update quickly, particularly when players might be performing shared cockpit flights.
+
+Upon a failed connection while dispatching updates to MobiFlight, the handle_device_update function uses `async for` with the websockets client. The failed message is put back in the queue, the loop continues to the next iteration which then reconnects again.
+The failed message is picked back up and dispatched to MobiFlight. This ensures a user's device eventually receives the updated display contents and doesn't hang which would require the user to cycle the page again.
+"""
+
 import asyncio
 import base64
 import json
@@ -68,7 +89,7 @@ def get_color(color: int) -> str:
     return COLOR_MAP.get(color, "w")
 
 
-def fetch_dataref_mapping(device: CduDevice):
+def fetch_dataref_mapping(device: CduDevice) -> dict[int, str]:
     with urllib.request.urlopen(BASE_REST_URL, timeout=5) as response:
         response_json = json.load(response)
 
@@ -81,7 +102,7 @@ def fetch_dataref_mapping(device: CduDevice):
     }
 
 
-def generate_display_json(values: dict[str, str], device: CduDevice):
+def generate_display_json(values: dict[str, str], device: CduDevice) -> str:
     display_data = [[] for _ in range(CDU_CELLS)]
 
     content_lines = [
@@ -131,7 +152,7 @@ async def handle_device_update(queue: asyncio.Queue, device: CduDevice):
 
             except websockets.exceptions.ConnectionClosed:
                 logging.error(
-                    "MobiFlight websocket connection was closed... Attempting to reconnect"
+                    "WinWing CDU websocket connection was closed... Attempting to reconnect"
                 )
                 await queue.put(values)
                 break


Scripts/Winwing/rotate_md11.py
```diff
@@ -0,0 +1,176 @@
...
+BASE_WEBSOCKET_URI = "ws://localhost:8086/api/v2"
+WS_CAPTAIN = "ws://localhost:8320/winwing/cdu-captain"
+
+CHAR_MAP = {35: "\u2610", 42: "\u00b0"}
+COLOR_MAP = {1: "g", 2: "g", 4: "e", 5: "g"}
+
+FONT = "Boeing"
+FONT_REQUEST = json.dumps({"Target": "Font", "Data": FONT})
+
+class CduDevice(StrEnum):
```
ColinM9991, line 53: This isn't currently using the enum approach in the right way. The enum approach allows for multi-CDU support by mapping each enumeration to a specific aircraft CDU. The implemented approach treats the aircraft as the CDU and then hardcodes the CDU to `cdu_0`

Check out one of the FlightFactor scripts for an example on how the enum should be used to dynamically support multiple CDUs

Scripts/Winwing/rotate_md11.py
```diff
@@ -0,0 +1,176 @@
...
+    RotateMD11 = "rotate"
+
+    def get_symbol_datarefs(self) -> list[str]:
```
ColinM9991: Both datarefs could be split out to separate functions such as `get_content_datarefs` and `get_style_datarefs`

Scripts/Winwing/rotate_md11.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def get_color(color: int) -> str:
+    return COLOR_MAP.get(color, "w")
+
+
+def fetch_dataref_mapping(device: CduDevice):
+    with urllib.request.urlopen(BASE_REST_URL, timeout=5) as response:
+        response_json = json.load(response)
+        target_names = device.get_symbol_datarefs()
```
ColinM9991: It may be more performance friendly to instead check that the dataref starts with `Rotate/aircraft/controls/{cduDevice}/mcdu_line` since that'll pull back both style and content datarefs.

To extract style and content values, we could separately pull out values where the dataref name ends with `_style` or `_content`

Scripts/Winwing/rotate_md11.py
```diff
@@ -0,0 +1,176 @@
...
+        except (websockets.exceptions.ConnectionClosed, ConnectionResetError):
+            await asyncio.sleep(reconnect_delay)
+            if 'values' in locals():
+                await queue.put(values)
+            continue
+
+
+async def main():
+    queue = asyncio.Queue()
+    device = CduDevice.RotateMD11
```
ColinM9991: See a FlightFactor script for the approach to multi-CDU support

Scripts/Winwing/rotate_md11.py
```diff
@@ -0,0 +1,176 @@
...
+                    if new_values != last_known_values:
+                        last_known_values = new_values
+                        await queue.put(new_values)
+
+        except (websockets.exceptions.ConnectionClosed, ConnectionResetError):
+            await asyncio.sleep(reconnect_delay)
+            continue
+
+
+async def handle_device_output(queue: asyncio.Queue):
```
ColinM9991: The requirement for debouncing was debunked as the main cause of the CPU-bottleneck of an X-Plane limitation with the WebSocket server being bottlenecked.

For consistency, it might be worth using the same approach as in other scripts since the eventual aim is to create a library to abstract most logic away while preventing duplication of code in many different scripts

Scripts/Winwing/rotate_md11.py
```diff
@@ -0,0 +1,176 @@
...
+            int(dr["id"]): dr["name"]
+            for dr in response_json["data"]
+            if dr["name"] in target_names
+        }
+
+
+def generate_display_json(values: dict[str, str]):
+    display_data = [[] for _ in range(CDU_COLUMNS * CDU_ROWS)]
+
+    content_lines = [
```
ColinM9991, line 108: Dataref naming duplication here and not multi-CDU device compatible

Scripts/Winwing/rotate_md11.py
```diff
@@ -0,0 +1,176 @@
...
+        values.get(f"Rotate/aircraft/controls/cdu_0/mcdu_line_{i}_content", "").ljust(24)
+        for i in range(CDU_ROWS)
+    ]
+    style_lines = [
+        values.get(f"Rotate/aircraft/controls/cdu_0/mcdu_line_{i}_style", [1] * 24)
```
ColinM9991: It looks like the style datarefs use the standard Laminar format documented here: https://developer.x-plane.com/article/datarefs-for-the-cdu-screen/#Formatting_info

We might need to do some bit-shifting to extract the relevant bits of information

Scripts/ScriptMappings.json
```diff
@@ -101,6 +101,12 @@
       "ProductIds": [ "WinwingCDUs" ],
       "AircraftIdSnippet": "boeing 737-",
       "ScriptName": "zibo_737_800x.py"
+    },
+    {
+      "VendorId": "0x4098",
+      "ProductIds": [ "WinwingCDUs" ],
+      "AircraftIdSnippet": "rotate",
```
ColinM9991: It's worth testing compatibility across the MD-11 and MD-80 since this script will run for both

Scripts/Winwing/rotate_md11.py
```diff
@@ -0,0 +1,176 @@
...
+COLOR_MAP = {1: "g", 2: "g", 4: "e", 5: "g"}
+
+FONT = "Boeing"
+FONT_REQUEST = json.dumps({"Target": "Font", "Data": FONT})
+
+class CduDevice(StrEnum):
+    RotateMD11 = "rotate"
+
+    def get_symbol_datarefs(self) -> list[str]:
+        base = "Rotate/aircraft/controls/cdu_0"
```
ColinM9991: `base` could be its own function as this could be used in `fetch_dataref_mapping` to pull back datarefs that start with the base of `Rotate/aircraft/controls/{cduDevice}/mcdu_line `

Brutarul: Thanks for all the comments, I will modify the script in the following days.

Scripts/Winwing/rotate_md11.py
```diff
@@ -0,0 +1,235 @@
...
+def get_color(color: int) -> str:
+    return COLOR_MAP.get(color, "w")
+
+
+def fetch_dataref_mapping(device: CduDevice):
+    with urllib.request.urlopen(BASE_REST_URL, timeout=5) as response:
+        response_json = json.load(response)
+
+    base_prefix = device.get_dataref_base()
+    target_names = set(device.get_symbol_datarefs())
```
ColinM9991: `target_names` is no longer required since we're filtering datarefs based on the base-name

Scripts/Winwing/rotate_md11.py
```diff
@@ -0,0 +1,249 @@
...
+            case CduDevice.Observer:
+                return WS_OBSERVER
+            case _:
+                raise KeyError(f"Invalid device specified {self}")
+
+    def get_dataref_base(self) -> str:
+        return f"Rotate/aircraft/controls/{self}/mcdu_line"
+
+    def get_content_dataref(self, i: int) -> str:
+        return f"Rotate/aircraft/controls/{self}/mcdu_line_{i}_content"
```
ColinM9991, line 73: You could reuse `get_dataref_base` here and the next function. This is a minor suggestion though and not mandatory

ColinM9991 (approved):
Great work mate. Thanks for persisting with this one - it'll be a great addition to the collection of X-Plane aircraft

@DocMoebiuz I'm happy with these changes

Koseng:
@Brutarul , @ColinM9991  there is one more thing before merging. The entry to the csproj is missing, so that the new python file will be installed. See for example here: https://github.com/MobiFlight/MobiFlight-Connector/pull/2246

@Brutarul please create PR with the missing entry to the csproj file.

PR #2323, opened 2025-09-12
Prevent exception by ensuring thread-safe access to updatedValues dictionary

### LINKED ISSUE #2324: MobiFlight throws Exception because of concurrent access to Collection
**Describe the bug**
```
See the end of this message for details on invoking 
just-in-time (JIT) debugging instead of this dialog box.

************** Exception Text **************
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.ValueCollection.Enumerator.MoveNext()
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at MobiFlight.ExecutionManager.FrontendUpdateTimer_Execute(Object sender, EventArgs e)
   at System.Windows.Forms.Timer.OnTick(EventArgs e)
   at System.Windows.Forms.Timer.TimerNativeWindow.WndProc(Message& m)
   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
```

**Expected behavior**
MobIFlight should be robust for concurrent access to the `updatedValues` dictionary

**Actual behavior**
It can throw an exception

**Details**
 - MobiFlight Version: 10.5.3.x (BETA)

**Additional context**
[Discord post](https://discord.com/channels/608690978081210392/635953880122785824/1411604295639371868)

Leonscape:
The `updatedValues` is now using `ConcurrentDictionary` and we don't have to worry about access from different threads anymore.

- [x] Use ConcurrentDictionary
- [x] Update unit-tests

fixes #2324

### DIFF
--- a/MobiFlightUnitTests/MobiFlight/ExecutionManagerTests.cs
+++ b/MobiFlightUnitTests/MobiFlight/ExecutionManagerTests.cs
@@ -516,7 +516,7 @@ namespace MobiFlight.Tests
             // Get references to private members using reflection
             var updatedValuesField = typeof(ExecutionManager).GetField("updatedValues",
                 BindingFlags.NonPublic | BindingFlags.Instance);
-            var updatedValues = (Dictionary<string, IConfigItem>)updatedValuesField.GetValue(_executionManager);
+            var updatedValues = (ConcurrentDictionary<string, IConfigItem>)updatedValuesField.GetValue(_executionManager);
 
             var frontendUpdateMethod = typeof(ExecutionManager).GetMethod("FrontendUpdateTimer_Execute",
                 BindingFlags.NonPublic | BindingFlags.Instance);
@@ -670,7 +670,7 @@ namespace MobiFlight.Tests
             // Get access to the updatedValues dictionary via reflection
             var updatedValuesField = typeof(ExecutionManager).GetField("updatedValues",
                 BindingFlags.NonPublic | BindingFlags.Instance);
-            var updatedValues = (Dictionary<string, IConfigItem>)updatedValuesField.GetValue(_executionManager);
+            var updatedValues = (ConcurrentDictionary<string, IConfigItem>)updatedValuesField.GetValue(_executionManager);
             
             var initialUpdatedValuesCount = updatedValues.Count;
 


MobiFlight/Execution/ConfigItemExecutor.cs
```diff
@@ -71,7 +72,7 @@ public void Execute(OutputConfigItem cfg, Dictionary<string, IConfigItem> update
                 cfg.Status[ConfigItemStatusType.Test] = "TESTING";
                 if (!cfg.Equals(originalCfg))
                 {
-                    updatedValues[cfg.GUID] = cfg;
+                    updatedValues.AddOrUpdate(cfg.GUID, cfg, (k,o) => cfg);
```
DocMoebiuz: What's the reason that you decided to use AddOrUpdate instead of the index setter access?

MobiFlight/ExecutionManager.cs
```diff
@@ -249,12 +250,14 @@ private void FrontendUpdateTimer_Execute(object sender, EventArgs e)
...
-            lock (updatedValues)
+            foreach (var kvp in updatedValues)
             {
-                list = updatedValues.Values.Select(cfg => new ConfigValueOnlyItem(cfg)).Cast<IConfigValueOnlyItem>().ToList();
-                updatedValues.Clear();
+                if (updatedValues.TryRemove(kvp.Key, out var value))
+                {
+                    list.Add(new ConfigValueOnlyItem(value));
+                }
             }
```
DocMoebiuz: I would have expected the following works:

```
var list = updatedValues.Values.Select(cfg => new ConfigValueOnlyItem(cfg)).Cast<IConfigValueOnlyItem>().ToList();
updatedValues.Clear();
````

File: MobiFlight/ExecutionManager.cs [hunk 3]
```diff
@@ -1207,7 +1210,7 @@ void mobiFlightCache_OnButtonPressed(object sender, InputEventArgs e)
                 }
             }
 
-            updatedInputValues.Keys.ToList().ForEach(k => this.updatedValues[k] = updatedInputValues[k]);
+            updatedInputValues.Keys.ToList().ForEach(k => updatedValues.AddOrUpdate(k, updatedInputValues[k], (ck,o) => updatedInputValues[ck]));
```
DocMoebiuz: the original style should still work without any modifications.

MobiFlight/ExecutionManager.cs
```diff
@@ -1240,7 +1243,7 @@ private void UpdateInputPreconditions()
 
                         if (!cfg.Status.SequenceEqual(originalCfg.Status))
                         {
-                            updatedValues[cfg.GUID] = cfg;
+                            updatedValues.AddOrUpdate(cfg.GUID, cfg, (k, o) => cfg);
```
DocMoebiuz: the original way would just work, in my opinion the benefit is: it is simple and clear to read, no lambda.

MobiFlight/Execution/ConfigItemExecutor.cs
DocMoebiuz: Same for all the subsequent changes.
I would prefer the old direct way which is easy to read.

MobiFlight/ExecutionManager.cs
Leonscape: updateValues.Values is a snapshot of the values of the dictionary which is then iterated by the select.
Any added items after the snapshot would not be processed, but would then be removed by the Clear.

So by doing it through the remove we know the item is added to list to be processed. We may leave some values to be processed on the next call.

Theres also wrapping in a count > 0, to guarentee emptying the dictionary, but if lots of things are being added that can cause a delay. or even infinite looping (had that) so a limit on the outer loop is also really needed.

MobiFlight/ExecutionManager.cs
Leonscape: This is part of my current work place policy, So concurrent setting is always AddOrUpdate. Just me doing the automatically, its not really necessary here. i've changed it back.

MobiFlight/ExecutionManager.cs
Leonscape: See previous comment changed them back.

MobiFlight/Execution/ConfigItemExecutor.cs
Leonscape: Changed it back got a little carried away, Its required where I work for ConcurrentDictionaries, I think mainly to show its a concurrent. I've reverted it.

MobiFlight/ExecutionManager.cs
DocMoebiuz: Then let us use the best of both worlds

```
// now we can itereate over the it and remove the keys safely
updatedValues.Keys.ToList().ForEach(key => {
  if (!updatedValues.TryRemove(key, out var value)) return;
  list.Add(new ConfigValueOnlyItem(value))
})

ToList() really creates a copy of the keys and would have its own stable enumerator.

Comment (DocMoebiuz) [reply MobiFlight/ExecutionManager.cs]: I always try to implement early returns instead of nested blocks.

Comment (Leonscape) [reply MobiFlight/ExecutionManager.cs]: Made the change.

File: MobiFlight/ExecutionManager.cs [hunk 5]
```diff
@@ -248,13 +249,15 @@ private void FrontendUpdateTimer_Execute(object sender, EventArgs e)
...
+            updatedValues.Keys.ToList().ForEach(key =>
             {
-                list = updatedValues.Values.Select(cfg => new ConfigValueOnlyItem(cfg)).Cast<IConfigValueOnlyItem>().ToList();
-                updatedValues.Clear();
-            }
+                if(!updatedValues.TryRemove(key, out var value))
+                    return;
+
+                list.Add(new ConfigValueOnlyItem(value));
+            });
```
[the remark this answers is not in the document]
Leonscape: Not a race the behaviour we want.

PR #2335, opened 2025-09-20
Controller definitions are organized in sub-folders

### LINKED ISSUE #2332: Allow nested folders for joystick definition files
Joystick definition files are all flat in one folder. This leads also to very long json file names because they all have to have a prefix.

Joystick definition files can be organized into folders.
At least grouped by vendor which would mean one additional level.

Keep it flat. NO!

**Additional context**
Do we need Community/Joysticks too?
We have grouping for MobiFlight based controllers on vendor level. So this would be similar to that.

DocMoebiuz:
Joystick definition files are no organized by sub-folders. Definition files are grouped by vendor. Existing MobiFlight installations will perform a migration and clean up of old files.

This should also apply to Midi Controller Definition files.

image

- [x] All existing joystick definition files are organized into sub-folders grouped by "vendor"
- [x] For existing installations, definition files are cleaned up or backuped:
  - [x] files that exists in the root-folder and in one of the vendor sub-folders, and are identical will be deleted
  - [x] files that exists in the root-folder and in one of the vendor sub-folders, and are NOT identical will be saved in "conflicted" folder
  - [x] files that ONLY exist in the root folder are preserved in "MIGRATED" folder, in case a user created their own profile which does not ship with MobiFlight.
- [x] Unit tests exist

fixes #2332

### DIFF
--- a/MobiFlightUnitTests/MobiFlight/Joysticks/JoystickDefinitionMigratorTests.cs
+++ b/MobiFlightUnitTests/MobiFlight/Joysticks/JoystickDefinitionMigratorTests.cs
@@ -42,7 +42,7 @@ namespace MobiFlight.Joysticks.Tests
             File.WriteAllText(uniqueFile, uniqueContent);
 
             // Act
-            JoystickDefinitionMigrator.MigrateDefinitions(_joysticksDirectory);
+            ControllerDefinitionMigrator.MigrateDefinitions(_joysticksDirectory, "*.joystick.json");
 
             // Assert
             Assert.IsFalse(File.Exists(uniqueFile), "Original file should be moved");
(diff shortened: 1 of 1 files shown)


Joysticks/s_tecs/s_tecs_modern_throttle_max.joystick.json
cbrauers, line 1: Separate folder from the other VKB stuff? Is that because of prefixes (STECS does not have VKBsim in the product name for some reason)?

DocMoebiuz: That happened only because of my ignorance :D - i just went by prefix.
We could move those now into a dedicated sub-folder inside of VKBsim

For now, I have simply moved them into the VKBsim

PR #2372, opened 2025-10-18
Bugfix - Set LED status via boolean or numerical value

ColinM9991:
#2353 adjusted how LED values are parsed. Some can be integers/doubles and others as booleans (True/False)

To preserve backwards compatibility with profiles shared on X-Plane.to or FlightSim.to, this parse function will handle numerical values and booleans, otherwise it'll return 0 as default.

There's an open design decision here as to whether this should be a breaking change. On one hand it makes sense to support a boolean indicator as it does state whether the LED should be on or off.

On the other, it opens up edge cases. Nonetheless, the `TryParse` methods are generally preferred over `Convert.To`

### DIFF
--- a/MobiFlight/Execution/ConfigItemExecutor.cs
+++ b/MobiFlight/Execution/ConfigItemExecutor.cs
@@ -9,8 +9,8 @@ using MobiFlight.xplane;
 using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
+using System.Globalization;
 using System.Linq;
-using System.Windows.Forms;
 
 
 namespace MobiFlight.Execution
@@ -583,12 +583,12 @@ namespace MobiFlight.Execution
             return result;
         }
 
-        private static int ParseValue(string value)
+        internal static int ParseValue(string value)
         {
             double doubleState = 0;
             if (value != "0")
             {
-                if (!double.TryParse(value, out doubleState))
+                if (!double.TryParse(value, NumberStyles.Number, CultureInfo.InvariantCulture, out doubleState))
                 {
                     doubleState = 1;
                 }
--- a/MobiFlightUnitTests/MobiFlight/Execution/ConfigItemExecutorTests.cs
+++ b/MobiFlightUnitTests/MobiFlight/Execution/ConfigItemExecutorTests.cs
@@ -299,5 +299,29 @@ namespace MobiFlight.Tests
                 Assert.AreEqual("Config reference not found", e.Message);
             }
         }
+
+        [TestMethod]
+        [DataRow("something", 1)]
+        [DataRow("0", 0)]
+        [DataRow("0.0", 0)]
+        [DataRow("-3", -3)]
+        [DataRow("-3.5", -3)]
+        [DataRow("22", 22)]
+        [DataRow("4.22223", 4)]
+        [DataRow("-500.8", -500)]
+        [DataRow("40,8", 408)]
+        [DataRow("10000.00", 10000)]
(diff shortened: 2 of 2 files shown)


ColinM9991:
Closing following Discord discussion. The preferred approach is to raise a more meaningful error message rather than continuing to support default values

ColinM9991:
> Let's do it 100% compatible as it was before. The behavior before was not to handle "True" and "False" correctly.
> 
> In the past it was: If it is not "0" then set it to 1. No matter what was the input. Every strange string defaulted to 1.
> 
> Then Sebastian extendend it to: If i is not "0", then try to parse into a byte, if it fails default to 1. No matter what was the input. My issue with that was, that numbers like "29.0" then default to 1 in the future.
> 
> We should do now: If it is not "0", then try to parse into a double, if it fails default to 1. No matter what was the input. Then convert the double to an int and then to a byte.
> 
> So, can you change it in your PR just to:
> 
> ```
>     default: // LED Output
>         double doubleState = 0;
>         if (value != "0")
>         {
>             if (!Double.TryParse(value, out doubleState))
>             {
>                 doubleState = 1;
>             }
>         }                       
>         byte state = (byte)Math.Max(Math.Min(255, (int)doubleState), 0);
>         joystick.SetOutputDeviceState((cfg.Device as Output).DisplayPin, state);
>         joystick.UpdateOutputDeviceStates();
>         joystick.Update();
>         break;
> ```

Implemented logic within the `ParseValue` function. Mainly to isolate the logic into a self-contained/readable function

DocMoebiuz:
First of all, we don't want breaking changes.
However, I don't want to have all this backward-compatibility code in the code base.

Alternative:
Instead of making the parsing code more complicated and cover all legacy cases, we could also migrate the JSON file to the correct structure. I am currently implementing https://github.com/MobiFlight/MobiFlight-Connector/pull/2370 and this scenario could actually be a use case for a JSON migration on load rather than having to deal with it downstream in the code base.

Koseng:
Or here my tests to integrate in `ConfigItemExecutorTests.cs'

        using System.Reflection;

        private void ParseValue_UsingReflection_CompareInputAndExpected(MethodInfo method, string input, int expected)
        {
            var result = method.Invoke(null, new object[] { input });
            Assert.AreEqual(expected, (int)result);
        }

        [TestMethod]
        public void ExecuteDisplay_ParseValue_ParseDifferentInputStringValues()
        {
            // Arrange
            var method = typeof(ConfigItemExecutor)
                .GetMethod("ParseValue", BindingFlags.NonPublic | BindingFlags.Static);

            Assert.IsNotNull(method); // Ensure method was found

            // Act
            ParseValue_UsingReflection_CompareInputAndExpected(method, "something", 1);
            ParseValue_UsingReflection_CompareInputAndExpected(method, "0", 0);
            ParseValue_UsingReflection_CompareInputAndExpected(method, "0.0", 0);
            ParseValue_UsingReflection_CompareInputAndExpected(method, "-3", -3);
            ParseValue_UsingReflection_CompareInputAndExpected(method, "-3.5", -3);
            ParseValue_UsingReflection_CompareInputAndExpected(method, "22", 22);
            ParseValue_UsingReflection_CompareInputAndExpected(method, "4.22223", 4);
            ParseValue_UsingReflection_CompareInputAndExpected(method, "-500.8", -500);
            ParseValue_UsingReflection_CompareInputAndExpected(method, "40,8", 408);
            ParseValue_UsingReflection_CompareInputAndExpected(method, "10000.00", 10000);            
        }

DocMoebiuz (approved):
Great! Thanks 🙏

PR #2414, opened 2025-11-21
Add Honeycomb Sierra support to MobiFlight

UntoldForce:
The Honeycomb Sierra TPM needs a joystick definition file and one minor code change to function properly in MobiFlight. Without the minor code change, the Sierra joystick would crash if any LED outputs were defined in the joystick definition file. With this change, all Sierra buttons and LEDs are functional in MobiFlight. Tested with the Honeycomb Sierra and the Honeycomb Bravo connected at the same time to verify no loss in functionality.

### DIFF
--- a/MobiFlight/Joysticks/Joystick.cs
+++ b/MobiFlight/Joysticks/Joystick.cs
@@ -532,6 +532,10 @@ namespace MobiFlight
 
         public virtual void UpdateOutputDeviceStates()
         {
+            // Honeycomb LED protocol: Feature report with bits set for each lit LED and cleared for each unlit LED.
+            // Byte 0 is unused on Bravo, but used as the report ID on Sierra. Report ID for the LED command must be 1 on Sierra.
+            // Setting this byte to 1 is supported by both throttles.
+            // Other bytes initialized to zero to ensure lights are only turned on if they are set to 1 by output events.
             var data = new byte[] { 1, 0, 0, 0, 0 };
 
             foreach (var light in Lights)


cbrauers:
Explanation of the code change: Versus the older Bravo, the Sierra added report IDs and multiple reports to the HID descriptor. The main report with all the DirectInput-related controller data and the LED controls has ID 1. The zero-initialized array introduced for the Bravo did not have a corresponding report ID.

From what we can tell, byte 0 of the report is unused, so initializing it with the report ID fixes Sierra without breaking Bravo compatibility. Medium-term, a more elegant solution might be to extract the Honeycomb LED logic into its own subclass and only set the report ID for newer devices (e.g. Sierra), but for now this allows both Honeycomb throttles to have full LED control.

MobiFlight/Joysticks/Joystick.cs
```diff
@@ -532,7 +532,7 @@ protected virtual void SendData(byte[] data)
 
         public virtual void UpdateOutputDeviceStates()
         {
-            var data = new byte[] { 0, 0, 0, 0, 0 };
+            var data = new byte[] { 1, 0, 0, 0, 0 };
```
neilenns, line 539: It'd be good to add a code comment here that echoes the explanation in the PR comments, detailing why on earth a random 1 gets put in this array.

cbrauers: New version should be clear on that.

neilenns: Perfect!

PR #2451, opened 2025-12-11
Fixed missing keyboard inputs in project settings dialog

### LINKED ISSUE #2448: creating new project from within "edit view" works strange
**Describe the bug**
Adding new project from My Projects view works ok. Adding new project from within another project (the config list view) behaves differently - you can do it, but for example you cannot use spaces in the project name field.

**Steps To Reproduce**
1. open mfproj, in edit view
2. `File -> New`
3. type project name
4. observe you cannot add whitespace in the project name field. IF you open the same dialog from the "My Projects" view you can, so something is inconsistent.

**Expected behavior**
should work the same regardless where new project is being added from.

DocMoebiuz:
fixes #2448

### DIFF
--- a/frontend/src/components/project/ProjectForm.tsx
+++ b/frontend/src/components/project/ProjectForm.tsx
@@ -42,24 +42,34 @@ const ProjectForm = ({
 
   const handleSubmit = (e: React.FormEvent) => {
     e.preventDefault()
-    if (name.trim() === "") {
+    const trimmedName = name.trim()
+
+    if (trimmedName === "") {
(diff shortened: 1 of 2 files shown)


tigert (approved):
Not code-fu skilled enough to comment on the diff itself, but this seems to fix the problem.

Of course the question is, whether we should even have save as etc in the menubar now that we have the projects dashboard. Because if you are editing an existing config, it is a bit strange flow to add a new project while editing an existing project. But this is outside scope here, this PR fixes the bug.

PR #2456, opened 2025-12-13
Tabs can scroll horizontally when not enough space available

### LINKED ISSUE #2450: Tabs dont overflow correctly with insufficient space
**Describe the bug**

**Steps To Reproduce**
1. Go to the config list view
2. Add many tabs so that they exceed the available space

**Expected behavior**
The tabs should become scrollable, or stay accessible in some way.

**Actual behavior**
The tabs grow beyond the visible screen and also break the layout of the config list table

**Details**
 - MobiFlight Version: 11.3.15

**Additional context**
#2425

DocMoebiuz:
https://github.com/user-attachments/assets/dec07237-d018-4a1b-9fcd-ab52c3f4f3c1

Tabs can now be scrolled horizontally if they require more than the available, visible space.
The active tab will scroll into view when the mouse leaves the tabs area.

- [x] tabs scroll beyond the available screen space
- [x] ~~i18n (not needed)~~
- [x] frontend tests

fixes #2450


joeherwig:
Thanks a lot! Great fix!

PR #2476, opened 2025-12-16
ProSim: Fix GraphQL boolean by using GraphQL variables

tristanmcpherson:
Fixes ProSim boolean write operations by using GraphQL variables instead of string interpolation for mutations.

When writing boolean values to ProSim, the values were being interpolated directly into the GraphQL query string. C# outputs `True`/`False` when converting booleans to strings, but GraphQL requires lowercase `true`/`false` . This caused boolean writes to fail.

Refactored the mutation to use GraphQL variables, letting the GraphQL library handle proper serialization of all value types

### DIFF
--- a/MobiFlightUnitTests/ProSim/ProSimCacheTests.cs
+++ b/MobiFlightUnitTests/ProSim/ProSimCacheTests.cs
@@ -99,16 +99,16 @@ namespace MobiFlight.Tests.ProSim
         {
             // Verify mutation lookup matches what's in ProSimCache
             // This is critical - wrong mutation = data won't be written correctly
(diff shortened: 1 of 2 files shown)


DocMoebiuz:
I will merge... you are addressing booleans in this PR

tristanmcpherson:
> I will merge... you are addressing booleans in this PR

It probably could be improved, I did mean to open this as a draft at first, but most likely this is still good to go. I do agree with copilot that Thread.Sleep is brittle, but it should be relatively safe. I will try to keep it in mind and probably refactor to a degree since there is a likely a better way to test this than a unit test that involves some non-async waiting

PR #2516, opened 2025-12-21
Reduce logging of VKB encoder events.

### LINKED ISSUE #2515: Logs getting spammed when large number of VKB devices are present
**Describe the bug**
Missed encoder messages on VKB devices are logged at debug level. Since they are sent out at a potentially rather high rate on a busy system, individual events are lost occasionally, which compounds as the user connects more and more devices. This causes unnecessary amounts of data to end up in the log file

**Steps To Reproduce**
- Connect many VKB devices
- Set logging to debug
- Observe the log

**Expected behavior**
Logs about missed encoder messages rarely appear

**Actual behavior**
A log entry about missing encoder messages appears every few seconds.

**Details**
 - MobiFlight Version: 10.5.3.16
 - FlightSim: irrelevant, appears without flight sim running
 - AddOn: irrelevant

**Additional context**
Since there is no data loss due to skipped messages (message contains absolute position), short interruptions can be ignored. Longer interruptions are recoverable, but it makes sense to monitor them for frequency - if it still occurs often, there may be connectivity issues.

cbrauers:
Fixes #2515

### DIFF
--- a/MobiFlight/Joysticks/VKB/VKBDevice.cs
+++ b/MobiFlight/Joysticks/VKB/VKBDevice.cs
@@ -8,6 +8,7 @@ namespace MobiFlight.Joysticks.VKB
 {
     internal class VKBDevice : Joystick
     {
+        const int ENCODER_MISSED_MESSAGE_THRESHOLD = 5; // How many missed encoder messages are allowed before we need to log them - chosen arbitrarily
         private HidStream Stream;
         private readonly HidDevice Device;
         private readonly new VKBLedContainer Lights = new VKBLedContainer();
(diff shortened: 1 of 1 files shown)


MobiFlight/Joysticks/VKB/VKBDevice.cs
```diff
@@ -167,10 +167,12 @@ private void ParseEncoderReport(byte[] Report)
             byte sequenceNo = Report[2];
             // Sequence number should increment once per report, but if the encoder is spun fast some reports can be missed.
             // Firmware is capable of transmitting up to 250 updates per second.
-            if (((lastSeqNo + 1) & 0xFF) != sequenceNo)
+            if (lastSeqNo != -1 && ((sequenceNo - lastSeqNo) & 0xFF) > 5) // Only log large skips, do not log first message.
```
DocMoebiuz: i started to write these tests a bit different to reduce the amount of required comments.

i typically would create a variable with a sensible name that hints to the meaning
e.g.

```
var isFirstMessage = lastSqNo == -1;
var hasMultipleMissedUpdates == ((sequenceNo - lastSeqNo) & 0xFF) > 5
```

then your test is
```
if (!istFirstMessage && hasMultipleMissedUpdates) {
...
}
```

DocMoebiuz: and magic numbers like the `5` are good candidates to be put at the top of the class as

`const int MULTIPE_MISSED_UPDATES_THRESHOLD = 5`

the name might not the best but you know what I mean.

Issue #2209, opened 2025-06-22
Dataref used in encoder input is not initially read


JaimeLeon2:
**Describe the bug**
Reported by Discord user ChezHJ - TIX0355
[](https://discord.com/channels/608690978081210392/1382438399956222163)
Xplane Dataref used in an encoder input config does not read the initial value correctly.

![Image](https://github.com/user-attachments/assets/3d354e2e-9b2a-453a-acbd-afbd051cd8ff)

![Image](https://github.com/user-attachments/assets/ad22adc7-a252-48d4-a4b2-39b1b12a6d8b)

**Steps To Reproduce**
1. Create input config with an encoder
2. Bind a dataref
3. Insert fast right value $+10
4. Turn encoder right fast
5. See that initial value is ignored and value is set starting at +10
6. Reported the formula works correctly on successive clicks of the encoder, but with the wrong initial value.

**Expected behavior**
Whatever initial value is on the dataref, should be used to replace $ in the formula on the first click.

**Actual behavior**
The value on the first click is apparently set to 0 regardless of the actual value in the dataref.

**Details**
 - MobiFlight Version: 10.5.3
 - FlightSim: Xplane
 - AddOn: Zibo

**Additional context**
None

Issue #1931, opened 2025-01-30
Xplane Datarefs as config items require start/stop to populate


aesellars:
Xplane Source Datarefs as config items require start/stop to populate. if an output is configured with an Xplane dataref value as it's source, and mobiflight is set to autorun, the values do not load and show 0 as output value. Starting and stopping the run is required to get them to read and populate

**Steps To Reproduce**
Load Mobiflight with the following output config:
Sim Variable -
  X-Plane Dataref
  Customized or preset code
    sim/aircraft/controls/acf_RSC_redline_prp[0]
Enabled autorun
Save and Close Mobiflight
Open X-plane
Begin a new flight with the Van's RV-12
Sim running, and engine running
Open Mobiflight
Navigate to the output tab
Note the Flight Sim Value and Output Value are 0 (should read the dataref value but doesn't)
Press Stop
Press Start
Not the correct Flight Sim Value and output Value are now displayed
screenshots attached

**Expected behavior**
Mobflight would load and populate the sim dataref values on startup/autorun, when the sim is already running and aircraft is loaded.

**Actual behavior**
Mobflight is not loading/reading the values on startup/autorun and requires a stop/start to populate them

**Details**
 - MobiFlight Version: 10.5.2 build 20250116
 - FlightSim: Xplane 12.1.3-r2 build 121313 Intel 64-bit
 - AddOn: Default Vans RV-10

![Image](https://github.com/user-attachments/assets/46a2d3e5-7411-4181-9620-e31bd3ef2b34)
![Image](https://github.com/user-attachments/assets/733c8121-5d66-4a86-bb0c-8f8f21ad8dbc)

JaimeLeon2:
At least three other people have reported the same problem.

tigert:
My friend has also faced this - needs stop and start to make his configs on Toliss work. He says Spad.NeXT works but it clearly waits maybe 6 seconds after X-Plane has finished loading, until it connects and starts doing its thing.

Maybe the X-Plane network interface needs a short delay before connecting to ensure the startup is actually finished?

MikeT0866:
Sound great…thankyou……how do I get this fix……is it a separate update via GitHub or do I download the latest Mobiflight from web to get the fix???

 

Regards

 

Mike Tavcar

DocMoebiuz:
@MikeT0866 we will release it soon, as beta first and then as regular release

tigert:
[...]

PR #2534, opened 2025-12-23
Stop subscribing to aircraft name dataref when closing xplanecache (#2533)

### LINKED ISSUE #2533: Stop subscribing to aircraft name dataref when stopping XplaneCache
[the problematic line of code](https://github.com/MobiFlight/MobiFlight-Connector/blob/main/xplane/XplaneCache.cs#L117)

**Steps To Reproduce**
1. Open XP12 (any airport, any aircraft) , and Mobiflight Connector after xp12 finishes loading
2. Close Mobiflight connector
3. Observe RREF traffic transmitted from XP12, which is originally targetted at Mobiflight, getting lost in wireshark

**Expected behavior**
Do not subscribe to any more dataref when closing XplaneCache

**Actual behavior**
Subscribe to aircraft name dataref when stopping XplaneCache by executing `XplaneCache.stop()`, which contains `CheckForAircraftName()` in its definition.

**Details**
 - MobiFlight Version: [10.5.3 - I built one myself from the latest github code]
 - FlightSim: XP12

**Additional context**
None.

sw08:
fixes #2533

### DIFF
--- a/xplane/XplaneCache.cs
+++ b/xplane/XplaneCache.cs
@@ -114,7 +114,6 @@ namespace MobiFlight.xplane
         public void Stop()
         {
             Connector?.Stop();
-            CheckForAircraftName();
         }
 
         public void Clear()


DocMoebiuz:
Ok, so we identified a couple of issues here:
* poor naming
* unclear behavior
* a bug with the profile auto-switching

I think instead of fixing parts... we should really focus on the original reason why this code was introduced: the auto-profile switching.

Every simulator cache interface shall provide a stable notification of aircraft change.

Right now it doesn't work correctly.

I would highly appreciate your support on this. and i do think it is possible to achieve, even though the XPlaneConnector library might have some flaws.

Potentially related issues:
#1931 
#2209

sw08:
I'm not planning to make any changes at this moment

PR #2536, opened 2025-12-24
Make donation button solid in dark mode for consistency

tigert:
image

Maintain consistent visual style from light mode, but just muted brightness like other elements.
Also avoid stroked outlines and favor filled shapes for less visual clutter, this is easier to the eyes and hopefully also easier flow to guide the user. I am of course pretty much making the reasoning up but I feel it has some truth in it. :-)


DocMoebiuz:
What does add more zen mean? Do you refer to less contrast, calmer?

PR #2541, opened 2025-12-25
display selected simulator name below the thumbnails

tigert:
Check that I did this correctly, as I am not really very familiar with everything yet. It seems to work though.

As the thumbnails are kinda similar looking, and especially FSX and MSFS both have a fairly similar logo, I think it is good to also display the selected simulator name below the choices.

image

image

image


DocMoebiuz:
I like it. Please add a playwright test to verify that the label is displayed.

tigert:
Ok, I have zero experience with playwright, so it is too much for this PR. I want to learn to write tests as a side quest at some point but not right now.

PR #2572, opened 2026-01-04
Use Embedded Python in MobiFlight

### LINKED ISSUE #2571: Use Python Embedded in MobiFlight
Currently MobiFlight uses a system wide installed python, which needs to be installed by the user himself. Thta leads to issues on some systems and is difficult for some users.

Embed python within MobiFlight, so that no separate installation of python is necessary any more.

Koseng:
fixes #2571

- Adds Python Runtime 3.14.2
- Adds the following packaging including their dependencies:
   - gql
   - simconnect
   - websockets

### DIFF
--- a/MobiFlight/Scripts/PythonEnvironment.cs
+++ b/MobiFlight/Scripts/PythonEnvironment.cs
@@ -37,10 +37,9 @@ namespace MobiFlight.Scripts
 
             // Ensure Python base folder exists
             if (!Directory.Exists(pythonBaseFolder))
-            {
-                string errorMessage = $"Python folder not found: {pythonBaseFolder}";
-                Log.Instance.log(errorMessage, LogSeverity.Error);
-                throw new DirectoryNotFoundException(errorMessage);
+            { 
+                Log.Instance.log($"Python folder not found: {pythonBaseFolder}", LogSeverity.Error);
+                return;
             }
 
             // Delete all existing folders in Python base folder before extraction
@@ -64,9 +63,8 @@ namespace MobiFlight.Scripts
 
             if (zipFiles.Length == 0)
             {
-                string errorMessage = "No Python runtime .zip files found to extract.";
-                Log.Instance.log(errorMessage, LogSeverity.Error);
-                throw new FileNotFoundException(errorMessage);
+                Log.Instance.log("No Python runtime .zip files found to extract.", LogSeverity.Error);         
+                return;
             }
 
             Log.Instance.log($"Found {zipFiles.Length} Python runtime archive(s) to extract.", LogSeverity.Info);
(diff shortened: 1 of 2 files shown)


MobiFlight/Scripts/PythonEnvironment.cs
```diff
@@ -0,0 +1,82 @@
...
+
+namespace MobiFlight.Scripts
+{
+    /// <summary>
+    /// Manages the Python runtime environment initialization
+    /// </summary>
+    internal class PythonEnvironment
+    {
+        public const string PYTHON_BASE_FOLDER = "Python";
+        public const string PYTHON_RUNTIME_FOLDER = "3.14.2";
```
DocMoebiuz: I would call this Version even though it is used as part of the folder.

MobiFlight/Scripts/PythonEnvironment.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// Initializes the Python runtime by extracting archived files if needed
+        /// </summary>
+        public static void Initialize()
+        {
+            // Get Python paths from settings
+            string pythonBaseFolder = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, PYTHON_BASE_FOLDER);
+            string pythonRuntimeFolder = Path.Combine(pythonBaseFolder, PYTHON_RUNTIME_FOLDER);
+
+            // Check if Python runtime is already extracted
+            if (Directory.Exists(pythonRuntimeFolder))
```
DocMoebiuz: i think it would be better to check for python.exe (or whatever it is called)
also the path to the python executable can be defined as constant here in python environment,

MobiFlight/Scripts/ScriptRunner.cs
```diff
@@ -39,31 +37,24 @@ internal class ScriptRunner
...
-            { "SimConnect", new Version(0,4) },
-        };
-
-
         public ScriptRunner(JoystickManager joystickManager, SimConnectCacheInterface msfsCache)
         {
+            // Get Python paths from settings                     
+            string pythonBaseFolder = Path.Combine(AppDomain.CurrentDomain.BaseDirectory, PythonEnvironment.PYTHON_BASE_FOLDER);
+            string pythonPath = Path.Combine(pythonBaseFolder, PythonEnvironment.PYTHON_RUNTIME_FOLDER);
+            PythonExecutable = Path.Combine(pythonPath, "python.exe");
```
DocMoebiuz: putting everything into PythonEnvironment is better in my opinion

MobiFlight/Scripts/PythonEnvironment.cs
Koseng: done

MobiFlight/Scripts/ScriptRunner.cs
Koseng: done

MobiFlight/Scripts/PythonEnvironment.cs
Koseng: done

DocMoebiuz (approved):

Issue #2331, opened 2025-09-15
Support dynamic connect and disconnect for Joysticks & Midi Controller


DocMoebiuz:
**Is your feature request related to a problem? Please describe.**
MobiFlight Boards can be connected and disconnected while MobiFlight is running. 
This is currently not available for Joysticks and Midi Controller.

**Describe the solution you'd like**
Support dynamic connect and disconnect for Joysticks & Midi Controller
- UI updates with controller changes
- Toast Notification if controller is connected or disconnected

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

Koseng:
I think reconnect for midi devices does already work. But only when on startup of mobiflight a midi device was connected.

tigert:
Also related is #2920 - the plug and unplug should also work if you plan to configure a device, but realize you dont have it connected. The legacy config wizards are not aware of this.

This already works for arduinos in the new device bindings dialog I think.

Ideally the controller selection and detection should be dynamic and consistent accross the UI.

PR #2662, opened 2026-01-26
Allow dynamic connect and disconnect for HID Controller

DocMoebiuz:
related #2331


GregWoods:
I've found an issue running this branch...

`device.GetProductName()` is used in 2 places. 
On my laptop, I have several listed devices where `device.ProductName` is an empty string. These "devices" don't appear to be real devices, but Windows still lists them.

If `device.ProductName` is an empty string then `device.GetProductName()` throws an error. This is currently treated as a critical error, and I don't believe it should be,

I would assume any error here would mean that we can simply skip that device, not adding it to the internal list of devices.

Issue #2764, opened 2026-02-25
FSUIPC Offset Input Action does not replace @ placeholder with current input value


tristanmcpherson:
**Describe the bug**
@ should be replaced with the current input value (for example the ProSim flight phase string), and that resolved value

**Steps To Reproduce**
1. Create an Output config with source ProSim and dataref aircraft.fms.flightPhase (verify it updates live).
2. In the same Output config, set Display Type to Input Action.
3. Configure OnChange action type FSUIPC - Offset.
4. Set target offset to string offset 0x2760 (size 255).
5. In More Options -> Set Value, enter @.
6. Run the sim and trigger a flight phase change.
7. Observe FSUIPC offset logging or PilotsDeck reading 0x2760.

**Expected behavior**
@ should be replaced with the current input value (for example the ProSim flight phase string), and that resolved value

**Actual behavior**
Offset 0x2760 receives literal text @.

**Details**
 - MobiFlight Version: Latest beta

PR #2765, opened 2026-02-25
Fix Output InputAction placeholder replacement for FSUIPC string offsets

tristanmcpherson:
Fixes https://github.com/MobiFlight/MobiFlight-Connector/issues/2764

Instead of checking args.Value, we should check args.StrValue if it is present. Additionally, we should send `value` and not `Value` for the 255 offset size. (Likely this could be better named)

### DIFF
--- a/MobiFlightUnitTests/MobiFlight/InputConfig/InputActionExecutionCacheTests.cs
+++ b/MobiFlightUnitTests/MobiFlight/InputConfig/InputActionExecutionCacheTests.cs
@@ -113,6 +113,32 @@ namespace MobiFlight.InputConfig.Tests
         }
 
         [TestMethod]
+        public void Execute_AnalogInputConfig_ExecutesWhenStrValueTransitionsFromNullToText()
+        {
+            var cache = new InputActionExecutionCache();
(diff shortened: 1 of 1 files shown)


MobiFlightUnitTests/MobiFlight/InputConfig/InputActionExecutionCacheTests.cs
```diff
@@ -0,0 +1,167 @@
...
+            var secondResult = cache.Execute(
+                config,
+                new CacheCollection(),
+                new InputEventArgs { Value = 42, StrValue = null },
+                new List<ConfigRefValue>());
+
+            Assert.IsTrue(firstResult);
+            Assert.IsFalse(secondResult);
+            Assert.AreEqual(1, action.ExecuteCount);
+        }
```
DocMoebiuz, line 87: The case where one event has strValue = null and the other one actually has strValue set, could be an interesting edge case that we would want to cover.

tristanmcpherson: Good idea, done

PR #2834, opened 2026-03-17
Restructure repository

ColinM9991:
To prepare for an upcoming .NET Core migration, this pull request restructures the repository to follow a more conventional structure containing `src` and `tests` directories.

The .NET Core SDK includes items automatically via glob patterns and so it is pertinent that projects are segregated. While it's possible to exclude files within the csproj file that could get quite messy and introduce risk down the line where files may accidentally be included in projects that they shouldn't be in.

A diff of the MobiFlightConnector.zip from this PR compared alongside a known good zip yields a positive result that shows the only left orphans (files in the known good but missing after this restructure) are the frontend/distassets that are hashed when generated, meaning the file name will always be different across runs.

image

# Note

Please don't merge without first validating the release process

Although I've updated some YAML workflow files, I have not tested the full release process to ensure the final artifact (that is attached to GitHub releases) is correct. This is where I will need some steer in terms of how this process is run locally.


DocMoebiuz (approved):
Fantastic!!!

PR #2872, opened 2026-04-03
Migration to .NET 10

ColinM9991:
# Description

Migrating from .NET Framework to .NET 10 is naturally a massive change, even if not at the surface. There are many changes in the SDK and runtime which improve security, deprecate APIs and some that affect the lower level memory-pinning to yield better performance. The SDK is also less forgiving which raises issues that may not have been observed prior to this migration.

# Notable changes

## Self-contained

The .NET core version is published as a self-contained application to ensure end users don't need to manually install .NET versions. The runtime ships with the application.

## MobiFlight Installer

The installer now uses an array of process paths for MFConnector with the first being `MFConnector.dll` for .NET Core's new compilation style and the second being `MFConnector.exe`. The installer will use the first file that exists.

This ensures thew Installer allows people to update in place between .NET Framework and .NET Core.

## Dependency Injection

Dependency injection has been added at the entry point via `IHostBuilder`. `MainForm` is currently the only service that's registered in the container. This paves the way for improving the testability of the application.

## DataSet and DataTable

These types are no longer part of the Windows Forms toolbox in later versions of .NET. These were originally configured incorrectly in that the type of the settings column was specified as `typeof(object)` while the actual type would be one of `OutputConfigItem` or `InputConfigItem`. .NET is more stringent than .NET Framework and the lower level implementation of the `DataSet` class, particularly with how types are resolved.

Short of redesigning the data loading process, `FsuipcConfigPanel` has been updated to use `XmlSerializer` to continue to support the existing functionality for the Presets.aip file

## UI Resource duplication

The resource files had a _lot_ of duplicate keys. These have all been removed

# TODO
- [x] Convert MobiFlightConnector to .NET 10
- [x] Migrate Unit Tests
- [x] Migrate Installer
- [x] Update workflow files to use .NET SDK
- [x] Test hardware and flight simulators


ColinM9991:
Non-exhaustive tests have been performed by tigert and RyanD across MSFS and X-Plane with Arduino and WINWING devices

tests/MobiFlightUnitTests/assets/Base/ConfigFile/interpolation-checkmark-test.mcc
DocMoebiuz, line 1: is this change to the mcc file necessary?

src/MobiFlight-Installer/model/MobiFlightUpdaterModel.cs
```diff
@@ -149,27 +162,12 @@ public static void GoExtractToDirectory(string zipPath, string destinationDirect
...
-
-                            if (File.Exists(backupFile))
-                                System.IO.File.Delete(backupFile);
-                            System.IO.File.Move(FileInstaller.FullName, backupFile);
-                        }
-                        else
-                        {
-                            Log.Instance.log("Mobiflight-Installer no need to be upgrade, extracting file skipped", LogSeverity.Info);
-                            continue;
-                        }
```
DocMoebiuz, line 169: I am not sure what use case this is.
I guess the MobiFlight-Installer that is contained in the zip file is newer than the local one.
We can only update the installer file if our process is not holding the file lock. That's why we move it to a different name and then.

With .NET we are now talking about dll files and not exe files anymore... but we will need something similar.

src/MobiFlight-Installer/UI/UpdaterMainForm.cs
```diff
@@ -14,15 +14,14 @@ public partial class UpdaterMainForm : Form
         private WebClient _webClient;
         private string CurrentFileName = "";
 
-        public UpdaterMainForm()
+        public UpdaterMainForm(bool useBeta = false)
```
DocMoebiuz: Do we have instances where we are still calling it without any parameter?
If no, we would not have to provide a default.

src/MobiFlightConnector/mobiflight.ico
DocMoebiuz, line 1: Did this file get lost in the other project folder structure change PR?

DocMoebiuz (commented):
Please review my comments.

src/MobiFlightConnector/mobiflight.ico
ColinM9991: It was previously referenced by relative path at the repository root but Visual Studio prefers copies that live in the same directory

tests/MobiFlightUnitTests/assets/Base/ConfigFile/interpolation-checkmark-test.mcc
ColinM9991: This was an interesting one. The file has been in the repository for some time but wasn't set to copy to output directory, so the test suite has never tested this file. Now that copy to output is done on a glob pattern, it turned out this file was out of sync with some changes over the years. The schema was slightly different to what was expected here.

ColinM9991:
Converted to a draft again as there are some issues with the installer workflow and the fact the installer needs to be able to self-update

DocMoebiuz (approved):
Amazing work! I am so happy that we can finally make this step.

PR #2879, opened 2026-04-06
README.md: Add information on development prerequisites

maboehme:


DocMoebiuz:
Thanks for adding some instructions. It is not really telling anything at the moment in regard to the new setup, i am sorry.
And we are in the process of migrating to .NET10. So there will be additional changes on how to setup the dev environment.

I really would like to establish the devcontainer as the primary dev environment for MobiFlight.
Dependencies can change and they will automatically show in the devContainer.

There are also scripts already, so that if anyone wanted to make it work outside the devContainer, then they could look at the scripts from the devconatiner json.

I think it easier to maintain in the future because it requires less update of a readme.md

maboehme:
> Thanks for adding some instructions. It is not really telling anything at the moment in regard to the new setup, i am sorry.
> And we are in the process of migrating to .NET10. So there will be additional changes on how to setup the dev environment.

No problem. Looking forward to hearing more about the new setup!

PR #2892, opened 2026-04-12
Move Updater to subdirectory and remove expert mode

ColinM9991:
Pre-requisite work for #2872 

The updater needs to be able to self update for now, but this isn't possible with the current architecture. This should really live in a separate directory. Ideally it would live up one directory but this wouldn't be compatible with existing installations so instead it moves down to a directory titled `Updater`. This allows the application to live by itself while updating MobiFlight-Connector in the parent directory.


ColinM9991:
Need to craft some XML to test the upgrade process

PR #2895, opened 2026-04-12
Use Direct3D9 to use different rendering pipeline than XPlane

### LINKED ISSUE #2894: UI doesn't update if focus on X-Plane window
**Describe the bug**
UI doesn't update if focus on X-Plane window.

**Steps To Reproduce**
1. Start MobiFlight
2. Add a config that changes, e.g. time elapsed
3. Start X-Plane and load into flight
4. Put MobiFlight into run mode 
5. Raw Value and Final value are updating
6. Move focus to xplane window

**Expected behavior**
Raw and Final Values will still update

**Actual behavior**
Raw and final values won't update

**Details**
 - MobiFlight Version: 11
 - FlightSim: X-Plane

DocMoebiuz:
fixes #2894


DocMoebiuz:
After rebooting my PC the problem went away. We won't merge this PR this time.

PR #2845, opened 2026-03-20
Octavi IFR-1 shift mode indicator

eFloh:
When applying this PR, the Octavi IFR-1 interface will have two new functions:

* added output for the blinking context LED (output bit 6 causes the current active context LED to flash starting in firmware ver. 1.2)
* added virtual Shift/Unshift-Mode Buttons that will be "pressed" when the IFR-1 is switching the shift context mode.
  these can be used to set the new output pin to make the context blink or to control a variable to detect the shift mode in Mobiflight.

Sample mobiflight rules:
```json
{
  "Name": "Test Shiftmode",
  "ConfigFiles": [
    {
      "Label": "New file",
      "FileName": null,
      "ReferenceOnly": false,
      "EmbedContent": true,
      "ConfigItems": [
        {
          "Source": {
            "MobiFlightVariable": {
              "TYPE": "number",
              "Name": "IsIFRShifted",
              "Number": 0.0,
              "Text": "",
              "Expression": "$"
            },
            "Type": "VariableSource"
          },
          "TestValue": {
            "type": 1,
            "Float64": 1.0,
            "String": null
          },
          "Device": {
            "Name": "Blink Context",
            "Pin": "Blink Context",
            "Brightness": 255,
            "PwmMode": false,
            "Type": "Output"
          },
          "DeviceType": "Output",
          "DeviceName": "Blink Context",
          "GUID": "aaa0b5d2-1e55-4048-a2b1-939fe9a845de",
          "Active": true,
          "Name": "blinky",
          "Type": "OutputConfigItem",
          "Controller": {
            "Name": "IFR1",
            "Serial": "JS-439ed980-c5f4-11ee-8001-444553540000"
          },
          "RawValue": "0",
          "Value": "0"
        },
        {
          "button": {
            "onPress": {
              "Type": "VariableInputAction",
              "Variable": {
                "TYPE": "number",
                "Name": "IsIFRShifted",
                "Number": 0.0,
                "Text": "0",
                "Expression": "0"
              }
            },
            "onRelease": null,
            "onLongRelease": null,
            "onHold": null,
            "LongReleaseDelay": 350,
            "HoldDelay": 350,
            "RepeatDelay": 0
          },
          "Device": {
            "Type": "Button",
            "Name": "Button_UNSHIFT_SW"
          },
          "DeviceType": "Button",
          "DeviceName": "Button_UNSHIFT_SW",
          "GUID": "43931555-2f52-4acb-ac1c-25582934624f",
          "Active": true,
[...]

PR #2918, opened 2026-04-20
Octavi IFR-1: Auto context mode LED blinking when shifted

eFloh:
This enabled Mobiflight to use the context mode LED blinking in shift mode automagically without the need to define some manual rules or variables.

- still supports manual output rules by renaming the output label (e.g. remove the "Auto" in the "Blink Context" entry). This will disable all automatic behaviour.
  - The Auto Blink Context Output Device Pin will not be usable in custom rules and will always automagically have the value 1 if the mode is shifted.
  - if renamed, the Output Device Pin can be used in custom rules and is never touched internally. 
- see SampleProject for config possibilities, also for possible rules for knob pushes that emulate shift mode (e.g. to indicate a FMS cursor is active)

downwards compatible with #2845: existing configs using the original joystick.json device name will not be touched. (joystick.json has to be replaced on every app update though.)


eFloh:
@felixc7 Das war leichter als gedacht, was meinst du dazu?

eFloh:
@felixc7 ich habe diesen PR nun mal aktiviert, ggf. kannst du gerne ein Update vorschlagen oder dich wg. einem Floowup melden.

PR #2874, opened 2026-04-05
Add WinCtrl CDU support for X-Plane default A330.

maboehme:
After submitting #2807, which adds support for the X-Plane default FMC,
I realized that the default A330 included in X-Plane uses the same
datarefs. So this PR simply adds a config to ScriptMappings.json to use
the existing script with the A330.

PR #2923, opened 2026-04-22
Fix WinCtrl CDU script loading for Toliss A330.

maboehme:
I've realized that #2874, which added WinCtrl CDU support for the
X-Plane default A330, inadvertently broke loading of the WinCtrl CDU
script for the Toliss A330.

The reason for this is that the `AircraftIdSnippet` that I specified for
the default A330, "airbus a330", is also a substring of the Aircraft ID
for the Toliss A330. As a result, _both_ scripts (toliss_a3xx.py and
xplane_default_fmc.py) are run at the same time, both try to control the
CDU, and as a result, neither controls it correctly.

This problem cannot be resolved with the existing `AircraftIdSnippet`
mechanism as the Aircraft ID for the default A330, "airbus a330", is a
substring of the Aircraft ID for the Toliss A330. As a result, it is not
possible to specify an `AircraftIdSnippet` for the default A330 that
does not also match the Toliss A330.

What I have therefore done is to introduce an alternative JSON field,
`AircraftIdRegex`, that matches the Aircraft ID using a regex. For the
default A330, I specify `^airbus a330`. Because the Aircraft ID for the
Toliss A330 starts with "Toliss", this regex does not match the Toliss
A330.

The remaining entries in ScriptMapping.cs continue to use
`AircraftIdSnippet` as there was no need to change them. As we add
support for more aircraft, I can however foresee that there will be more
"conflicts" in the future that can't be resolved with the
`AircraftIdSnippet` mechanism, and we will be able to use the regex
mechanism for these too.

### DIFF
--- a/src/MobiFlightConnector/MobiFlight/Scripts/ScriptMapping.cs
+++ b/src/MobiFlightConnector/MobiFlight/Scripts/ScriptMapping.cs
@@ -5,6 +5,7 @@
         public string VendorId { get; set; }
         public string[] ProductIds { get; set; }
         public string AircraftIdSnippet { get; set; }
+        public string AircraftIdRegex { get; set; }
         public string ScriptName { get; set; }
     }
 }
--- a/src/MobiFlightConnector/MobiFlight/Scripts/ScriptMapping.cs
+++ b/src/MobiFlightConnector/MobiFlight/Scripts/ScriptMapping.cs
@@ -4,8 +4,11 @@
     {
         public string VendorId { get; set; }
         public string[] ProductIds { get; set; }
+        // Deprecated, use AircraftMatchPattern instead.
         public string AircraftIdSnippet { get; set; }
-        public string AircraftIdRegex { get; set; }
+        // Regex to match against the aircraft ID. The script is run if the
+        // pattern matches somewhere within the aicraft ID.
+        public string AircraftMatchPattern { get; set; }
         public string ScriptName { get; set; }
     }
 }
--- a/src/MobiFlightConnector/MobiFlight/Scripts/ScriptRunner.cs
+++ b/src/MobiFlightConnector/MobiFlight/Scripts/ScriptRunner.cs
@@ -302,14 +302,14 @@ namespace MobiFlight.Scripts
 
         private static bool AircraftMatchesScriptMapping(string aircraftDescription, ScriptMapping scriptMapping)
         {
-            if (!string.IsNullOrEmpty(scriptMapping.AircraftIdSnippet))
-            {
-                return aircraftDescription.Contains(scriptMapping.AircraftIdSnippet);
-            }
-
             if (!string.IsNullOrEmpty(scriptMapping.AircraftMatchPattern))
             {
                 return Regex.IsMatch(aircraftDescription, scriptMapping.AircraftMatchPattern);
+            }
+
+            if (!string.IsNullOrEmpty(scriptMapping.AircraftIdSnippet))
+            {
+                return aircraftDescription.Contains(scriptMapping.AircraftIdSnippet);
             }
 
             return false;
--- a/src/MobiFlightConnector/MobiFlight/Scripts/ScriptRunner.cs
+++ b/src/MobiFlightConnector/MobiFlight/Scripts/ScriptRunner.cs
@@ -307,9 +307,9 @@ namespace MobiFlight.Scripts
                 return aircraftDescription.Contains(scriptMapping.AircraftIdSnippet);
             }
 
-            if (!string.IsNullOrEmpty(scriptMapping.AircraftIdRegex))
+            if (!string.IsNullOrEmpty(scriptMapping.AircraftMatchPattern))
             {
-                return Regex.IsMatch(aircraftDescription, scriptMapping.AircraftIdRegex);
+                return Regex.IsMatch(aircraftDescription, scriptMapping.AircraftMatchPattern);
             }
 
             return false;
--- a/src/MobiFlightConnector/MobiFlight/Scripts/ScriptRunner.cs
+++ b/src/MobiFlightConnector/MobiFlight/Scripts/ScriptRunner.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -299,6 +300,21 @@ namespace MobiFlight.Scripts
             }
         }
 
+        private static bool AircraftMatchesScriptMapping(string aircraftDescription, ScriptMapping scriptMapping)
+        {
+            if (!string.IsNullOrEmpty(scriptMapping.AircraftIdSnippet))
+            {
+                return aircraftDescription.Contains(scriptMapping.AircraftIdSnippet);
+            }
+
+            if (!string.IsNullOrEmpty(scriptMapping.AircraftIdRegex))
+            {
+                return Regex.IsMatch(aircraftDescription, scriptMapping.AircraftIdRegex);
+            }
+
+            return false;
+        }
+
         private void CheckAndExecuteScripts(string aircraftDescription)
         {
             var executionList = new List<string>();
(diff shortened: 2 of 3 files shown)


theegg52:
After a user report on the discord I think it would be beneficial to switch the xplane md-82 over to the regex format as well. In a rare edge case it is conflicting with the msfs md-82 causing mobiflight to try and load both scripts. I've done some testing with the PR build and a regex of "^mcdonnell douglas md-82" fixes the issue and both aircraft are functioning normally. link to discord thread - https://discord.com/channels/608690978081210392/1497140948948226218

maboehme:
> After a user report on the discord I think it would be beneficial to switch the xplane md-82 over to the regex format as well. In a rare edge case it is conflicting with the msfs md-82 causing mobiflight to try and load both scripts. I've done some testing with the PR build and a regex of "^mcdonnell douglas md-82" fixes the issue and both aircraft are functioning normally. link to discord thread - https://discord.com/channels/608690978081210392/1497140948948226218

Thanks for the heads up!

I've submitted a separate PR to fix this issue as #2925. I don't think it's necessary to use a regex in this particular case, as the string "mcdonnell douglas" does not appear to occur in the aircraft ID for the Leonardo Maddog, so I've chosen to use the regular `AircraftIdSnippet`, which also allows this to be submitted as a separate PR.

DocMoebiuz:
@maboehme, @Koseng - I am not familiar with how the AircraftIdSnippet is used in other parts of the script.
The naming doesn't really tell what it is but if we are essentially using it for matching the aircraft based on the path maybe we can simply get away with one property anyways.

The matching logic is already scoped to a sim (xplane vs msfs) right?

maboehme:
> @maboehme, @Koseng - I am not familiar with how the AircraftIdSnippet is used in other parts of the script. The naming doesn't really tell what it is but if we are essentially using it for matching the aircraft based on the path maybe we can simply get away with one property anyways.

I believe on MSFS the Aircraft ID may be the entire path of the addon folder, but on X-Plane at least, the Aircraft ID is only the "UI name" returned by the sim/aircraft/view/acf_ui_name dataref. This is queried [here](https://github.com/MobiFlight/MobiFlight-Connector/blob/0db7ca42438b44335cb327da26d9bc4d74dabf5f/src/MobiFlightConnector/xplane/XplaneCache.cs#L111).

The Aircraft ID for the X-Plane default A330 is "airbus a330". For the Toliss, it is "toliss airbus a330-941 hi def", and so any `AircraftIdSnippet` that one might choose for the default A330 by necessity also triggers the script for the Toliss A330.

A regex gives more flexibility here because it makes it possible, for example, to specify that a string should occur at the beginning of the Aircraft ID (e.g. `^airbus a330`).

> The matching logic is already scoped to a sim (xplane vs msfs) right?

I don't believe so. Take a look at [ScriptMappings.json](https://github.com/MobiFlight/MobiFlight-Connector/blob/0db7ca42438b44335cb327da26d9bc4d74dabf5f/src/MobiFlightConnector/Scripts/ScriptMappings.json) -- there's nothing here that refers to the sim, and the file contains a mix of MSFS and X-Plane addone. I've also thought though that we might at some point need a way to specify which sim the script is for -- but for now this hasn't turned out to be necessary yet.

DocMoebiuz:
I wonder why we would need a new property. I think allowing the snippet to contain regex should be just fine. Am I missing something?

src/MobiFlightConnector/MobiFlight/Scripts/ScriptMapping.cs
```diff
@@ -5,6 +5,7 @@ public class ScriptMapping
         public string VendorId { get; set; }
         public string[] ProductIds { get; set; }
         public string AircraftIdSnippet { get; set; }
+        public string AircraftIdRegex { get; set; }
```
DocMoebiuz: I would like to avoid introducing a new property.
Let us use the existing AircraftIdSnippet.

DocMoebiuz (changes requested):

maboehme:
> I wonder why we would need a new property. I think allowing the snippet to contain regex should be just fine. Am I missing something?

Sorry for the delay in responding.

Makes sense to keep it simple. Two thoughts:

* I should check that none of the existing snippets in ScriptMappings.json contain characters that have special meaning in regexes.
* Should we keep the name "AircraftIdSnippets" or rename it? "Snippets" implies that this is just a substring, so people adding new entries might not realize they need to escape characters with special meaning in regexes.

DocMoebiuz:
i would try to find a name that conveys the intention better

aircraftMatchPattern 

or similar

DocMoebiuz:
@Koseng please chime in too

maboehme:
> i would try to find a name that conveys the intention better
> 
> aircraftMatchPattern
> 
> or similar

Sorry for the delay in replying.

What do you think of `AircraftRegex`?

DocMoebiuz:
I would avoid mentioning technical details like regex because those might change over time.

A stand by my suggestion: `aircraftMatchPattern`
It clearly says: this value is used for matching the script to the aircraft. It is not too explicit about name or path or technique. I think this can be covered by docs or a simple comment.


Oh. We have to think of backward compatibility. Users might have their own script which is not bundled with MobiFlight.

This would be a reason to keep the current name, or introduce a new field and label the old one as deprecated and update all of our scripts that are bundled but still check for the old name in case the new one is not present.

Koseng:
I

> I would avoid mentioning technical details like regex because those might change over time.
> 
> A stand by my suggestion: `aircraftMatchPattern` It clearly says: this value is used for matching the script to the aircraft. It is not too explicit about name or path or technique. I think this can be covered by docs or a simple comment.
> 
> Oh. We have to think of backward compatibility. Users might have their own script which is not bundled with MobiFlight.
> 
> This would be a reason to keep the current name, or introduce a new field and label the old one as deprecated and update all of our scripts that are bundled but still check for the old name in case the new one is not present.

Also like that proposal.

maboehme:
> I would avoid mentioning technical details like regex because those might change over time.
> 
> A stand by my suggestion: `aircraftMatchPattern` It clearly says: this value is used for matching the script to the aircraft. It is not too explicit about name or path or technique. I think this can be covered by docs or a simple comment.

Done. (I've capitalized it as `AircraftMatchPattern` to match the capitalization of the other fields in ScriptMappings.json.)

> Oh. We have to think of backward compatibility. Users might have their own script which is not bundled with MobiFlight.
> 
> This would be a reason to keep the current name, or introduce a new field and label the old one as deprecated and update all of our scripts that are bundled but still check for the old name in case the new one is not present.

Makes sense. I've updated all entires in ScriptMappings.json to use `AircraftMatchPattern`, but I've kept the support for `AircraftIdSnippet` in place.

I obviously haven't been able to test all the addons in ScriptMappings.json to make sure they still work, but I've verified that the existing strings don't use any characters that have special meaning in regexes, so they _should_ still continue to work as before.

maboehme:
PS I see there's a merge conflict that has to be resolved, but I wanted to hold off on rebasing the branch to head until the review is complete.

DocMoebiuz:
@maboehme - ~~it looks like i cannot review the changes as long as there are merge conflicts.~~~
Ok, i was able to review it.

src/MobiFlightConnector/MobiFlight/Scripts/ScriptRunner.cs
```diff
@@ -275,6 +276,21 @@ private void Process_OutputDataReceived(object sender, DataReceivedEventArgs e)
...
+        {
+            if (!string.IsNullOrEmpty(scriptMapping.AircraftIdSnippet))
+            {
+                return aircraftDescription.Contains(scriptMapping.AircraftIdSnippet);
+            }
+
+            if (!string.IsNullOrEmpty(scriptMapping.AircraftMatchPattern))
+            {
+                return Regex.IsMatch(aircraftDescription, scriptMapping.AircraftMatchPattern);
+            }
```
DocMoebiuz: I believe it would make more sense to switch the order. 
This is the new default, and should go first, fallback is the old AircraftIdSnippet.

DocMoebiuz (approved):
I leave it up to you if you change the order or not (see comment)
The rest is looking great. Thanks!

src/MobiFlightConnector/MobiFlight/Scripts/ScriptRunner.cs
maboehme: Good point -- done!

maboehme:
> I leave it up to you if you change the order or not (see comment) The rest is looking great. Thanks!

Thanks. I've rebased the branch now!

DocMoebiuz:
@maboehme - thanks! I will merge this PR now. Please update the description so that it uses the correct property names.

Issue #2885, opened 2026-04-09
Noisy potentiometers cause the WASM module to stop working


neilenns:
**Describe the bug**
Noisy potentiometers flood the connector with messages, which in turn floods the WASM module with messages, leading to the WASM module stopping.

This seems to be more prevalent with v11, and users report it didn't happen until after they upgraded from 10.5.3, although we have not independently confirmed that.

We first saw this from users with Rowsfire products but have since seen it from WinCtrl USB HID devices as well.

See these Discord threads for examples:

* https://discord.com/channels/608690978081210392/1491315006300749825
* https://discord.com/channels/608690978081210392/1489993174456729672
* https://discord.com/channels/608690978081210392/1469359131457753240
* https://discord.com/channels/608690978081210392/1491734380211867688

neilenns:
Example log file of a WinCtrl device with a light sensor sending a ton of messages.

[winctrl.txt](https://github.com/user-attachments/files/26602432/winctrl.txt)

tigert:
https://discord.com/channels/608690978081210392/1491734380211867688/1491764185770496070

Could be electrical noise? It's odd that a lot of HID devices are so chatty.

Need to analyze the log and see whether there actually are patterns there, or a constant interval which would indicate that.

AI suggested that but I want to see it for myself with some kind of plotting tool.

Another pattern here is probably that many people are using MobiFlight only to manage those devices, and we currently lack any means to set dead zones or calibrate the axis, and I am pretty sure manufacturers own apps do quite a bit of filtering on their software to fix issues like this in hardware.

We probably could do some basic filtering of repetitive noise patterns by applying some low pass filtering on all axis, since for example in this case its a repetitive jitter

tigert:
Ok, so in this particular case with the EFIS, it was a matter of calibrating the device in sim app pro.

I think we just need end point and dead zone calibration (for centering controls) for both potentiometers and joystick axis in MobiFlight. Ideally it should be something like Ralf was envisioning with the autocalibration that you move the axis around and during the calibration we measure jitter and just add a bit of safety margin around it. 

But this encounter with calibration / noise issues with HID devices made me think we need to handle it for all axis, not just Arduino potentiometers.

Because if you only use MobiFlight, there is a high chance that the person never for example calibrated the unit in simapp pro. I know I never install manufacturers drivers because they are just annoying and want to do too many things.

[...]

PR #2929, opened 2026-04-30
Apply filter to analog inputs for bodnar boards

DocMoebiuz:
This PR filters the noisy raw values which are coming from the Leo Bodnar Boards.

Additionally, it introduces a couple of modifiers which were created during testing. Some modifiers are used by the Bodnar Board directly, statically. Other modifiers did not turn out to be helpful for this use case but we keep them for the future.

- [x] Simple Moving Average filter applied to raw data
- [x] Quantize modifier applied to filtered data
- [x] Exponential Moving Average modifier added, not used yet
- [x] Minor formatting/cleanup updates in a few existing files and updated contributor Copilot instructions. 
- [x] unit tests 

part of #2885


neilenns:
@DocMoebiuz reminder that this will need doc updates, we have documentation for all filters and troubleshooting for pots. Adding myself to this so I know when it merges.

PR #2936, opened 2026-05-02
Add playwright results and report artifacts to .gitignore

tigert:
These feel like they don't belong to the repository, so adding them to .gitignore.

```
# Playwright test artifacts
frontend/playwright-report/*
frontend/test-results/*

```
In the case that I would be grossly mistaken, I am pre-emptively supplying myself with a handy paper-bag to wear. Like, maybe there is a "clean up the artifacts" command one tends to as part of the playwright test workflow, that takes care of this. So, just in case:

```
,-._,--.
|      |
| O  O |
!~~~~~~!

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -490,5 +490,7 @@ FodyWeavers.xsd
 # Python directory - include only RuntimeAndPackages*.zip files
 Python/*
 !Python/RuntimeAndPackages*.zip
+
+# Playwright test artifacts
 frontend/playwright-report/*
 frontend/test-results/*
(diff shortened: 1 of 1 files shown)


tigert:
also added the devcontainer lock file so it doesnt show up as git changes in vscode.

DocMoebiuz:
They are already included in the frontend/.gitignore list.
not sure why these would show up for you.

DocMoebiuz:
do you know where the devcontainer.lock file is coming from because i don't have that on my machine.

PR #2957, opened 2026-05-08
Add support for IIDB Steering Tiller (CPT & FO) with backlight control

IIDBpro:
Added support for IIDB Steering Tiller (CPT & FO versions) with backlight control via custom HID class

### DIFF
--- a/src/MobiFlightConnector/Joysticks/iidb/iidb_tiller_a32x_cpt.joystick.json
+++ b/src/MobiFlightConnector/Joysticks/iidb/iidb_tiller_a32x_cpt.joystick.json
@@ -1,0 +1,21 @@
+{
+  "$schema": "./mfjoystick.schema.json",
+  "InstanceName": "IIDB Steering Tiller A32x CPT",
+  "VendorId": "0x99DB",
+  "ProductId": "0x0102",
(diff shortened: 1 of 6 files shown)


DocMoebiuz (approved):
Cool!

DocMoebiuz (changes requested):
See copilot feedback. Many of the things make sense.

IIDBpro:
We have addressed the feedback from the code review: renamed the class to IIDBDevice, removed empty try-catch blocks with proper logging, and extracted magic numbers into constants. Ready for another look!

DocMoebiuz:
> We have addressed the feedback from the code review: renamed the class to IIDBDevice, removed empty try-catch blocks with proper logging, and extracted magic numbers into constants. Ready for another look!

@IIDBpro so you basically could not use just the json file because it is limited to setting bits instead of a byte. Is that correct?

IIDBpro:
The protocol doesn't use a bits or byte system.
You can specify any bit in the JSON file, and everything will work.
Because it works by functions our class.
As far as I understand, the JSON file is needed so that the app will have the required output option when selecting a device.

DocMoebiuz (approved):

PR #2968, opened 2026-05-17
Disable DTR for megas in general, enable it for only the one that requires it.

tigert:
Enabling DTR for all boards breaks ESP32 based custom device projects currently. The reason DTR is enabled, is because one variant of Arduino Mega needs it. Most megas and also esp32 use CH340 that share the same VID & PID, but luckily the affected Arduino Mega variant uses different ID's.

- Disable DTR for all Mega boards
- Enable DTR specifically for the Mega wth Atmega16U2 that needs it


tigert:
marked this as draft so we can give it some testing first.

tigert:
We are currently applying a fix to all megas regardless of whether they need it or not.

Yet if I understand correctly, the problem only seems to affect the megas that have the Atmega16U2 instead of CH340, or custom device projects with ESP32. I know @ccrawford has been using this fix for a while already.

elral:
What is about VID 2341/PID 0010? This is also a Mega 2560 from Arduino which connects via CDC ACM. For me it looks like DTR enable is also required for this chip.

DocMoebiuz:
This PR has been draft for more than 3 weeks

DocMoebiuz:
Nothing has happened here. I will close this draft PR.

PR #2972, opened 2026-05-18
Fix editing in ControllerBindingsDialog's dropdown search field

### LINKED ISSUE #2971: Keyboard Events inside Dialog are processed outside and cause unexpected issues
**Describe the bug**
When you search for controller in the dropdown menu's built in search, you cannot use del or backspace to edit the search string it seems, something blocks that. At the same time using backspace, you will accidentally delete the currently selected item(s) from the config list.

Same thing happens with <spacebar> - the selected item from the list will toggle active/inactive state.

This happens because the keyboard events bubble up the DOM tree and are processed outside the dialog, in this case the config list view.

For settings dialog we are already handling all keydown events on the form, and also stopping ALL propagation. This has been by accident. That's why we didn't see the issue happening there.

**Steps To Reproduce**
1. Open existing project.
2. Select a config entry from the list
3. Open the Controller Binding Dialog
4. Hit space bar or open the combobox and type and use backspace 
5. See error

**Expected behavior**
With <spacebar> nothing should happen.
With DEL the text from the combobox input element should be deleted.

**Actual behavior**
<spacebar> toggles active state from selected config list item.
<DEL> deletes the currently selected config list item - but the text doesnt change.

**Details**
 - MobiFlight Version: 11.1

tigert:
I wasnt expecting to procrastinate by fiddling with html and css, but here we are.

- ~~remove `select-none` from top level ControllerBindingsDialog~~ not the case anymore
- add `select-text` to CommandInput component instead, so it is enabled on all input fields
- stop ConfigItemTable's EventListener from eating our keystrokes if we are in a text field (it's active when ControllerBindingDialog is open)

Fixes #2971

### DIFF
--- a/src/MobiFlightConnector/frontend/src/components/tables/config-item-table/items/ConfigItemTableBody.tsx
+++ b/src/MobiFlightConnector/frontend/src/components/tables/config-item-table/items/ConfigItemTableBody.tsx
@@ -30,6 +30,15 @@ const ConfigItemTableBody = forwardRef<
 
   useEffect(() => {
     const handleKeyDown = (e: KeyboardEvent) => {
+      const target = e.target as HTMLElement
+  
+      const isInputField = 
+        target.tagName === "INPUT" ||
+        target.tagName === "TEXTAREA" ||
+        target.isContentEditable === true
+
+      if (isInputField) return // bail out when an editable widget has focus
+
       const supportedKeyPress = ["Delete", " ", "Enter", "Escape", "Backspace"]
 
       if (!supportedKeyPress.includes(e.key)) return
--- a/src/MobiFlightConnector/frontend/src/components/ui/command.tsx
+++ b/src/MobiFlightConnector/frontend/src/components/ui/command.tsx
@@ -42,7 +42,7 @@ const CommandInput = React.forwardRef<
     <CommandPrimitive.Input
       ref={ref}
       className={cn(
-        "flex h-11 w-full rounded-md bg-transparent py-3 text-sm outline-hidden placeholder:text-muted-foreground disabled:cursor-not-allowed disabled:opacity-50",
+        "flex h-11 w-full rounded-md bg-transparent py-3 text-sm outline-hidden placeholder:text-muted-foreground disabled:cursor-not-allowed disabled:opacity-50 select-text",
         className
       )}
       {...props}


tigert:
Hm, looks like this broke the profile tab toolbar tests. Will look into it tomorrow.

src/MobiFlightConnector/frontend/src/components/controllers/ControllerBindingsDialog.tsx
```diff
@@ -121,7 +121,7 @@ const ControllerBindingsDialog = ({
 
   return (
     <Dialog open={isOpen} onOpenChange={onOpenChange}>
-      <DialogContent className="vsm:min-h-[75%] vxl:min-h-[60%] flex min-h-[90%] flex-col overflow-y-auto select-none sm:max-w-150 lg:max-w-200 xl:max-w-250">
+      <DialogContent className="vsm:min-h-[75%] vxl:min-h-[60%] flex min-h-[90%] flex-col overflow-y-auto sm:max-w-150 lg:max-w-200 xl:max-w-250">
```
DocMoebiuz: i would leave select-none as default for the dialog and only add select-text to the input fields if that was needed.

src/MobiFlightConnector/frontend/src/components/tables/config-item-table/items/ConfigItemTableBody.tsx
```diff
@@ -30,6 +30,15 @@ const ConfigItemTableBody = forwardRef<
 
   useEffect(() => {
     const handleKeyDown = (e: KeyboardEvent) => {
+      const target = e.target as HTMLElement
+
```
DocMoebiuz: i prefer to create a const variable with a name and initialize it with the test expression:


const isInputField = test1 || test2 || ...


and then you have

if(isInputField)

This style improves readability.

tigert:
Thanks for the comments, I will have a look. Good learning opportunity 😎

src/MobiFlightConnector/frontend/src/components/controllers/ControllerBindingsDialog.tsx
tigert: you are actually right, removing that is not necessary since the input style overrides it anyway.

tigert:
gah I had to force-push, since I ended up with duplicate commits, let me check everything works and we can then proceed.

so if the commit looks a bit *cough* squashed, that's why ;)

src/MobiFlightConnector/frontend/src/components/tables/config-item-table/items/ConfigItemTableBody.tsx
tigert: So it is now like this:

```typescript
      const target = e.target as HTMLElement
  
      const isInputField = 
        target.tagName === "INPUT" ||
        target.tagName === "TEXTAREA" ||
        target.isContentEditable === true

      if (isInputField) return // bail out when an editable widget has focus

```

tigert: I agree it makes sense.

And now I actually understand why there are very little actual inline comments in the code, when it is written like this, it is indeed more clear in its own.

src/MobiFlightConnector/frontend/src/components/ui/command.tsx
```diff
@@ -42,7 +42,7 @@ const CommandInput = React.forwardRef<
     <CommandPrimitive.Input
       ref={ref}
       className={cn(
-        "flex h-11 w-full rounded-md bg-transparent py-3 text-sm outline-hidden placeholder:text-muted-foreground disabled:cursor-not-allowed disabled:opacity-50",
+        "flex h-11 w-full rounded-md bg-transparent py-3 text-sm outline-hidden placeholder:text-muted-foreground disabled:cursor-not-allowed disabled:opacity-50 select-text",
```
DocMoebiuz, line 45: this is unrelated to the behavior. we can remove this.

DocMoebiuz (changes requested):

DocMoebiuz:
I don't think this is the right fix for the issue.
What happens is that the keyboard events bubble up to the confglistview and if you hit backspace for example then the item that is currently selected in the list will be deleted. 

I will double check why this behavior doesn't happen for the new input config wizards and how we prevent it there.

DocMoebiuz:
I will close this PR. Sorry.

PR #3022, opened 2026-06-10
Updated fslabs_winwing_cdu.py

crazydunc:
### Summary
Fixes the FSLabs Winwing CDU bridge (fslabs_winwing_cdu.py) loading a blank screen until a button is pressed, changes Cyan colour to Blue, inline with in Sim MCDU plus several related reliability issues with connection handling.

### Problems

- Blank screen on load: the first display frame was fetched and de-duplicated before the MobiFlight WebSocket finished connecting, so it was silently dropped. Because the fetcher only re-sends on change, nothing was sent until the display next changed — i.e. until the user pressed a button.
- CDU could die permanently: the WebSocket task gave up after 3 retries and the wrapper then cancelled the fetch/process tasks for that CDU. This killed the CDU for the rest of the session if the script started before the sim was ready, or after any sustained disconnect.
- Disconnects went undetected: after connecting, the task busy-looped on a short sleep and never read the socket, so a dropped connection was never noticed and the reconnect path never armed.
- Persistent HTTP connection corruption: non-200 responses were left undrained, putting the persistent HTTPConnection into a ResponseNotReady state on the next request.
- Font-ordering race: the connection was exposed before the font command had settled, so the first frame could render in the wrong font.

### Changes

- Re-send current frame on (re)connect: track the latest frame per MCDU (last_frames) and push it immediately whenever the socket connects, draining any stale backlog first. Fixes both the cold-start blank screen and repaint-after-reconnect.
- Reconnect indefinitely after first connect: the retry cap now only applies before the first successful connection (the genuine "no CDU attached" case); transient drops no longer permanently kill the CDU.
- Detect dropped connections: block on the socket iterator (like the Fenix bridge) instead of busy-looping, and reset the connection to None in a finally so the reconnect/resync logic actually fires. Removes the busy-loop.
- Drain non-200 HTTP responses to keep the persistent connection usable.
- Set the font and let it settle before publishing the connection, eliminating the first-frame font race.
- Crash-proof sends: wrap frame sends in try/except so a socket dropping mid-send logs and resyncs on reconnect instead of crashing the process task.
- Remove dead subs entries: the five string keys could never match, since lookups are always by integer ASCII value.

### DIFF
--- a/src/MobiFlightConnector/Scripts/Winwing/fslabs_winwing_cdu.py
+++ b/src/MobiFlightConnector/Scripts/Winwing/fslabs_winwing_cdu.py
@@ -8,7 +8,7 @@ import http.client
 
 FSL_COLOR_MAP = {
     0: "w",  # black (ignore)
-    1: "c",  # cyan
+    1: "o",  # cyan
     2: "e",  # gray
     3: "y",  # yellow
     4: "g",  # light green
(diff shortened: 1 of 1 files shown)


Koseng:
Made a short test, and looks good to me. Ready to be merged.

PR #3040, opened 2026-06-15
Define Aircraft on Project level which determines options in Config Wizard

### LINKED ISSUE #2819: Associate aircraft with project
A project is typically created for one specific aircraft, maybe a set of aircraft. 
All input and output configurations will utilize presets that are relevant to only these aircraft.

As a user I can define the context of one or more aircraft for my project.
Subsequent dialogs will honor this selection.
The set of aircraft will also be relevant for the auto-link feature.

related to #1691

DocMoebiuz:
### Summary
You can now define the specific aircraft that you would like to use in your project. This will reduce the number of presets later in the config wizard dialogs based on your selection.

MSFS and X-Plane have this option currently only because we are deriving it right now from the presets categories. Other sims don't have such presets.

#### MSFS summary 
image

#### MSFS aircraft selection
image

### Acceptance criteria

- [ ] User can define a set of aircraft for the project
  - [x] Option available for MSFS and X-Plane projects
  - [x] Options are derived from preset categories
  - [ ] MSFS: Options show preset type, output, input or potentiometer
  - [ ] X-Plane: Options show preset type, output, input, outputinput, or potentiometer
- [x] Sensible default categories are used when switching simulator type
  - [x] MSFS: Microsoft and Asobo Generic
  - [x] XPlane: Laminar Generic.

### DoD Checklist

- [x] Unit tests available
  - [ ] .NET backend
  - [ ] Frontend
- [x] Frontend tests
  - [ ] core (en,de,es)
  - [ ] additional langs
- [ ] documentation
     
## Related issues

fixes #2819

### Out-of-scope
- Auto-linking settings


FoxtrotSierra6829:
Please consider adding "show more" for the dialogs then. Most aircraft do not have functions defined for every possible switch, however presets for other aircraft are frequently compatible.

DocMoebiuz:
> Please consider adding "show more" for the dialogs then. Most aircraft do not have functions defined for every possible switch, however presets for other aircraft are frequently compatible.

I think i would be adding this in a next iteration. Trial and error for like aircraft that are not generic or not of the name that we expect is kind of weird, isn't it? How many users will actually really dig in other planes? Maybe limiting it will encourage users to submit the event under the correct aircraft category.

FoxtrotSierra6829:
So far I had to do this for every single plane I configured. If something isn't there I just search by name through everything. For quite a number of things I also had to do custom code. Do other users just not use certain functions of their hardware?
Using the WinWing EFIS, FCU, thrust lever, PAC and AGP.

(On that note, I should probably submit some of my learnings to HubHop)

DocMoebiuz:
> So far I had to do this for every single plane I configured. If something isn't there I just search by name through everything. For quite a number of things I also had to do custom code. Do other users just not use certain functions of their hardware? Using the WinWing EFIS, FCU, thrust lever, PAC and AGP.
> 
> (On that note, I should probably submit some of my learnings to HubHop)

I am absolutely paying attention to what you are describing and I can see that this is something we are doing today.  The question though is: is it the correct way or are we navigating with limitations provided by the system so far.

I still believe though that this kind of filtering will actually have a positive effect on the classification of our presets.
Right now you might find the preset of a totally unrelated aircraft which fits... but it is kind of random and you are not going to do anything about it, I guess - like fixing it and putting it in the right category.

PR #3057, opened 2026-06-20
Stop hold/repeat timers on ExecutionManager.Stop() so they dont keep running forever

### LINKED ISSUE #2890: On hold config with repeat timer set continues to run if config is disabled or mobiflight is stopped.
**Describe the bug**
On hold configs with a repeat continue to run if precondition disables config or mobiflight is stopped with the button/switch still being held.

**Steps To Reproduce**
1. Create a config which has a repeating on hold action.
2. Disable the config either by switching it off or by precondition. alternatively stop mobiflight from running
3. Repeating config will continue to run.

**Expected behavior**
If the config becomes inactive either by being disabled using the toggle or by precondition or mobiflight is stopped I would expect the repeating action to stop running.

**Actual behavior**
The repeating action continues to run even if mobiflight is stopped. sim responds according to what the action is doing. re enabling the config and pressing and releasing the button stops it running.

**Details**
 - MobiFlight Version: 11.0.1
 - FlightSim: MSFS 2024
 - AddOn: Ifly 737 Max8

**Additional context**
Tested using both a switch on an Arduino device and a joystick button.

tigert:
### Summary
When a button config has OnHold event defined with a timeout, and a OnHold event is received, a timer is created to execute the event at specified intervals until the button is released. When that happens, the button OnRelease event clears the timer. 

But if you stopped MobiFlight while the button was being held in pressed state (like a toggle switch) or if the input config has a precondition that became unsatisfied, the release event would be ignored, and the timer would keep running forever. 

Did this with Claude while learning more of the codebase.

- [x] Added StopAllHoldTimers() and hooked it to Stop();
- [x] Also added try/catch on RepeatTimer_Elapsed() and HoldTimer_Elapsed() using the pattern from inputEventExecutor.Execute() to log error instead of crashing if an event throws a fit for some reason.
- [x] tested that OnHold events are still working, with and without repeat

### DoD Checklist

- [x] Unit tests available
  - [x] .NET backend
  - [x] ~~Frontend~~
- [x] ~~Frontend tests~~
- [x] ~~i18n - All user-facing strings are translated~~
- [x] ~~documentation~~
     
## Related issues

fixes #2890

### DIFF
--- a/src/MobiFlightConnector/Base/InputEventExecutor.cs
+++ b/src/MobiFlightConnector/Base/InputEventExecutor.cs
@@ -50,6 +50,12 @@ namespace MobiFlight.Execution
             inputCache.Clear();
         }
 
+        public void StopAllHoldTimers()
+        {
+            foreach (var cfg in _configItems.OfType<InputConfigItem>())
+                cfg.button?.StopTimers();
+        }
+
         public Dictionary<string, IConfigItem> Execute(InputEventArgs e, bool isStarted)
         {
             var updatedValues = new Dictionary<string, IConfigItem>();
--- a/src/MobiFlightConnector/Base/InputEventExecutor.cs
+++ b/src/MobiFlightConnector/Base/InputEventExecutor.cs
@@ -85,6 +85,7 @@ namespace MobiFlight.Execution
                 if (!cfg.Active)
                 {
                     Log.Instance.log($"{msgEventLabel} => Skipping inactive config \"{cfg.Name}\".", LogSeverity.Warn);
+                    cfg.button?.StopTimers();
                     continue;
                 }
 
@@ -93,10 +94,14 @@ namespace MobiFlight.Execution
                     if (!CheckPreconditions(cfg))
                     {
                         Log.Instance.log($"{msgEventLabel} => Preconditions not satisfied for \"{cfg.Name}\".", LogSeverity.Debug);
+                        cfg.button?.StopTimers();
                         continue;
                     }
 
                     Log.Instance.log($"{e.Controller.Name} => Executing \"{cfg.Name}\". ({e.GetEventActionLabel()})", LogSeverity.Info);
+
+                    if (cfg.button != null)
+                        cfg.button.CanExecute = () => cfg.Active && CheckPreconditions(cfg);
 
                     cfg.RawValue = e.GetEventActionLabel();
                     cfg.Value = " ";
--- a/src/MobiFlightConnector/MobiFlight/ExecutionManager.cs
+++ b/src/MobiFlightConnector/MobiFlight/ExecutionManager.cs
@@ -633,6 +633,9 @@ namespace MobiFlight
             var configItemIndex = ConfigItems.FindIndex(i => i.GUID == item.GUID);
             if (configItemIndex == -1) return;
 
+            if (ConfigItems[configItemIndex] is InputConfigItem oldInputConfig)
+                oldInputConfig.button?.StopTimers();
+
             ConfigItems[configItemIndex] = item;
             MessageExchange.Instance.Publish(new ConfigValuePartialUpdate(item));
             OnInputConfigSettingsChanged(item, null);
--- a/src/MobiFlightConnector/MobiFlight/ExecutionManager.cs
+++ b/src/MobiFlightConnector/MobiFlight/ExecutionManager.cs
@@ -855,6 +855,8 @@ namespace MobiFlight
             joystickManager.Stop();
             midiBoardManager.Stop();
             inputActionExecutionCache.Clear();
+            foreach (var executor in _inputEventExecutors.Values)
+                executor.StopAllHoldTimers();
             ClearConfigItemStatus();
             ClearErrorMessages();
         }
--- a/src/MobiFlightConnector/MobiFlight/ExecutionManager.cs
+++ b/src/MobiFlightConnector/MobiFlight/ExecutionManager.cs
@@ -845,6 +845,8 @@ namespace MobiFlight
             mobiFlightCache.StopKeepAwake();
 
             isExecuting = false;
+            foreach (var executor in _inputEventExecutors.Values)
+                executor.StopAllHoldTimers();
 #if ARCAZE
             arcazeCache.Clear();
 #endif
@@ -855,8 +857,6 @@ namespace MobiFlight
             joystickManager.Stop();
             midiBoardManager.Stop();
             inputActionExecutionCache.Clear();
-            foreach (var executor in _inputEventExecutors.Values)
-                executor.StopAllHoldTimers();
             ClearConfigItemStatus();
             ClearErrorMessages();
         }
--- a/src/MobiFlightConnector/MobiFlight/InputConfig/ButtonInputConfig.cs
+++ b/src/MobiFlightConnector/MobiFlight/InputConfig/ButtonInputConfig.cs
@@ -23,6 +23,7 @@ namespace MobiFlight.InputConfig
 
         private Timer HoldTimer = new Timer();
         private Timer RepeatTimer = new Timer();
+        private volatile bool _holdTimersStopped = false;
 
         public ButtonInputConfig()
         {
@@ -210,6 +211,7 @@ namespace MobiFlight.InputConfig
 
         private void ExecuteRepeatOnHold()
         {
+            if (_holdTimersStopped) return;
             if (RepeatDelay > 0)
             {
                 RepeatTimer.Interval = RepeatDelay;
@@ -258,6 +260,7 @@ namespace MobiFlight.InputConfig
 
         public void StopTimers()
         {
+            _holdTimersStopped = true;
             CheckAndStopTimer();
         }
 
@@ -282,6 +285,7 @@ namespace MobiFlight.InputConfig
                                           InputEventArgs args,
                                           List<ConfigRefValue> configRefs)
         {
+            _holdTimersStopped = false;
             lock (LastOnPressLock)
             {
                 LastOnPressEvent = args;
--- a/src/MobiFlightConnector/MobiFlight/InputConfig/ButtonInputConfig.cs
+++ b/src/MobiFlightConnector/MobiFlight/InputConfig/ButtonInputConfig.cs
@@ -2,6 +2,7 @@
 using System.Timers;
 using System.Collections.Generic;
 using System.Xml.Serialization;
+using Newtonsoft.Json;
 
 namespace MobiFlight.InputConfig
 {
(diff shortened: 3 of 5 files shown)


src/MobiFlightConnector/Base/InputEventExecutor.cs
```diff
@@ -50,6 +50,12 @@ public void ClearCache()
             inputCache.Clear();
         }
 
+        public void StopAllHoldTimers()
+        {
+            foreach (var cfg in _configItems.OfType<InputConfigItem>())
+                cfg.button?.StopTimers();
+        }
```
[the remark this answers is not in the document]
tigert: Oh this is a good catch!

I'll poke at this in a separate PR so this doesn't mushroom too big for me.

src/MobiFlightConnector/MobiFlight/InputConfig/ButtonInputConfig.cs
```diff
@@ -212,9 +219,19 @@ private void ExecuteRepeatOnHold()
...
+            }
+            catch (Exception ex)
+            {
+                Log.Instance.log($"Error executing onHold action: {ex.Message}", LogSeverity.Error);
+            }
+            finally
+            {
+                HoldTimer.Stop();
+                ExecuteRepeatOnHold();
+            }
```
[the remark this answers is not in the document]
tigert: added volatile flag to mark that timers are being stopped

src/MobiFlightConnector/MobiFlight/ExecutionManager.cs
```diff
@@ -854,6 +854,8 @@ public void Stop()
             joystickManager.Stop();
             midiBoardManager.Stop();
             inputActionExecutionCache.Clear();
+            foreach (var executor in _inputEventExecutors.Values)
+                executor.StopAllHoldTimers();
```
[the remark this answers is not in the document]
tigert: good catch, moved it to the start.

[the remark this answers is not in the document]
tigert: Good one - we added a test that StopTimers is safe with no hold timer set. Adding this too.

DocMoebiuz:
@tigert good job. The case where the precondition state changes and the timer is not stopped in that case is not covered yet.

DocMoebiuz:
Just saw the comment: https://github.com/MobiFlight/MobiFlight-Connector/pull/3057/changes#r3447032204 that is exactly what i mean.

The input config executor just has to call stoptimers anytime the precondition is not satisified. that could be probably easy to add. Same for inactive configs.

DocMoebiuz (changes requested):
Have a look at the prcondition skip in the input config executor which then could call "stop all timers" on those configs. same for the active state ... then we would have fixed it right away. No additional PR required.

tigert:
> Have a look at the prcondition skip in the input config executor which then could call "stop all timers" on those configs. same for the active state ... then we would have fixed it right away. No additional PR required.

Yeah, I can have a look and amend the branch. Basically the same function there.

tigert:
> > Have a look at the prcondition skip in the input config executor which then could call "stop all timers" on those configs. same for the active state ... then we would have fixed it right away. No additional PR required.
> 
> Yeah, I can have a look and amend the branch. Basically the same function there.

OK, it was a bit more involved, because a precondition is only checked when you have an input event. And the OnHold timer by definition exists to carry on while there isn't one coming. 

I know preconditions dont "do things", they just cause a config to be ignored, but I think in this case, a "on hold" event should not continue to fire if a precondition comes unsatisfied, because the timer should honor the precondition.

DocMoebiuz:
> > > Have a look at the prcondition skip in the input config executor which then could call "stop all timers" on those configs. same for the active state ... then we would have fixed it right away. No additional PR required.
> > 
> > 
> > Yeah, I can have a look and amend the branch. Basically the same function there.
> 
> OK, it was a bit more involved, because a precondition is only checked when you have an input event. And the OnHold timer by definition exists to carry on while there isn't one coming.
> 
> I know preconditions dont "do things", they just cause a config to be ignored, but I think in this case, a "on hold" event should not continue to fire if a precondition comes unsatisfied, because the timer should honor the precondition.

I agree. If a config item is skipped because of the precondition evaluation, we should call stop on the config which then would stop all timers.

src/MobiFlightConnector/Base/InputEventExecutor.cs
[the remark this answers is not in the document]
DocMoebiuz: Let us fix it right away please.

src/MobiFlightConnector/MobiFlight/InputConfig/ButtonInputConfig.cs
```diff
@@ -23,6 +24,11 @@ public class ButtonInputConfig : IXmlSerializable, ICloneable
 
         private Timer HoldTimer = new Timer();
         private Timer RepeatTimer = new Timer();
+        private volatile bool _holdTimersStopped = false;
```
DocMoebiuz: Why do we need this?
HoldTimerStopped  could be !HoldTimer.Enabled

And why is there TimersStopped? Are there more than one HoldTimers?

src/MobiFlightConnector/MobiFlight/InputConfig/ButtonInputConfig.cs
```diff
@@ -23,6 +24,11 @@ public class ButtonInputConfig : IXmlSerializable, ICloneable
...
+
+        [XmlIgnore]
+        [JsonIgnore]
+        public Func<bool> CanExecute { get; set; }
```
DocMoebiuz, line 33: What's the reason for using a Function?
Is there a benefit with late binding or something like that? Probably because accessing the property will not immediately execute the function.

There could be a benefit to it, and that should be documented in the code next to the property.

src/MobiFlightConnector/Base/InputEventExecutor.cs
```diff
@@ -79,6 +85,7 @@ public Dictionary<string, IConfigItem> Execute(InputEventArgs e, bool isStarted)
                 if (!cfg.Active)
                 {
                     Log.Instance.log($"{msgEventLabel} => Skipping inactive config \"{cfg.Name}\".", LogSeverity.Warn);
+                    cfg.button?.StopTimers();
```
DocMoebiuz, line 88: I don't think that we are actually hitting these in the right moment.

The timer dependent events are independent from this InputEventExecutor processing.

Same is true for the Preconditions.

Registering the callback for CanExecute would actually be sufficient to prevent executing the event... but it doesn't stop the timer either.

src/MobiFlightConnector/Base/InputEventExecutor.cs
```diff
@@ -87,11 +94,15 @@ public Dictionary<string, IConfigItem> Execute(InputEventArgs e, bool isStarted)
...
                         Log.Instance.log($"{msgEventLabel} => Preconditions not satisfied for \"{cfg.Name}\".", LogSeverity.Debug);
+                        cfg.button?.StopTimers();
                         continue;
                     }
 
                     Log.Instance.log($"{e.Controller.Name} => Executing \"{cfg.Name}\". ({e.GetEventActionLabel()})", LogSeverity.Info);
 
+                    if (cfg.button != null)
+                        cfg.button.CanExecute = () => cfg.Active && CheckPreconditions(cfg);
+
```
DocMoebiuz, line 105: Add a comment


```suggestion
                    // register callback function so that onHold and onRepeat 
                    // can check if they should still execute or not
                    // onHold has only access to the button config, not the this scope here
                    if (cfg.button != null)
                        cfg.button.CanExecute = () => cfg.Active && CheckPreconditions(cfg);

```

src/MobiFlightConnector/MobiFlight/InputConfig/ButtonInputConfig.cs
DocMoebiuz: Remove. I am pretty sure this edge case is not worth the confusion.

src/MobiFlightConnector/MobiFlight/InputConfig/ButtonInputConfig.cs
```diff
@@ -198,11 +205,19 @@ private void ExecuteOnHoldAction()
...
+            }
+            catch (Exception ex)
+            {
+                Log.Instance.log($"Error executing onHold action: {ex.Message}", LogSeverity.Error);
+            }
         }
 
         private void ExecuteRepeatOnHold()
         {
+            if (_holdTimersStopped) return;
```
DocMoebiuz: Can be removed

src/MobiFlightConnector/MobiFlight/InputConfig/ButtonInputConfig.cs
```diff
@@ -239,6 +264,12 @@ private void CheckAndStopTimer()
                 RepeatTimer.Stop();
         }
 
+        public void StopTimers()
+        {
+            _holdTimersStopped = true;
```
DocMoebiuz: Can be removed

src/MobiFlightConnector/MobiFlight/InputConfig/ButtonInputConfig.cs
```diff
@@ -260,6 +291,7 @@ private void ExecuteOnPressAction(CacheCollection cacheCollection,
                                           InputEventArgs args,
                                           List<ConfigRefValue> configRefs)
         {
+            _holdTimersStopped = false;
```
DocMoebiuz: Can be removed.

src/MobiFlightConnector/MobiFlight/ExecutionManager.cs
```diff
@@ -633,6 +633,9 @@ private void HandleCommandUpdateConfigItem(ConfigItem item)
             var configItemIndex = ConfigItems.FindIndex(i => i.GUID == item.GUID);
             if (configItemIndex == -1) return;
 
+            if (ConfigItems[configItemIndex] is InputConfigItem oldInputConfig)
+                oldInputConfig.button?.StopTimers();
+
```
DocMoebiuz, line 638: This is really important.

tests/MobiFlightUnitTests/Base/InputEventExecutorTests.cs
```diff
@@ -577,6 +577,83 @@ public void Execute_AnalogInput_ExecutesSuccessfully()
...
+                button = new ButtonInputConfig
+                {
+                    onHold = new MSFS2020CustomInputAction { Command = "(>K:HoldCommand)", PresetId = "p1" }
+                }
+            };
+            _configItems.Add(configItem);
+
+            var holdTimer = GetHoldTimer(configItem.button);
+            holdTimer.Interval = 10000;
+            holdTimer.Start();
```
DocMoebiuz: I wonder why we have to direct access the timer and why we cannot trigger it with the simulated respective onPress event.

DocMoebiuz (changes requested):

DocMoebiuz (approved):
Looking good to  me!

Thanks!

PR #3066, opened 2026-06-21
Restore minimize-to-tray after badge feature

### LINKED ISSUE #3039: Minimize to Tray Gone
Minimze to tray was killed, but it is a pretty important feature. 

Please make the minimize to tray an option again
I've reverted back to version without forced taskbar status introduced in latest update.

ccrawford:
The badge PR (#2976) removed Hide() from minimizeMainForm to keep the taskbar overlay visible, but that killed the minimize-to-tray behavior beta users relied on. The overlay is moot while the form is hidden anyway -- the tray icon badge covers that case. Restore Hide() and the balloon tip, drop the now-redundant WindowState=Minimized in CheckAutoRun, and re-apply the badge on restore so the overlay re-attaches to the freshly-created taskbar button.

### Summary
The badge PR (#2976) removed Hide() from minimizeMainForm to keep the taskbar overlay visible, but that killed the minimize-to-tray behavior beta users relied on. The overlay is moot while the form is hidden anyway -- the tray icon badge covers that case. Restore Hide() and the balloon tip, drop the now-redundant WindowState=Minimized in CheckAutoRun, and re-apply the badge on restore so the overlay re-attaches to the freshly-created taskbar button.

### Screenshots / recordings
image

image
<p>Icon returned to tray with badge. Out of taskbar. Toast restored.

### Acceptance criteria
- Icon with running/stopped badge visible in taskbar if Connector on-screen.
- Icon with running/stopped badge visible in tray but not taskbar if Connector minimized.
- Toast with user reminder when Connector minimized.
- Clicking tray icon restores Connector
- Start/Stop works from Tray context menu. Badge correctly reflects state.
- Minimize on Autorun correctly returns to tray, not taskbar

### DoD Checklist
- [x] Frontend tests
  - [x] core (en,de,es)
     
## Related issues
fixes #3039

### Out-of-scope
If application is pinned to taskbar and minimized, the pinned icon WILL NOT DISPLAY A BADGE. This is Windows limitation that would require hidden windows and other janky changes to implement fully. As this is minor use case, I did not implement. The tray icon can be made visible for users who need always-visible, real-time status.

### Notes
I underestimated how much people valued minimize-to-tray! I guess lesson here is "don't change how things work just because you think your way of working is better."

### DIFF
--- a/src/MobiFlightConnector/UI/MainForm.cs
+++ b/src/MobiFlightConnector/UI/MainForm.cs
@@ -1568,7 +1568,6 @@ namespace MobiFlight.UI
                 execManager.Start();
                 if (Properties.Settings.Default.MinimizeOnAutoRun)
                 {
-                    this.WindowState = FormWindowState.Minimized;
                     minimizeMainForm(true);
                 }
             }
@@ -1820,6 +1819,10 @@ namespace MobiFlight.UI
             if (minimized)
             {
                 notifyIcon.Visible = true;
+                notifyIcon.BalloonTipTitle = i18n._tr("uiMessageMFConnectorInterfaceActive");
+                notifyIcon.BalloonTipText = i18n._tr("uiMessageApplicationIsRunningInBackgroundMode");
+                notifyIcon.ShowBalloonTip(1000);
+                this.Hide();
             }
             else
             {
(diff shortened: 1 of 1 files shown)


DocMoebiuz:
Good job! Thanks for the prompt fix.

PR #3035, opened 2026-06-13
WinCtrl CDU support for Aerosoft A340-600 Pro

Koseng:
### Summary
WinCtrl CDU support for Aerosoft A340-600 Pro

### Screenshots / recordings
IMG_0660

### DoD Checklist
- [x] documentation
  - [x] User docs (docs.mobiflight.com) -> @neilenns please add new supported aircraft when PR is merged "Aerosoft A340-600 Pro". All 3 CDUs are supported. MSFS.
  
## Related issues

fixes #3033 

### Notes
At least version V01.03 of the plane is needed.

PR #3087, opened 2026-06-29
WinCTRL CDU script for Contrail FA50

liz3:
### Summary
Adds a Python script for the Contrail Dessault Falcon 50 to the WInCTRL CDU display scripts

### Acceptance criteria
- [ ] I'm not that familiar with pythons struct API, the update check condition always fell through when comparing the byte string but succeeded when using the JSON string created from `create_mobi_json`, might be worth a fix
- [ ] I'm unsure what simvar is used for the aircraft detection in ScriptMappings(AircraftMatchPattern) which might need to be more exact. works currently though.

### Notes
I also developed the implementation in the aircraft, to implement changes there in case needed


DocMoebiuz:
@liz3 thanks for your contribution, you gotta make sure that the py-script is added to the project and copied to the output directory, otherwise it will be msising.

DocMoebiuz:
@liz3 see PR #3035 for the change in the project file

neilenns:
@liz3 So I can update the docs, does this work for captain and first officer? What simulator is this for?

src/MobiFlightConnector/Scripts/Winwing/contrail_fa50_winwing_cdu.py
```diff
@@ -0,0 +1,173 @@
...
+# --- Data Conversion ---
+def create_mobi_json(data:bytes)->str:
+    out = {"Target":"Display","Data":[[] for _ in range(MCDU_CHARS)]}
+    for r in range(FA50_CDU_ROWS):
+        if r > 0 and r < 6:
+            left,right = data[(FA50_CDU_COLUMNS * FA50_CDU_ROWS * 3) + ((r-1) * 2)], data[(FA50_CDU_COLUMNS * FA50_CDU_ROWS * 3) + ((r-1) * 2)+1]
+            row_offset = ((r-1) * 2) + 1
+            if left == 1:
+                 idxo = (MCDU_COLUMNS * (row_offset))
+                 out["Data"][idxo] = ['<', 'w', True]
```
Koseng: Size element is `False`/`True` instead of `0`/`1`, and the small-font flag is ignored (`contrail_fa50_winwing_cdu.py:104` and `:94`/`:97`)

Per the MobiFlight wire format, the third cell element is the size (`0` large, `1` small). Here it is always `False`, and `True` for the LSK arrows. At the same time, the third data byte `s` is interpreted as an amber flag (`"a" if s == 1 else …`), even though `MCDU_FLAG_SMALL_FONT = 0x01` is defined at the top (and never used). The A340 original computes the size as `int(bool(flg & MCDU_FLAG_SMALL_FONT))`. Two possibilities:

- If the third byte on the Contrail FA50 really means "highlight/amber": fine, but then delete the dead constant `MCDU_FLAG_SMALL_FONT` and send an explicit `0` (int) as the size instead of `False`.
- If the byte is a font flag like on the other add-ons, the code wrongly recolours instead of shrinking the text.

This should be verified on the device — right now the code contradicts itself.

src/MobiFlightConnector/Scripts/Winwing/contrail_fa50_winwing_cdu.py
```diff
@@ -0,0 +1,173 @@
...
+            if right == 1:
+                idxo = (MCDU_COLUMNS * (1+row_offset)) -1
+                out["Data"][idxo] = ['>', 'w', True]
+        for c in range(FA50_CDU_COLUMNS):
+            idx = (r * (FA50_CDU_COLUMNS * 3)) + (c * 3)
+            try:
+                sym, color, s = chr(data[idx]), data[idx+1], data[idx+2]
+                if sym in (" ","\0", "\n"): continue
+                idxo = (MCDU_COLUMNS * r) + (c +1)
```
Koseng, line 101: `special_chars` is never applied (`contrail_fa50_winwing_cdu.py:14`)

The table is copied 1:1 from the A340 script, but the substitution line is missing in `create_mobi_json`. The A340 original applies `sym = special_chars.get(sym, sym)` (`ini_a340_winwing_cdu.py:100`) before writing the cell — this line does not exist in the FA50 script. As a result, arrows, ballot box, degree sign etc. will show up as the letters `a`, `b`, `o`, `d`, … on the display. Fix (around line 101/104):

```python
sym = special_chars.get(sym, sym)
```

src/MobiFlightConnector/Scripts/Winwing/contrail_fa50_winwing_cdu.py
```diff
@@ -0,0 +1,173 @@
...
+                Enum.SIMCONNECT_CLIENT_DATA_REQUEST_FLAG.SIMCONNECT_CLIENT_DATA_REQUEST_FLAG_DEFAULT,0,0,0)
+            sc.register_client_data_handler(self.on_data)
+            return True
+        except Exception as e: 
+            logging.error(f"SimConnect setup failed: {e}")
+            return False
+
+    def on_data(self, d:Any):
+        if d.dwDefineID!=self.def_id or not hasattr(d,"dwData"): return
+        count=int(MCDU_DATA_SIZE)
```
Koseng: `on_data` reads and packs 4× too much data (`contrail_fa50_winwing_cdu.py:137-138`)

`count = MCDU_DATA_SIZE` is the **byte** count (670), but it is used as a **DWORD** count: `d.dwData[:670]` yields 670 DWORDs = 2680 bytes, of which only the first 670 are needed (the rest is an uninitialised region of the 8192-DWORD buffer). This runs every visual frame, per CDU. The PMDG script correctly computes `bytes / 4`. Fix:

```python
count = (MCDU_DATA_SIZE + 3) // 4   # 168 DWORDs
```

Or even simpler and faster, without the loop: `data = ctypes.string_at(d.dwData, MCDU_DATA_SIZE)`.

Koseng (changes requested):
Review done with Claude.

@liz3 " I'm unsure what simvar is used for the aircraft detection in ScriptMappings(AircraftMatchPattern) which might need to be more exact. works currently though."
-> It checks the aircraft path on the hardisk to the aircraft.cfg file. If you active logging in MobiFlight and show the logging window, you see those paths.

src/MobiFlightConnector/Scripts/Winwing/contrail_fa50_winwing_cdu.py
liz3: This is not applicable to the LSK data area

src/MobiFlightConnector/Scripts/Winwing/contrail_fa50_winwing_cdu.py
liz3: will adjust it, might not be needed

liz3: map was removed as the GNSXLS FMS does not have special characters

src/MobiFlightConnector/Scripts/Winwing/contrail_fa50_winwing_cdu.py
liz3: good catch, applying the fix to reduce the area size

liz3:
> @liz3 So I can update the docs, does this work for captain and first officer? What simulator is this for?

Yes and MSFS 2020/2024

neilenns:
Added to https://github.com/neilenns-projects/mobiflight-docs/pull/519

src/MobiFlightConnector/Scripts/Winwing/contrail_fa50_winwing_cdu.py
Koseng: Please use 0 and 1 instead of False and True for the third cell element. And the MCDU_FLAG_SMALL_FONT constant can be removed?

liz3: was addressed

Koseng (approved):
Ready to be merged.

Koseng:
@liz3 thanks a lot for the contribution

PR #3092, opened 2026-07-01
Input modifiers

### LINKED ISSUE #1203: Interpolation in Analog Input Configs
Currently interpolation is only available in output configs. It would be very convenient to have this feature in Analog Input Configs to be able to calibrate potentiometers. 

Create a new interpolation function for analog input, similar to the one in output configs. Sending the interpolated value directly to the Sim. This will allow direct calibration of potentiometer values to send to the Sim. 

Presently, to use interpolation in Analog Input requires sending the potentiometer value to an output config, using the interpolation function, to then reroute the value back to an input. 

**Additional context**
Current method is awkward, difficult for beginners and can only be considered as a workaround.  A new analog input interpolation would streamline pot calibrations.

### LINKED ISSUE #3081: Support modifiers for inputs
There is a common use case for potentiometers where you would like to transform the input value before using it in the input action. Right now you have to apply a workaround where you are writing to a MobiFlight variable first and then create an output config with an input action.

The workaround is not intuitive and it also comes with a downside: it is processing the input values on the display update timer (50ms) instead of the moment that the event actually occurs.

Define modifiers for input actions. They can use @ for the current value of the device (e.g. poti) and also have access to all config references. The action will then use the final value which is the result of the modifier chain.

workaround as described is current status quo

DocMoebiuz:
### Summary
This adds all the available modifiers to the input config wizard. Modifiers are applied to input device values.

#### Editor view
image

#### Summary view
image

### Acceptance criteria

- [x] Added input modifier UI components
- [x] Modifiers are evaluated for input values
- [x] Modifiers order can be rearranged
- [x] Summary view shows raw and final value
- [x] Auto-focus and edit is only applied for fresh-looking configs (default name and no controller assigned)

### DoD Checklist

- [x] Unit tests available
  - [x] .NET backend
- [x] Frontend tests
  - [x] core (en,de,es)
  - [x] additional langs
- [ ] documentation
  - [ ] User docs (docs.mobiflight.com) (@neilenns )
     
## Related issues

fixes #3081 
fixes #1203


neilenns:
Tracking issue for the doc changes: https://github.com/neilenns-projects/mobiflight-docs/issues/522

PR #3111, opened 2026-07-07
WingFlex DAP500 - Fix photo sensor value reading

### LINKED ISSUE #3084: WingFlex DAP500 missing analog brightness sensor input
**Describe the bug**
The WingFlex DAP500 reports its brightness/light sensor as the joystick X axis, but the device definition (`wingflex_dap500.joystick.json`) declares no Axis input (just the 20 or so buttons). As a result the axis value is parsed but can never be selected as a trigger in an Input Config. There is also an off-by-one in the HID report buffer size, at least for my DAP500 device recently purchased directly from WingFlex.

**Steps To Reproduce**
  1. Connect a WINGFLEX DAP500.
  2. Create/edit an Input Config and select it as the controller.
  3. Open the trigger device dropdown — no axis is listed. (Note the "brightness" is listed at the bottom. This is NOT the sensor but the backlight brightness setting and is reported as a separate bug)

**Expected behavior**
  The light-sensor axis should be selectable as an input trigger.

**Actual behavior**
  No axis appears; only the 20 buttons are available.

**Details**
   - MobiFlight Version: [e.g. 11.1.x / current main]
   - FlightSim: N/A
   - AddOn: N/A

**Additional context**
Adding an Axis input to `wingflex_dap500.joystick.json` (Id matching the X-axis usage) exposes the brightness axis (X axis => 48).
```    
   ,{
      "Id": 48,
      "Type": "Axis",
      "Label": "Light Sensor"
    }
```
This makes it selectable, but exposes a second issue: The joystick definition is off by one (at least for my WingFlex DAP500).  Fixing an off-by-one in Dap500Report.ToJoystickState() resolves it for me. 

Buffer byte 0 is the HID report ID, bytes 1-3 are buttons, byte 4 is the light sensor. The button loop already offsets for the prefix (startingByte = 1), but the axis read does not, so it reads the last button byte instead of the sensor byte. Change:
```
   state.X = LastInputBufferState[3];
```
to
```
  state.X = LastInputBufferState[4];
```
The sensor now functions properly.

DocMoebiuz:
### Summary
Fixes the missing photo sensor reading

### Acceptance criteria

- [x] Photo sensor reports value via Axis X

### DoD Checklist

- [x] Unit tests available
  - [x] .NET backend
     
## Related issues

fixes #3084

### DIFF
--- a/src/MobiFlightConnector/Joysticks/wingflex/wingflex_dap500.joystick.json
+++ b/src/MobiFlightConnector/Joysticks/wingflex/wingflex_dap500.joystick.json
@@ -103,6 +103,11 @@
       "Id": 20,
       "Type": "Button",
       "Label": "ALT"
+    },
+    {
+      "Id": 48,
+      "Type": "Axis",
+      "Label": "Light Sensor"
     }
   ],
   "Outputs": [
--- a/src/MobiFlightConnector/MobiFlight/Joysticks/WingFlex/Dap500Report.cs
+++ b/src/MobiFlightConnector/MobiFlight/Joysticks/WingFlex/Dap500Report.cs
@@ -72,8 +72,8 @@ namespace MobiFlight.Joysticks.WingFlex
             }
 
             // Axes
-            // Background Light Brightness
-            state.X = LastInputBufferState[3];  // Light sensor value
+            // Light sensor value
+            state.X = LastInputBufferState[4];
 
             return state;
         }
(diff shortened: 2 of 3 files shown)


DocMoebiuz:
@ccrawford can you please verify it works? I am not sure about the uint_8 -- might have to be a int_8 -- but 0-128 is a weird value range (according to their docs)

ccrawford:
> @ccrawford can you please verify it works? I am not sure about the uint_8 -- might have to be a int_8 -- but 0-128 is a weird value range (according to their docs)

Looks good on my unit. Value goes from 0 to 128 in MF, which is totally fine for resolution for a brightness sensor

DocMoebiuz:
Oh. it does work for me too! Yeah!

PR #3133, opened 2026-07-11
Fix iniBuilds A300 MCDU display mapping

### LINKED ISSUE #2917: MCDU Display for iniBuilds A300 does not work.
**Describe the bug**
MCDU Display for iniBuilds A300 does not work. It does not show anything.

**Expected behavior**
Winctrl MCDU display should reflect the MCDU display in the sim.

**Actual behavior**
WinCtrl MCDU display is not updated.

**Details**
 - MobiFlight Version: 11.0.1.2
 - AddOn: iniBuilds A300
 
**Comment from Discord User**
I found the reason why it was not working. In ScriptMappings.json it has to be:
"AircraftIdSnippet": "a300",
Not:
"AircraftIdSnippet": "A300",

neilenns:
### Summary

The match pattern for the iniBuilds A300 uses a capital `A`, but the aircraft reports a lowercase `a`.

### Screenshots / recordings
n/a

### Acceptance criteria
- [x] Verified this fix works with Discord users

### DoD Checklist
n/a
     
## Related issues
fixes #2917

### Out-of-scope
n/a

### Notes

Used a regex of `[aA]` to match both cases just in case.

### DIFF
--- a/src/MobiFlightConnector/Scripts/ScriptMappings.json
+++ b/src/MobiFlightConnector/Scripts/ScriptMappings.json
@@ -135,7 +135,7 @@
     {
       "VendorId": "0x4098",
       "ProductIds": [ "WinwingCDUs" ],
-      "AircraftMatchPattern": "A300",
+      "AircraftMatchPattern": "[aA]300",
       "ScriptName": "ini_a300_winwing_cdu.py"
     },
     {


neilenns:
Confirmed working here: https://discord.com/channels/608690978081210392/1516894873393827965/1525608240044314694

DocMoebiuz:
why do we need the regex here and is this precise enough? a300 sounds like it could match different vendors too.

neilenns:
> why do we need the regex here and is this precise enough? a300 sounds like it could match different vendors too.

It was originally written to use the uppercase A and I assume it was intentional, so I just did the regex to also capture the lowercase as reported in Discord.

I don't know why it was just "A300" before, and I don't have the plane so I can't do any validation on this. I was just trying to do an absolutely minimal change to make it work.

DocMoebiuz:
> > why do we need the regex here and is this precise enough? a300 sounds like it could match different vendors too.
> 
> It was originally written to use the uppercase A and I assume it was intentional, so I just did the regex to also capture the lowercase as reported in Discord.
> 
> I don't know why it was just "A300" before, and I don't have the plane so I can't do any validation on this. I was just trying to do an absolutely minimal change to make it work.

I understand that, but it is not ideal to provide a patch without analyzing and understanding the actual problem.

We changed from simple text comparison (.contains()) to regex. I expected to see some .lowercase() in the old code but actually could not find it. Looks like TheEgg might have the addon. We also need urgently a unit test to verify that script mappings actually resolve a provided aircraft path to the correct script.

DocMoebiuz:
Created https://github.com/MobiFlight/MobiFlight-Connector/issues/3137 as a follow up and to bring this to our coding standards.

PR #3135, opened 2026-07-11
Show live modifier chain values in the input config editor.

tigert:
### Summary

Show live modifier chain values in the input modifier editor.

This is work in progress on the UI part, just shows the values as simple "bubble" labels between modifiers that is not particularly pretty but works as a demonstration.

### Screenshots / recordings

image

Known limitations:
- needs you to save the config and then reopen the modifier editor again to see the values update, as we currently dont update configs live while editing.
- also needs MF running to get any updates at all.

The backend already evaluates the chain once per input event to compute the final value; so we capture the running value before each step of that same pass (no extra evaluation, no round trip) and ships it on the existing ConfigValueRawAndFinalUpdate message:
- ConfigItem: new ModifierInputValues, runtime-only ([JsonIgnore] so it is not persisted to the project file); carried on the value-only projection + copy ctor
- InputEventExecutor: record the value entering each Items index (active and inactive, keeping it 1:1 with Modifiers.Items) while applying only active ones
- ConfigValueOnlyItemConverter: serialize ModifierInputValues (omitted when empty)

Frontend threads the values through liveData into the modifier editor drawer and renders them as a vertical pipeline (Raw chip, between-row value chips, Final chip) in ModifierEditor; ModifierItem and ModifierSummary are unchanged.

Also consolidate the raw/final value translation strings into a generic General.Terms.*, now shared by the config table headers and the editor (old ConfigList.Header.RawValue/.FinalValue removed).

### Acceptance criteria

- [ ] X added
- [ ] Y still working
- [ ] Z changed

### DoD Checklist

- [ ] Unit tests available
  - [ ] .NET backend
  - [ ] Frontend
- [ ] Frontend tests
  - [ ] core (en,de,es)
  - [ ] additional langs
- [ ] documentation
     
## Related issues

fixes #ISSUE_NUMBER

### DIFF
--- a/src/MobiFlightConnector/Base/ConfigItem.cs
+++ b/src/MobiFlightConnector/Base/ConfigItem.cs
@@ -13,6 +13,10 @@ namespace MobiFlight.Base
         string GUID { get; set; }
         string RawValue { get; set; }
         string Value { get; set; }
+        // The live value entering each modifier in the chain, aligned 1:1 to Modifiers.Items
+        // by index. ModifierInputValues[i] is the value that enters Items[i]; index 0 is the
+        // numeric value entering the first modifier (distinct from the event-label RawValue).
+        List<string> ModifierInputValues { get; set; }
         Dictionary<ConfigItemStatusType, string> Status { get; set; }
     }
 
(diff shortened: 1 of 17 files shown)


tigert:
closing this, we had a discussion with Sebastian the other day and will likely do a different approach to this.

Issue #1747, opened 2024-03-05
[EPIC] Config Wizards with new UI framework 


DocMoebiuz:
**Is your feature request related to a problem? Please describe.**
We need a replacement for the Config Wizards that we have had in the past.

**Describe the solution you'd like**
As a user I can click on an item in the config list and then edit its details.

Confirmations:
* Edit an existing config item
  * Enable / Disable the item
  * Description
  * Add and remove Tags
  * Define and configure the trigger action
  * Define and configure modifiers
  * Define and configure Actions

PR #3147, opened 2026-07-15
Final touch up for input config wizard

### LINKED ISSUE #3148: Missing functionality: Show source of placeholder in config reference
In the old UI the config reference panel was showing both the referenced config and placeholder:

Image

Our UI style has edit functions always in the drawer panel, and the main wizard shows summary only, but this is not so great in terms of usability, since we only show the placeholder character in the summary, and there is no reference to what is what, so you need to open the config reference edit panel to verify if you don't remember.

Editor:
Image

Summary: 
Image

Maybe we should just show a simple placeholder = config in the summary labels?

```
( # = motor 2 control pwm ) ( ! = motor 2 enable ) 
( ? = Motor 2 CCW )
```

The summary view should show both the reference placeholder and the config it is using.

### LINKED ISSUE #3154: New Config Wizard Apply button should not close the wizard
**Describe the bug**
Many other software have the convention of either having these actions in dialogs:

- Cancel and OK
- Cancel, Apply, and Close

Where "Apply" does save the values being edited, but leaves the dialog open for further editing.

Right now our "Apply changes" also closes the dialog, but an actual "Apply" would be useful since we do not yet have fully "live editing" of a config, we only save the config when you close the dialog.

Intermediate saving would be useful though, since for example modifiers do not immediately update their values, they need to be saved so the backend gets the updated config. So actual "Apply" would be useful as an intermediate solution.

**Steps To Reproduce**
1. Add new input config 
2. autosense for example a potentiometer
3. add interpolate modifier
4. See modifiers summary, the values are not updating
5. Click "Apply changes"
6. Open the config again
7. See the modifier summary now shows the correct "Raw" and "Finald" values.

**Expected behavior**
It would be very useful to see the values after you have edited the modifiers. Of course we eventually want to have them show live also while editing modifiers (for example draft PR #3135) but if "Apply" would save the config to the backend while keeping the dialog open, and we would add a separate "Close" button, it would be a pretty reasonable workaround.

**Actual behavior**
lots of clicking

**Details**
 - MobiFlight Version: 11.1.0.8
 - FlightSim: [e.g. MSFS2020]

**Additional context**

### LINKED ISSUE #3155: Double click on all new wizard summaries should open edit drawer
For consistency we should open the edit function when double clicking a panel that has "edit" icon in the wizards.

Input actions already open for editing if you double click them, or press the edit button. What does not currently work:

- Preconditions
- References
- Modifiers

- If the panel opens a drawer with a list of all editable items, like modifiers, double clicking inside the panel should open the drawer. (Modifiers for example)
- If a panel has multiple, separately editable items (like Actions), those items would open in edit view when double clicked (currently input actions)

DocMoebiuz:
### Summary
Final touch ups for the input config wizard to make styles and functions aligned across the different components, like Modifier, Precondition, etc.

### Screenshots / recordings

https://github.com/user-attachments/assets/c1d70f86-5501-47f4-9491-a605ae446f6e

### Acceptance criteria

- [x] Visual styles like padding and spacing is the same across the components
- [x] Preconditions Add button has popup with types similar to Modifiers
- [x] Preconditions can be moved up and down, like Modifiers
- [x] You can close the dialog
  - [x] Without pending changes, with ESC, or clicking outside of the dialog
  - [x] With pending changes, only through Cancel (Discard) or Apply changes (Keep) 
- [x] Double Click on summary panels (reference, precondition, modifier) opens the editor

### DoD Checklist

- [x] ~~Unit tests available~~
- [x] Frontend tests
  - [x] core (en,de,es)
  - [x] additional langs
- [x] documentation
  - [x] User docs (docs.mobiflight.com) @neilenns 
     
## Related issues

relates to #1747
fixes #3148
fixes #3154
fixes #3155


tigert:
(this is a test whether this would make sense in the PR template or if it is too much)

> [!tip] 
> ### How to test a Pull Request build?
> If you would like to provide feedback on a PR build like this one, here's how:
> 1. Scroll down to the bottom of this page, to find the _last_ comment from _github-actions_
> 2. Download the MobiFlightConnector.zip file link on that comment and open it from your browser downloads
> 3. Navigate inside the zip with the Windows file manager and locate MFConnector.exe and double click it to open
> 4. Choose "Extract all" and select the default suggested destination, it will save to your Documents/MobiFlightConnector-0.0.NNNN.XXXX where NNNN is the pull request number, so you can later know which is which, if you are testing many PR builds.
>
> Now you can open the MFConnector.exe from the created folder. The PR build will have its own settings, and profile history, so it does not mess up your default MobiFlight installation.

tigert:
The zip seems to work ok for me at least.

We should swap the Preconditions and References buttons around, since the badges are in different order:

image

PR #3169, opened 2026-07-20
Duplicated config item gets selected and opens wizard for editing

### LINKED ISSUE #2832: Duplicating item doesn't select it afterwards
**Describe the bug**
Duplicating item doesn't select it afterwards

**Steps To Reproduce**
1. Duplicate item in profile list

**Expected behavior**
The new item should be added and selected

**Actual behavior**
The new item is added but not selected

**Details**
 - MobiFlight Version: 11.0.0

OK434:
### Summary

Improve add/duplicate item handling.
The table now correctly detects newly created items, selects them, scrolls them into view, and opens them for editing after both add and duplicate actions.

### Acceptance criteria

- [x] After duplicating a row, the newly created row is selected and the original row is deselected.
- [x] The newly created row is scrolled into view if it is outside the visible area.
- [x] After duplicating a row, opens for editing.
### DoD Checklist

- [x] Frontend tests

     
## Related issues

fixes #2832


src/MobiFlightConnector/frontend/src/components/tables/config-item-table/ConfigItemTableContext.ts
DocMoebiuz, line 1: Check file name. ConfigItemTableContext

src/MobiFlightConnector/frontend/src/components/tables/config-item-table/config-item-table.tsx
```diff
@@ -107,9 +108,14 @@ export function ConfigItemTable<TValue>({
   const tableRef = useRef<HTMLTableElement>(null)
   const tableBodyRef = useRef<HTMLTableSectionElement>(null)
   const prevDataLength = useRef(data.length)
```
DocMoebiuz: this is not required anymore if we store the entire data object anyways.

src/MobiFlightConnector/frontend/src/components/tables/config-item-table/config-item-table.tsx
```diff
@@ -180,39 +186,62 @@ export function ConfigItemTable<TValue>({
   })
 
   useEffect(() => {
-    if (addedItem.current && data.length === prevDataLength.current + 1) {
+    if (
+      (duplicateItem.current || addedItem.current) &&
```
DocMoebiuz: I started to prefer introducing a local variable with a speaking name for test expression elements.

Example

```
const newItemWasCreatedOrDuplicated = duplicateItem.current || addedItem.current
const listHasGrownByOneItem = data.length === prevDataLength.current + 1

if (newItemWasCreatedOrDuplicated && listHasGrownByOneItem)
```
This way you have an inline documentation and the test reads better.

src/MobiFlightConnector/frontend/src/components/tables/config-item-table/config-item-table.tsx
```diff
@@ -180,39 +186,62 @@ export function ConfigItemTable<TValue>({
...
       )
-      if (rowElement) {
-        rowElement.scrollIntoView({ behavior: "smooth", block: "center" })
-        const row = table.getRowModel().rows.find((r) => r.id === lastItem.GUID)
-        if (row) {
-          table.setRowSelection({ [row.id]: true })
+
+      const newItem = data.find((item) => !previousGuids.has(item.GUID))
+
+      if (newItem) {
```
DocMoebiuz: If possible, prefer an return early code style -> https://medium.com/swlh/return-early-pattern-3d18a41bba8

src/MobiFlightConnector/frontend/src/components/tables/config-item-table/config-item-table.tsx
```diff
@@ -289,80 +318,82 @@ export function ConfigItemTable<TValue>({
...
-                  onToggleSelected={toggleSelected}
-                />
-              </Table>
-            </div>
-          ) : (
-            <div className="border-primary flex flex-col gap-2 rounded-lg border-2 border-solid pb-6">
-              <div className="bg-primary mb-4 h-12"></div>
-              <div className="text-center" role="alert">
-                {t("ConfigList.Table.NoResultsMatchingFilter")}
+    <ConfigItemTableContext.Provider value={{ onDuplicate: handleDuplicate }}>
```
DocMoebiuz, line 314: i think it would make sense to use the same pattern when adding config items

src/MobiFlightConnector/frontend/src/components/tables/config-item-table/config-item-table.tsx
```diff
@@ -107,9 +108,14 @@ export function ConfigItemTable<TValue>({
   const tableRef = useRef<HTMLTableElement>(null)
   const tableBodyRef = useRef<HTMLTableSectionElement>(null)
   const prevDataLength = useRef(data.length)
+  const previousData = useRef(data)
   const addedItem = useRef(false)
   const showInvisibleToastOnDialogClose = useRef<string | null>(null)
+  const duplicateItem = useRef(false)
```
DocMoebiuz: So the reason to handle addItem and duplicateItem is that we want to open the edit dialog for added items but not for the duplicate items. Is that right?

DocMoebiuz: duplicatedItem

DocMoebiuz: I was actually thinking to move this state to the context and not have it in this component.

I was thinking of the context like: 
lastAction = "add" or "duplicate"

and that could already be enough to determine what to do.

DocMoebiuz (changes requested):

DocMoebiuz (approved):
Looking good now!

PR #3189, opened 2026-07-25
Don't filter preset list by project aircraft for legacy Config Wizard

DocMoebiuz:
### Summary
We have introduced the Aircraft in Project Settings - which filters the preset list using the selected aircraft as default filter.
The new Input Config Wizard also supports removing the filtering in case the user would like to see all presets.

The existing Config Wizard (old UI) does not provide an option to remove the aircraft filter, so with this PR we are reverting back to showing the full list. There are no plans to spend effort on old UI components.

### Acceptance criteria

- [x] All presets are displayed no matter what aircraft is defined in Project Settings.
     
## Related issues

part of #3078 
part of #2819


OK434:
Looks good to me. The change makes sense and the implementation is clear.

PR #3197, opened 2026-07-28
Fix stale precondition status after deletion

### LINKED ISSUE #3156: Deleting a precondition from config does not clear the precondition status icon (purple exclamation mark)
**Describe the bug**
when you delete a precondition and save the config while a precondition is unsatisfied, the precondition flag is not cleared from the config table row.

The actual deletion of the precondition works correctly, the error is only in the status icon that does not get cleared when a config changes, it will show until you stop and start MobiFlight.

**Steps To Reproduce**
1. Add precondition to a config and close the wizard
2. Make precondition unsatisfied
3. Edit config, delete precondition while it is unsatisfied, and close wizard
4. Note how unsatisfied icon is still visible

**Expected behavior**
The flag should be re-evaluated when config is modified, so it should show only if the precondition actually is still unsatisfied and exists.

**Actual behavior**
Seems the flags are only cleared when config Stops or Starts.

**Details**
 - MobiFlight Version: 11.1.0.8
 - FlightSim: INOP
 - AddOn: INOP

**Additional context**

Guflly:
### Summary

Removing the last precondition now clears the stale precondition status and queues the config update, so the icon disappears without restarting MobiFlight.

### Acceptance criteria

- [x] Removing the last precondition clears the old status
- [x] The config update is sent to the frontend
- [x] Existing precondition behavior still passes the full test suite

### DoD Checklist

- [x] Unit tests available
  - [x] .NET backend
- [x] Frontend production build

## Related issues

fixes #3156

### Notes

1,057 tests passed and 31 existing tests were skipped.

### DIFF
--- a/src/MobiFlightConnector/MobiFlight/ExecutionManager.cs
+++ b/src/MobiFlightConnector/MobiFlight/ExecutionManager.cs
@@ -1483,24 +1483,20 @@ namespace MobiFlight
 
                     var originalCfg = cfgItem.Clone() as InputConfigItem;
 
-                    // if there are preconditions check and skip if necessary
-                    if (cfg.Preconditions.Count > 0)
+                    ConnectorValue currentValue = new ConnectorValue();
+
+                    if (!PreconditionChecker.CheckPrecondition(cfg, currentValue, ConfigItems, arcazeCache, mobiFlightCache))
                     {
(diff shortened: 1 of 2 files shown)


DocMoebiuz (approved):
Thanks!

PR #3198, opened 2026-07-29
Allow PR firmware versions during firmware update check

### LINKED ISSUE #3191: Unable to upload configs to PR builds of firmware
**Describe the bug**
PR builds of MF firmware have `0.0.PRNUMBER` firmware version. However if you try to upload a config to a board with such firmware, Connector refuses because it flags the firmware as "out of date".

Image

**Steps To Reproduce**
1. Upload a PR build firmware to a board
2. Change device settings
3. Click "Upload config"

**Expected behavior**
PR builds should be treated specially (0.0.* pattern in firmware is not something you encounter in the wild with stable or beta firmwares)

We could of course allow a "I know what I am doing, upload anyway" option, but we force users to upgrade the firwmare in sync with MobiFlight version to avoid incompatibility issues, so it would be a bad idea to create an excuse to not upgrade.

**Actual behavior**
Unable to upload configs to devices with PR firmware.

**Details**
 - MobiFlight Version: 11.1.0.10 (BETA)
 - FlightSim: [e.g. MSFS2020]

**Additional context**

OK434:
### Summary

Allow PR firmware (`0.0.<PR_NUMBER>`) to bypass the outdated firmware check so configuration uploads work correctly for pull request firmware builds.
### Acceptance criteria

- [x] Unable to upload configs to devices with PR firmware.

### DoD Checklist

 - [x] .NET backend

     
## Related issues

fixes #3191

### DIFF
--- a/src/MobiFlightConnector/MobiFlight/MobiFlightModuleInfo.cs
+++ b/src/MobiFlightConnector/MobiFlight/MobiFlightModuleInfo.cs
@@ -49,10 +49,10 @@ namespace MobiFlight
                 return false;
             }
             // and update when version lower than latest
-            return true ;
-                
-                
-               
+            return currentVersion.CompareTo(latestVersion) < 0;
+
+
+
         }
 
     }
--- a/src/MobiFlightConnector/MobiFlight/MobiFlightModuleInfo.cs
+++ b/src/MobiFlightConnector/MobiFlight/MobiFlightModuleInfo.cs
@@ -40,18 +40,19 @@ namespace MobiFlight
             {
                 currentVersion = new Version("0.0.0");
             }
-            bool isSpecialFirmware =
+            bool isDevOrPrFirmware =
                 currentVersion.Major == 0 &&
                 currentVersion.Minor == 0;
             // PR firmware uses version 0.0.<PR_NUMBER>, so don't require an update.
-            if (isSpecialFirmware)
+            if (isDevOrPrFirmware)
             {
                 return false;
             }
-            return (
-                // and update when version lower than latest
-                currentVersion.CompareTo(latestVersion) < 0
-                );
+            // and update when version lower than latest
+            return true ;
+                
+                
+               
         }
 
     }
--- a/src/MobiFlightConnector/MobiFlight/MobiFlightModuleInfo.cs
+++ b/src/MobiFlightConnector/MobiFlight/MobiFlightModuleInfo.cs
@@ -31,6 +31,7 @@ namespace MobiFlight
         {
             Version latestVersion = new Version(Board.Info.LatestFirmwareVersion);
             Version currentVersion;
+
             try
             {
                 currentVersion = new Version(Version != null ? Version : "0.0.0");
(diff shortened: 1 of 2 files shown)


src/MobiFlightConnector/MobiFlight/MobiFlightModuleInfo.cs
```diff
@@ -39,6 +40,15 @@ internal bool FirmwareRequiresUpdate()
...
+                currentVersion.Minor == 0 &&
+                currentVersion.Build > 1;
+            // PR firmware uses version 0.0.<PR_NUMBER>, so don't require an update.
+            if (isPrFirmware)
+            {
+                return false;
+            }
             return (
                 // ignore the developer board that has 0.0.1
                 currentVersion.CompareTo(new Version("0.0.1")) != 0 &&
```
DocMoebiuz: That's a good start!

I like that you are giving it a descriptive local variable!

If you look at line 52, you will see that there is also a similar test. And compared to your approach, it is not as readable. But it also checks for 0.0.1

So 0.0.1 is not only a PR version of the firmware, it is the one that a developer uses. So basically, similar.

Can you spot the optimization?

DocMoebiuz (changes requested):
Also don't forget the unit test.

src/MobiFlightConnector/MobiFlight/MobiFlightModuleInfo.cs
```diff
@@ -40,20 +40,18 @@ internal bool FirmwareRequiresUpdate()
...
                 return false;
             }
             return (
-                // ignore the developer board that has 0.0.1
-                currentVersion.CompareTo(new Version("0.0.1")) != 0 &&
                 // and update when version lower than latest
-                currentVersion.CompareTo(latestVersion) < 0);
+                currentVersion.CompareTo(latestVersion) < 0
+                );
         }
```
DocMoebiuz: Exactly! You got it.

I suggest to name the variable "isDevOrPrFirmware" to make it really explicit.

and you don't need the return () anymore for the last return

DocMoebiuz (changes requested):

tests/MobiFlightUnitTests/MobiFlight/MobiFlightModuleInfoTests.cs
```diff
@@ -36,5 +36,28 @@ public void FirmwareInstallPossibleTest()
             info.Board = board;
             Assert.IsTrue(info.FirmwareInstallPossible());
         }
+        [TestMethod()]
```
DocMoebiuz, line 40: The method name should contain the method name under test and the different cases

eg.

FirmwareRequiresUpdate_TrueForOldVersion 

FirmwareRequiresUpdate_IgnoresDevAndPrBuilds

and you got one test case missing. can you guess it?

DocMoebiuz (changes requested):

DocMoebiuz (changes requested):
Great! There is still one case which we haven't covered by a unit test.

DocMoebiuz (approved):

PR #3199, opened 2026-07-29
Support floats for all internal number representation

### LINKED ISSUE #1832: Support floats for all internal number representation
MobiFlight handles results of transformations and many internal values as integers. This is more an historical reason rather than a design requirement.

Support floats for all internal number representations.

**Additional context**
 For most calculations like modifieres this should already be the case. But some inputActions or output devices should work with floats rather than using integer values.
 
relates to #1827
relates ro #1571
relates to #1234

OK434:
### Summary

This PR fixes the handling of floating-point values for Input Actions. Decimal values are now preserved instead of being treated as non-numeric and converted to 0. A regression test was added to verify the new behavior.

### Acceptance criteria

  - [x] Support floats for all internal number representations.

### DoD Checklist

- [x] Unit tests available
  - [x] .NET backend

## Related issues

fixes #1832

### DIFF
--- a/src/MobiFlightConnector/MobiFlight/Modifier/Interpolation.cs
+++ b/src/MobiFlightConnector/MobiFlight/Modifier/Interpolation.cs
@@ -105,7 +105,7 @@ namespace MobiFlight.Modifier
             {
                 case FSUIPCOffsetType.Float:
                 case FSUIPCOffsetType.Integer:
-                    result.Float64 = Value(connectorValue.Float64);
+                    result.Float64 = CalculateInterpolatedValue(connectorValue.Float64);
                     break;
 
                 case FSUIPCOffsetType.String:
@@ -122,13 +122,13 @@ namespace MobiFlight.Modifier
 
             if (Count > 0 && float.TryParse(strValue, out float value))
             {
-                result = Value(value).ToString();
+                result = CalculateInterpolatedValue(value).ToString();
             }
 
             return result;
         }
 
-        public double Value(double x)
+        public double CalculateInterpolatedValue(double x)
         {
             double first = Values.Keys.First();
             if (x <= first) return Values[first];
--- a/src/MobiFlightConnector/MobiFlight/Modifier/Interpolation.cs
+++ b/src/MobiFlightConnector/MobiFlight/Modifier/Interpolation.cs
@@ -119,9 +119,10 @@ namespace MobiFlight.Modifier
         public string Apply(string strValue)
         {
             string result = strValue;
-            if (Count > 0)
+
+            if (Count > 0 && float.TryParse(strValue, out float value))
             {
-                result = Value(float.Parse(strValue)).ToString();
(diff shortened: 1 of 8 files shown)


src/MobiFlightConnector/MobiFlight/Modifier/Interpolation.cs
```diff
@@ -121,7 +121,7 @@ public string Apply(string strValue)
             string result = strValue;
             if (Count > 0)
             {
-                result = Math.Round(Value(float.Parse(strValue)), 0).ToString();
+                result = Value(float.Parse(strValue)).ToString();
             }
 
             return result;
```
DocMoebiuz, line 128: these changes are not covered by unit tests yet

src/MobiFlightConnector/MobiFlight/Modifier/Interpolation.cs
```diff
@@ -121,7 +121,7 @@ public string Apply(string strValue)
             string result = strValue;
             if (Count > 0)
             {
-                result = Math.Round(Value(float.Parse(strValue)), 0).ToString();
+                result = Value(float.Parse(strValue)).ToString();
```
DocMoebiuz: A lot of the old code just uses this `.Parse`

newer code use .TryParse to properly handle the case when we cannot parse the number. 

I suggest we use TryParse for all the code we touch

DocMoebiuz (changes requested):

DocMoebiuz:
I also found one Math.Round for FSUIPC for the 8-byte values which are actually typed as Float and where we are setting Float64 value. that actually doesn't make sense.

tigert:
Ooh, awesome to get this fixed!

src/MobiFlightConnector/MobiFlight/Modifier/Interpolation.cs
DocMoebiuz: Unrelated to your changes, but when reviewing the pull request I really stumbled over this method's name.

`Value()`

I think it would be helpful to have a better name:
`CalculateInterpolatedValue` for example.

DocMoebiuz (changes requested):
I'd suggest renaming that one method. Everything else looks good to me. Good job!

DocMoebiuz (approved):

JaimeLeon2:
Does this fix include interpolation modifier input and output configs, both left and right side of interpolation table?

DocMoebiuz:
> Does this fix include interpolation modifier input and output configs, both left and right side of interpolation table?

what do you mean?

JaimeLeon2:
> > Does this fix include interpolation modifier input and output configs, both left and right side of interpolation table?
> 
> what do you mean?

It was recently talked in Discord that the latest release of Mobiflight with the interpolation input modifier cannot handle floating point numbers. 
Hence my question if this floating point fix includes this interpolation modifier limitation.

PR #3219, opened 2026-08-05
Add support for additional keyboard key codes

### LINKED ISSUE #3207: error KeyInputAction: Invalid key code: Digit8 since 11.2.0
**Describe the bug**
Since update to 11.2.0 it is no longer possible to use KeyInputAction. Loading profile shows empty Key combo. Setting it again will result in the following error during execution:

Image

Image

Image

error KeyInputAction: Invalid key code: Digit8

Unfortunately, falling back to 11.1.0.2 is not possible, because I already edited the profile, which made in no longer compatible with older version.

**Steps To Reproduce**
1. Load a profile with KeyInputAction
2. Check setting of the input configuration
3. Fix "none" issue (lost profile information)
4. Press input button on device
5. error appears

**Expected behavior**
This was working in at least 11.1.0.2
Also, when upgrading profiles I excpect that a backup is created with, before MobiFlight makes it incompatible with older versions. Or at least, a big red warning for the user to create a manual backup and that the profile will be updated with manual confirmation by the user before upgrading profiles to incompatible version. With this behavior I lost my whole fallback.

**Actual behavior**
KeyInputAction is not executed and also the key mapping is lost after loading profile. It just throws the error.

**Details**
 - MobiFlight Version: 11.2.0
 - FlightSim: MSFS2024
 - AddOn: FBW A320, but I use these KeyInputAction for all

**Additional context**

OK434:
### Summary

Fixes keyboard input mappings for `KeyboardEvent.code` values introduced in newer MobiFlight profiles.

Added support for:
- `Digit0`–`Digit9`
- `Numpad0`–`Numpad9`
- `NumpadAdd`
- `NumpadSubtract`
- `NumpadMultiply`
- `NumpadDivide`
- `NumpadDecimal`
- `NumpadEnter`
- `IntlBackslash`

### Acceptance criteria

- [x] Added mappings for `Digit0`–`Digit9`
- [x] Added mappings for common `Numpad` key codes
- [x] Added mapping for `IntlBackslash`

### DoD Checklist

- [x] Unit tests available
  - [x] .NET backend
 
## Related issues

fixes #3207


tests/MobiFlightUnitTests/MobiFlight/InputConfig/KeyInputActionTests.cs
```diff
@@ -165,5 +165,68 @@ public void JsonDeserialize_WithNewCodeTest_BackslashKey()
...
+
+            var action = Newtonsoft.Json.JsonConvert.DeserializeObject<KeyInputAction>(jsonString);
+
+            Assert.AreEqual("Digit1", action.Code);
+            Assert.IsTrue(action.Alt);
+            Assert.IsFalse(action.Control);
+            Assert.IsFalse(action.Shift);
+
+            Assert.AreEqual(Keys.D1, KeyInputAction.MapCodeToKeys(action.Code));
+        }
```
DocMoebiuz: the name of the unit tests suggests that we are validating json deserialization.
In fact, we are validating MapCodeToKeys though.

I'd suggest to rename the methods and remove the serialization part, you can simply initialize the InputAction directly.

DocMoebiuz (changes requested):

tigert:
Numbers work now. Also many other keys, nice work!

These are still strange and feel like they come from a wrong keymap or something. This is finnish/swedish layout.

image

- `§` gives "backquote"
- ´+´ gives "minus"
- `´` gives "equal"
- `umlaut dots` gives "bracketright"
- `'` gives _"backslash"
- and `- gives "slash`

tigert:
To me those "wrong" ones look like they come from the US keymap, except for ´§´ which is an oddity.

image

But my example swedish keymap picture is not exactly what this is either - since it has no AltGr that I use to do for example $ (AltGr + 4)

OK434:
I don't think we can realistically add every possible key mapping. There are many layout-specific keys, and some don't have a direct equivalent in System.Windows.Forms.Keys. I added support for the reported regression and the most common key codes, but if additional unmapped keys are reported we can add them as needed. @tigert

DocMoebiuz:
the mechanism is using the OEM key codes on purpose. You wont get necessarily the right label of your key displayed. But it will work nevertheless. Every key has a code that is independent from your localized layout.

And in practice it should not matter or we will have to revert to the old method entirely.

Right now the keyboard mechanism works like:
I press this key. And something happens.
it is not.

I press $ and something happens 

Maybe my assumption was wrong about this. Because you might wanna look up the associated key, eg. in case you forgot

DocMoebiuz:
> To me those "wrong" ones look like they come from the US keymap, except for ´§´ which is an oddity.
> 
> image
> But my example swedish keymap picture is not exactly what this is either - since it has no AltGr that I use to do for example $ (AltGr + 4)

Your picture doesnt show a typical US Layout

DocMoebiuz (approved):
Thanks @OK434

PR #3225, opened 2026-08-06
Virtualize the HubHop preset lists in the config wizard

### LINKED ISSUE #3223: Preset list renders all 25,000+ HubHop presets with no virtualization — 9 s to open an MSFS action editor, 2 s to close
### Describe the bug

Opening the action editor for an input config (Actions → edit) takes **9–10 seconds**, and
going back takes **2 seconds**. During that time the whole UI is unresponsive.

The cause is visible in the dialog itself: the preset picker reports
**`25152 preset(s) found`**, and `MsfsPresetPanel` renders every one of them into the DOM.
The list lives in a `ScrollArea` of fixed height `h-56` — about five rows are visible — so
~25,000 `PresetListItem` elements, each ~5 nested `div`s, are built to display five.

The asymmetry (9 s to open, 2 s to close) matches DOM construction vs. teardown; it is not
data loading. `useQuery` caches the preset file with `staleTime: Infinity`, so repeated
opens skip the fetch entirely and are still slow.

### Steps to reproduce

1. Open a project containing an input config whose MSFS action uses a **Custom Preset**
   (i.e. a hand-written code such as `34804 (>K:ROTOR_BRAKE)`, not a preset picked from
   the HubHop library).
2. Open the config item for editing.
3. In Actions, click the edit (pencil) icon on the action.
4. Wait for "Step 2 - Select preset" to appear.
5. Scroll the preset list.

Adding a brand-new MSFS action to any input config reproduces it just as well — no preset
is selected, so the list is unfiltered.

### Expected behavior

The editor opens promptly regardless of how many presets exist in the library.

### Actual behavior

9–10 s to open, 2 s to go back, UI frozen throughout. Once open, **scrolling the preset
list is also visibly choppy** — expected with ~25,000 items in the DOM, each carrying a
`hover:` style that forces style recalculation over the whole subtree on pointer movement.

**Typing in the "Step 3 – Customize preset" Code field is also laggy**, one keystroke at a
time. That field has nothing to do with the preset list, but each keystroke updates draft
state in the parent, which re-renders `MsfsPresetPanel`, which re-runs all five filter
passes and re-reconciles all 25,152 list items. `PresetListItem` is a `forwardRef`
component with no `memo`, so none of that work can be skipped.

### Why the worst case is the normal case

The filter state is seeded from the currently selected preset:

```ts
const [filter, setFilter] = useState({
  vendor: selectedPreset?.vendor || "",
  aircraft: selectedPreset?.aircraft || "",
  system: selectedPreset?.system || "",
  search: "",
})
```

`selectedPreset` comes from `validPresets.find((p) => p.id === selectedPresetId)`. When it
resolves, the initial filter narrows the list to a few hundred rows and the dialog opens
quickly. When it does **not** resolve, every filter is empty and all 25,152 rows are
rendered.

It fails to resolve in three situations, and none of them is unusual:

1. **The user wrote their own RPN code.** Step 3 of the dialog is titled "Customize preset
   (optional)" with the hint "This preset code is used with your simulator. You can edit it
   manually" — writing custom codes is an intended, documented workflow, not a workaround.
   Many people build entire cockpits this way and never pick anything from the HubHop
   library at all.
2. **The preset was removed or renamed upstream in HubHop.** The stored `PresetId` stays in
   the project, stops resolving after a library update, and the dialog shows *"You
   customized this preset or it is no longer available."* The user did nothing; their
   config simply became slow to edit over time.
3. **A brand-new action is being created**, where nothing is selected yet.

In the project used for this report, 211 of the actions are `MSFS2020CustomInputAction`.
Every one of them that falls into the cases above pays the full ten seconds, both to open
and to edit.

This is also why the problem reads as random. Some config items open instantly and others
hang for ten seconds, with nothing visibly different about them — the difference is whether
a preset id happens to still resolve against the current library.

### Environment

- MobiFlight version: 11.2.0
- WebView2 Runtime: 151.0.4129.59
- OS: Windows 11 25H2, build 26200.8457
- CPU: AMD Ryzen 7 5700X3D (8C/16T)
- Simulator: MSFS 2024
- Presets in library: 25,152 (as reported by the dialog)
- Config items in project: **270** across 5 config files (116 / 80 / 55 / 11 / 8);
  177 input, 93 output; 255 active. 211 of the actions are `MSFS2020CustomInputAction`.

CPU affinity note: measured with all 16 logical processors available to every MobiFlight
process, verified before the test. No CPU tuning software was active during the
measurement.

### Where the cost is

All references are to `main` at the time of writing.

**1. The preset list is not virtualized.**
[`MsfsPresetPanel.tsx#L256`](https://github.com/MobiFlight/MobiFlight-Connector/blob/main/src/MobiFlightConnector/frontend/src/components/wizard/components/InputActions/MsfsPresetPanel.tsx#L256)
maps the entire `filteredPresets` array into `PresetListItem` elements inside a fixed-height
`ScrollArea`. `@tanstack/react-virtual` is already a dependency of the project.

**2. Five full passes over the preset array per render, none memoized.**
[`#L66`](https://github.com/MobiFlight/MobiFlight-Connector/blob/main/src/MobiFlightConnector/frontend/src/components/wizard/components/InputActions/MsfsPresetPanel.tsx#L66)
`validPresets`,
[`#L78`](https://github.com/MobiFlight/MobiFlight-Connector/blob/main/src/MobiFlightConnector/frontend/src/components/wizard/components/InputActions/MsfsPresetPanel.tsx#L78)
`filteredPresets`,
[`#L85`](https://github.com/MobiFlight/MobiFlight-Connector/blob/main/src/MobiFlightConnector/frontend/src/components/wizard/components/InputActions/MsfsPresetPanel.tsx#L85)
`filteredCategoryPresets`,
[`#L91`](https://github.com/MobiFlight/MobiFlight-Connector/blob/main/src/MobiFlightConnector/frontend/src/components/wizard/components/InputActions/MsfsPresetPanel.tsx#L91)
`filteredVendorPresets`,
[`#L97`](https://github.com/MobiFlight/MobiFlight-Connector/blob/main/src/MobiFlightConnector/frontend/src/components/wizard/components/InputActions/MsfsPresetPanel.tsx#L97)
`filteredAircraftPresets` — followed by three `[...new Set(...)].sort()` builds at
[`#L103-L120`](https://github.com/MobiFlight/MobiFlight-Connector/blob/main/src/MobiFlightConnector/frontend/src/components/wizard/components/InputActions/MsfsPresetPanel.tsx#L103-L120).
There is no `useMemo` anywhere in the file, so every keystroke in the text filter re-runs
all of it.

**3. `filterPresetByText` allocates per preset, per pass.**
[`lib/configWizard.ts`](https://github.com/MobiFlight/MobiFlight-Connector/blob/main/src/MobiFlightConnector/frontend/src/lib/configWizard.ts)
builds a fresh concatenated, lower-cased haystack string for every preset on every call —
roughly 100k string allocations per render across the four filtered arrays.

**4. `XplanePresetPanel.tsx` has the same structure** (`#L75`, `#L90`, `#L98`, `#L105`,
`#L112`, `#L273`), so any fix should cover both.

**5. Minor: the "Project aircraft only" toggle silently does nothing when the project has
no aircraft list.** `projectAircraftFilter` returns `true` for every preset when
`project?.Aircraft?.length` is 0, so the switch appears active while filtering nothing.
Worth either disabling the control or explaining why it is inert.

### Suggested fixes, cheapest first

1. Wrap the five filtered arrays and the three category/aircraft/vendor lists in `useMemo`
   keyed on `presets` + `filter` + `favoritesOnly`.
2. Virtualize the preset list with `@tanstack/react-virtual` (already a dependency) — the
   container is fixed-height, so this is a natural fit.
3. `memo` `PresetListItem` and give `setSelectedPreset` a stable identity, so unrelated
   state changes elsewhere in the dialog stop re-reconciling the list.
4. Precompute the search haystack once per preset when the query resolves, instead of
   rebuilding it inside `filterPresetByText` on every pass.
5. Debounce the text filter input.
6. Apply the same to `XplanePresetPanel.tsx`.
7. Grey out or hide "Project aircraft only" when the project has no aircraft list.

Two more that would help even without the above:

8. **Persist the preset's `vendor` / `aircraft` / `system` alongside `PresetId` in the
   config item.** Then a customized or since-removed preset can still seed the filter, and
   the list stays narrow instead of falling back to all 25,152 rows. This also makes the
   *"you customized this preset or it is no longer available"* state far less punishing.
9. **Do not render the full library when nothing is selected.** For a new action, showing
   an empty list with a "type to search" hint would be both faster and easier to use than
   25,152 unfiltered rows of which five are visible.

Happy to test a build with any subset of the above.

chegevarko1982:
### Summary
Opening an MSFS action editor took about nine seconds, and closing it about
two. The preset panel rendered the entire HubHop library into a 224px tall
scroll area — 25,152 rows for the input list, 6,977 for the output one — so
every open built tens of thousands of DOM nodes that were never visible. The
X-Plane panel does the same with 7,960 and 7,085 rows.

The lists are now virtualized with @tanstack/react-virtual, which is already
a dependency. At most 21 rows exist at a time: about five fit in the viewport,
plus eight rows of overscan on each side. Opening and closing the editor is
immediate, scrolling is smooth, and typing in the Code field no longer stalls
on each character.

Three smaller changes come along, each of which was also costing a full pass
over the library on every render:

- The passes that build the filtered lists and the vendor/aircraft/system
  dropdowns are memoized instead of re-running on every render.
- The text filter no longer rebuilds a lowercased concatenation per preset per
  call; it caches one per preset in a WeakMap and short-circuits when the
  filter is empty.
- PresetListItem is memoized, and the panel gives it a callback with a stable
  identity so the memo is effective without requiring changes in callers.

One detail worth flagging for review: getItemKey is deliberately wrapped in
useCallback. The virtualizer keeps it in a memo dependency list, compares it
by reference, and notifies React from that memo's onChange — and setOptions
runs during render. An inline arrow there schedules a render from inside a
render on every pass, which React ends with "Too many re-renders". The key is
derived from the preset id rather than the row index so cached row
measurements follow the preset when the filter changes; ids are unique across
both preset files, codes are not.

### Screenshots / recordings
No visual change is intended — same layout, same styled ScrollArea, same row
markup.

### Acceptance criteria
- [x] MSFS preset list renders only the visible rows
- [x] X-Plane preset list renders only the visible rows
- [x] Selecting a preset still fills in the code and the Step 3 details
- [x] Opening a config with a valid preset still scrolls to it and highlights it
- [x] Vendor / aircraft / system / text filters and "Reset filters" behave as before
- [x] "Project aircraft only" behaves as before
- [x] Rows with a two-line description are not clipped or overlapped

### DoD Checklist
- [ ] Unit tests available
  - [ ] .NET backend
  - [ ] Frontend
- [ ] Frontend tests
  - [x] core (en,de,es)
  - [x] additional langs
- [ ] documentation

## Related issues
fixes #3223

### Out-of-scope
The report also suggests storing vendor/aircraft/system next to the PresetId,
and not showing the whole library when nothing is selected. Both change the
project format or the UX, and virtualization removes the performance reason
for them, so they are left out of this PR.

### Notes
Backend untouched; no new dependencies. Built against 11.2.0 and verified in a
real install with a 270 config item project before being rebased onto main.

### DIFF
--- a/src/MobiFlightConnector/frontend/src/components/ui/scroll-area.tsx
+++ b/src/MobiFlightConnector/frontend/src/components/ui/scroll-area.tsx
@@ -5,14 +5,19 @@ import { cn } from "@/lib/utils"
 
 const ScrollArea = React.forwardRef<
   React.ElementRef<typeof ScrollAreaPrimitive.Root>,
-  React.ComponentPropsWithoutRef<typeof ScrollAreaPrimitive.Root>
->(({ className, children, ...props }, ref) => (
+  React.ComponentPropsWithoutRef<typeof ScrollAreaPrimitive.Root> & {
+    viewportRef?: React.Ref<HTMLDivElement>
+  }
+>(({ className, children, viewportRef, ...props }, ref) => (
   <ScrollAreaPrimitive.Root
     ref={ref}
     className={cn("relative overflow-hidden", className)}
     {...props}
   >
-    <ScrollAreaPrimitive.Viewport className="absolute inset-0 h-full w-full rounded-[inherit]">
+    <ScrollAreaPrimitive.Viewport
+      ref={viewportRef}
+      className="absolute inset-0 h-full w-full rounded-[inherit]"
+    >
       {children}
     </ScrollAreaPrimitive.Viewport>
     <ScrollBar />
--- a/src/MobiFlightConnector/frontend/src/components/wizard/components/InputActions/MsfsPresetPanel.tsx
+++ b/src/MobiFlightConnector/frontend/src/components/wizard/components/InputActions/MsfsPresetPanel.tsx
@@ -12,13 +12,26 @@ import { Preset, XplanePreset } from "@/types/preset"
 import { AircraftInfo } from "@/types/project"
 import { IconX } from "@tabler/icons-react"
 import { useQuery } from "@tanstack/react-query"
-import { useCallback, useEffect, useRef, useState } from "react"
+import { useVirtualizer } from "@tanstack/react-virtual"
+import {
+  useCallback,
+  useDeferredValue,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react"
 import { useTranslation } from "react-i18next"
 export type MsfsPresetPanelProps = {
   variant: "input" | "output"
   selectedPresetId: string | null
   setSelectedPreset: (preset: Preset | XplanePreset | null) => void
 }
+
+// Outside the component so it isn't a fresh value on every render (it is a
+// dependency of scrollActivePresetIntoView below).
+const SCROLL_INTO_VIEW_TIMEOUT = 800
+
 const MsfsPresetPanel = ({
   variant,
   selectedPresetId,
@@ -27,8 +40,7 @@ const MsfsPresetPanel = ({
   const { t } = useTranslation()
   const { project } = useProjectStore()
   const [favoritesOnly, setFavoritesOnly] = useState(true)
-  const SCROLL_INTO_VIEW_TIMEOUT = 800
-  const refActiveElement = useRef<HTMLDivElement | null>(null)
+  const viewportRef = useRef<HTMLDivElement | null>(null)
   const scrollTimeoutRef = useRef<number | null>(null)
   const cancelScrollIntoView = () => {
     if (scrollTimeoutRef.current !== null) {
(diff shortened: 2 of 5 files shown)


DocMoebiuz:
@chegevarko1982 when i checkout your branch and run  it locally on my computer... it shows an error message as soon as i select the MSFS Presets Input Action Type

DocMoebiuz (changes requested):
Please rebase with main and force push again.
Then I will review. Right now, it doesnt work for me locally and the playwright workflow has not been triggered.

DocMoebiuz:
implemented in https://github.com/MobiFlight/MobiFlight-Connector/pull/3307

Thanks for your efforts though!

PR #3231, opened 2026-08-09
Add external LED control for FreeJoy boards

### LINKED ISSUE #3230: Support external LED control for FreeJoy boards

FreeJoy is an open source STM32 based joystick/button box firmware used by many home cockpit builders. MobiFlight already detects FreeJoy boards as regular DirectInput controllers, so buttons and axes work fine. Their LEDs however cannot be driven from MobiFlight, even though the firmware supports exactly that through a documented "external LED control" mode.

Today a FreeJoy LED can only be tied to a physical button on the board itself, so panel annunciators cannot follow simulator state.

Native support for FreeJoy external LED control, so the LEDs show up as output devices in MobiFlight and can be bound to sim variables like any other output.

The protocol is documented in the FreeJoy wiki: https://github.com/FreeJoy-Team/FreeJoyWiki/blob/master/rus/%D0%92%D0%BD%D0%B5%D1%88%D0%BD%D0%B5%D0%B5-%D1%83%D0%BF%D1%80%D0%B0%D0%B2%D0%BB%D0%B5%D0%BD%D0%B8%D0%B5-%D1%81%D0%B2%D0%B5%D1%82%D0%BE%D0%B4%D0%B8%D0%BE%D0%B4%D0%B0%D0%BC%D0%B8.md

The board exposes a second, vendor specific HID interface (Usage Page `0xFF00`) which accepts an Output Report with Report ID 6 carrying a 4 byte little-endian bitmask, one bit per LED, for up to 24 LEDs. Only LEDs flagged "External" in the FreeJoy Configurator follow this bitmask, the rest keep using the board's own button logic.

- The generic LED mechanism in the `Joystick` base class. It does not fit: it sends a Feature Report rather than an Output Report, and it resolves the device by vendor/product id alone, which is ambiguous here because the board exposes both its joystick interface and its LED interface under the same ids.
- Driving the LEDs from a separate MobiFlight compatible board. This works, but it needs extra hardware and duplicate wiring for panels that already have their LEDs wired to FreeJoy.

**Additional context**

I have a working implementation that follows the existing WinCtrl controller pattern, and I have verified it on hardware (FreeJoy v1.7.1, VID `0x0483` / PID `0x5757`, MSFS). Happy to open a PR if you are interested.

chegevarko1982:
### Summary

Adds support for the LEDs on FreeJoy boards, so they can be bound to sim variables like any other MobiFlight output. Buttons and axes on these boards already worked through DirectInput and are untouched — this only adds the output side.

FreeJoy firmware has a documented "external LED control" mode. The board exposes a second, vendor specific HID interface (usage page `0xFF00`) which takes an Output Report with report ID 6 carrying a little-endian bitmask, one bit per LED, for up to 24 LEDs. Only LEDs flagged "External" in the FreeJoy Configurator follow the bitmask, the rest keep using the board's own button logic. Protocol reference: https://github.com/FreeJoy-Team/FreeJoyWiki/blob/master/rus/%D0%92%D0%BD%D0%B5%D1%88%D0%BD%D0%B5%D0%B5-%D1%83%D0%BF%D1%80%D0%B0%D0%B2%D0%BB%D0%B5%D0%BD%D0%B8%D0%B5-%D1%81%D0%B2%D0%B5%D1%82%D0%BE%D0%B4%D0%B8%D0%BE%D0%B4%D0%B0%D0%BC%D0%B8.md

The output path in the `Joystick` base class could not be reused: it sends a Feature Report rather than an Output Report, and it resolves the device with `GetHidDeviceOrNull(vid, pid)`, which is ambiguous here because the board exposes both its joystick interface and its LED interface under the same ids. `FreeJoyController` therefore picks the interface by usage page and writes an output report, while leaving everything else to the base class.

Changes:
- `MobiFlight/Joysticks/FreeJoy/FreeJoyReport.cs` — builds the bitmask from the output device states. No device access, so it is unit tested like the other report classes.
- `MobiFlight/Joysticks/FreeJoy/FreeJoyController.cs` — keeps input on DirectInput, opens the vendor specific interface only for the LED report.
- `Joysticks/freejoy/freejoy.joystick.json` — vendor/product id and the LED addresses, so the LED list is not hardcoded. `EnumerateOutputDevices`, `SetOutputDeviceState` and `Stop` are all the base class implementations.
- vendor/product id dispatch in `ControllerFactory`, next to the existing vendor branches.

No build file changes: source globbing and the existing `<None Update="Joysticks\**">` item cover everything.

This no longer depends on #3236. `Joystick` on main already exposes the HidSharp `Device`/`Stream` fields this uses, and `HidReportReceiver` is not needed here because only the output direction is involved.

### Acceptance criteria

- [x] FreeJoy LEDs are exposed as outputs and can be assigned in the UI
- [x] Setting an output on/off drives the matching LED on the board
- [x] Buttons and axes on the FreeJoy board still work through the regular DirectInput path
- [x] LEDs are turned off on stop, and the LED interface is closed on shutdown
- [x] Boards without the external LED interface log one warning and otherwise behave as before

### DoD Checklist

- [x] Unit tests available
  - [x] .NET backend
  - [x] ~~Frontend~~
- [x] ~~Frontend tests~~
- [x] ~~i18n - All user-facing strings are translated~~
- [ ] documentation
  - [x] User docs (docs.mobiflight.com) https://github.com/neilenns-projects/mobiflight-docs/pull/556
  - [x] ~~Developer docs~~

Happy to write the user docs section if you want this merged.

## Related issues

fixes #3230

### Out-of-scope

- Vendor/product ids are still hardcoded in `ControllerFactory` for dispatch, `0x0483` / `0x5757`, the FreeJoy defaults, the same way the other vendor branches do it. The device itself is then resolved from the definition file. The FreeJoy Configurator lets users change these ids, and such boards will not be picked up.
- The shipped definition can only offer generic `LED 0` … `LED 23` labels, because every FreeJoy board is different. See the open question in #3230 about user-supplied definitions — that only changes the definition file, not this code.
- LED brightness and RGB are not covered, the external control protocol is a plain on/off bitmask.
- No changes to button or axis handling.

### Notes

Verified on hardware: FreeJoy v1.7.1 board (VID `0x0483` / PID `0x5757`) with MSFS, LEDs switching on and off from the Output Config Wizard. Enumeration on that board shows the two interfaces, `mi_00` with usage `0x00010004` for the joystick and `mi_01` with usage page `0xFF00` and a 64 byte output report for the LEDs. Full unit test run is green.

This is my first PR to this repo, so please tell me if you'd prefer the feature shaped differently.

### DIFF
--- a/src/MobiFlightConnector/Joysticks/freejoy/freejoy.joystick.json
+++ b/src/MobiFlightConnector/Joysticks/freejoy/freejoy.joystick.json
@@ -1,0 +1,152 @@
+{
+  "$schema": "./mfjoystick.schema.json",
+  "InstanceName": "FreeJoy",
+  "VendorId": "0x0483",
+  "ProductId": "0x5757",
+  "Outputs": [
+    {
+      "Label": "LED 0",
+      "Id": "led.0",
+      "Byte": 0,
+      "Bit": 0
+    },
+    {
+      "Label": "LED 1",
+      "Id": "led.1",
+      "Byte": 0,
+      "Bit": 1
+    },
+    {
+      "Label": "LED 2",
+      "Id": "led.2",
+      "Byte": 0,
+      "Bit": 2
+    },
+    {
+      "Label": "LED 3",
+      "Id": "led.3",
+      "Byte": 0,
+      "Bit": 3
+    },
+    {
+      "Label": "LED 4",
+      "Id": "led.4",
+      "Byte": 0,
+      "Bit": 4
+    },
+    {
+      "Label": "LED 5",
+      "Id": "led.5",
+      "Byte": 0,
+      "Bit": 5
+    },
+    {
+      "Label": "LED 6",
+      "Id": "led.6",
+      "Byte": 0,
+      "Bit": 6
+    },
+    {
+      "Label": "LED 7",
+      "Id": "led.7",
+      "Byte": 0,
+      "Bit": 7
+    },
+    {
+      "Label": "LED 8",
+      "Id": "led.8",
+      "Byte": 1,
+      "Bit": 0
+    },
+    {
+      "Label": "LED 9",
+      "Id": "led.9",
+      "Byte": 1,
+      "Bit": 1
+    },
+    {
+      "Label": "LED 10",
+      "Id": "led.10",
+      "Byte": 1,
+      "Bit": 2
+    },
+    {
+      "Label": "LED 11",
+      "Id": "led.11",
+      "Byte": 1,
+      "Bit": 3
+    },
+    {
+      "Label": "LED 12",
+      "Id": "led.12",
+      "Byte": 1,
+      "Bit": 4
+    },
+    {
+      "Label": "LED 13",
+      "Id": "led.13",
+      "Byte": 1,
+      "Bit": 5
+    },
+    {
+      "Label": "LED 14",
+      "Id": "led.14",
+      "Byte": 1,
+      "Bit": 6
+    },
+    {
+      "Label": "LED 15",
+      "Id": "led.15",
+      "Byte": 1,
+      "Bit": 7
+    },
+    {
+      "Label": "LED 16",
+      "Id": "led.16",
+      "Byte": 2,
+      "Bit": 0
+    },
+    {
+      "Label": "LED 17",
+      "Id": "led.17",
+      "Byte": 2,
+      "Bit": 1
+    },
+    {
+      "Label": "LED 18",
+      "Id": "led.18",
+      "Byte": 2,
+      "Bit": 2
+    },
+    {
+      "Label": "LED 19",
+      "Id": "led.19",
+      "Byte": 2,
+      "Bit": 3
+    },
+    {
+      "Label": "LED 20",
+      "Id": "led.20",
+      "Byte": 2,
+      "Bit": 4
+    },
+    {
+      "Label": "LED 21",
+      "Id": "led.21",
+      "Byte": 2,
+      "Bit": 5
+    },
+    {
+      "Label": "LED 22",
+      "Id": "led.22",
+      "Byte": 2,
+      "Bit": 6
+    },
+    {
+      "Label": "LED 23",
+      "Id": "led.23",
+      "Byte": 2,
+      "Bit": 7
+    }
+  ]
+}
--- a/src/MobiFlightConnector/MobiFlight/Joysticks/ControllerFactory.cs
+++ b/src/MobiFlightConnector/MobiFlight/Joysticks/ControllerFactory.cs
@@ -1,3 +1,4 @@
+using MobiFlight.Joysticks.FreeJoy;
 using MobiFlight.Joysticks.VKB;
 using MobiFlight.Joysticks.WinCtrl;
 using MobiFlight.Joysticks.IIDB;
@@ -38,6 +39,12 @@ namespace MobiFlight.Joysticks
 
             // Check for IIDB devices
             if (vendorId == IIDBDevice.IIDB_VENDOR_ID)
+            {
+                return true;
+            }
+
+            // Check for FreeJoy boards
+            if (vendorId == FreeJoyController.FREEJOY_VENDOR_ID && productId == FreeJoyController.FREEJOY_PRODUCT_ID)
             {
                 return true;
             }
@@ -119,6 +126,12 @@ namespace MobiFlight.Joysticks
             if (vendorId == IIDBDevice.IIDB_VENDOR_ID)
             {
                 return new IIDBDevice(diJoystick, definition);
+            }
+
+            // Handle FreeJoy boards by vendor/product ID
+            if (vendorId == FreeJoyController.FREEJOY_VENDOR_ID && productId == FreeJoyController.FREEJOY_PRODUCT_ID)
+            {
+                return new FreeJoyController(diJoystick, definition);
             }
 
             // Handle AuthentiKit by instance name
--- a/src/MobiFlightConnector/MobiFlight/Joysticks/FreeJoy/FreeJoyController.cs
+++ b/src/MobiFlightConnector/MobiFlight/Joysticks/FreeJoy/FreeJoyController.cs
@@ -1,0 +1,176 @@
+using HidSharp;
+using System;
+using System.Linq;
+
+namespace MobiFlight.Joysticks.FreeJoy
+{
+    /// <summary>
+    /// Adds external LED control for FreeJoy boards (https://github.com/FreeJoy-Team/FreeJoyWiki).
+    /// </summary>
+    /// <remarks>
+    /// Buttons and axes keep using the regular DirectInput path in the base class: FreeJoy is
+    /// user-flashed firmware, so there is no fixed input report layout to parse.
+    /// <para>
+    /// Only the LEDs need a device of their own. They are addressed on a second, vendor-specific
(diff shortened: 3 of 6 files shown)


Koseng:
Hi, thanks a lot for your contribution! Timing is a little bit unfortunate, cause we are in the process of removing Hid.Net, so you need to adjust. See https://github.com/MobiFlight/MobiFlight-Connector/pull/3236

And have a look at the pattern in WingFlex `BaseCube.cs`. Can't you use that pattern and configure the LEDs the same way via configuration file?

chegevarko1982:
Thanks for the quick review, and no problem about the timing — better to find this out now than after a merge. I've put this PR back into draft and will rebase onto #3236 and rewrite the LED path on HidSharp once it lands. HidSharp actually removes one wart here: Hid.Net's write transform doesn't pad the payload, so I had to size the buffer from `WriteBufferSize - 1` by hand.

On driving the LEDs from the definition file: agreed, and it fits well. The FreeJoy external LED report is a 4 byte bitmask, so it maps one to one onto the `Byte`/`Bit` outputs the definition files already use, and `Joystick.EnumerateOutputDevices()` builds `Lights` from that for free. That also removes the hardcoded VID/PID and turns the report builder into a pure function I can unit test the way `WinCtrlReportTests` does — so it answers both open questions I had listed.

On `BaseCube` itself I'd rather not derive from it, and I want to check that with you before writing anything. `BaseCube` passes `null` as the DirectInput device and reads input reports on its own thread with a fixed report parser. That works for the Cubes because they are fixed products. FreeJoy is user-flashed firmware: the button and axis layout differs from board to board, so there is no fixed input report to parse, and DirectInput already handles the input side correctly today. My plan is to take the output/definition/report pattern from `BaseCube` but leave input on the regular DirectInput path in `Joystick`. Tell me if you'd rather have it the other way.

Two things I'd like your call on, since they are design decisions rather than implementation details:

1. **User-specific definitions.** Definitions are loaded only from the `Joysticks` folder in the installation (`JoystickManager.cs:78`). WingFlex and WinCtrl are fixed hardware, one definition per model. A FreeJoy board is whatever the user built — LED count, which LEDs are flagged "External", and the labels all differ per user, and the FreeJoy Configurator lets users change VID/PID as well. So a single shipped `freejoy.joystick.json` can realistically only offer 24 generic `LED 0..23` entries. Is a user-supplied definition folder something you'd consider, or should I ship the generic definition for now and leave that for later?

2. **Reaching the LED interface.** The LEDs sit on a second HID interface (Usage Page `0xFF00`) behind the same VID/PID as the joystick interface, and they take an Output Report, not a Feature Report — so the `GetHidDeviceOrNull(vid, pid)` lookup in `Joystick.SendData()` can't address them. Does #3236 give me a seam for a usage-page filtered device, or should I enumerate with HidSharp inside the controller?

I'll post a short design outline in #3230 too, so the plan is reviewable before I write the code.

chegevarko1982:
Short update while #3236 is still open: I've done the rework you asked for and verified it on hardware, so it's ready whenever your PR lands.

It sits on top of your `UsbHid` branch rather than main, so I haven't touched this PR — pushing it here would drag all of #3236 into the diff. It's a single commit if you want to glance at it: https://github.com/chegevarko1982/MobiFlight-Connector/commit/9ac183a758248194481c6724e39f9493657fdc2d

- HidSharp instead of Hid.Net, using the `Device`/`Stream` fields of the base class the way `WinCtrlRmp` does
- the LEDs come from a new `freejoy.joystick.json` instead of a hardcoded list, so `EnumerateOutputDevices`, `SetOutputDeviceState` and `Stop` are all the base class implementations now
- `FreeJoyReport` is a pure function, unit tested in the style of `WinCtrlReportTests`

Verified on the board: the LED interface is selected by usage page (`mi_01`, usage `0xFF000001`) and not the joystick interface (`0x00010004`), and LEDs light up from the Output Config Wizard. Full unit test run is green.

No rush on the two questions from my previous comment — they only affect what goes into the definition file, not this code.

tigert:
ok, I dug up a bluepill board because this would be a neat way to do for example a throttle console.

tigert:
Hmm, the freejoy firmware also supports "simhub" mode where you can control up to 20 WS2812 neopixel leds via serial link.

https://github.com/FreeJoy-Team/FreeJoy/blob/master/application/Src/simhub.c

That is definitely out of scope for this PR, to keeps separate, but I am just mentioning this because it would also be interesting to explore WS2812 led strips as output device in the future. It would be a bigger thing to ponder because it would be a whole new output device, to think how that would work. Maybe we can get to that after this.

chegevarko1982:
> Hmm, the freejoy firmware also supports "simhub" mode where you can control up to 20 WS2812 neopixel leds via serial link.
> 
> https://github.com/FreeJoy-Team/FreeJoy/blob/master/application/Src/simhub.c
> 
> That is definitely out of scope for this PR, to keeps separate, but I am just mentioning this because it would also be interesting to explore WS2812 led strips as output device in the future. It would be a bigger thing to ponder because it would be a whole new output device, to think how that would work. Maybe we can get to that after this.

Yap, its cool future, for control multicolor indication as example

ColinM9991:
We've migrated to .NET 10 which introduces an entirely new csproj format. For the conflict on this pull request, you can discard your csproj changes and take accept the file from main.

Source file globbing comes automatically with the .NET SDK so this is no longer necessary to specify in the csproj.

chegevarko1982:
Rebased onto main and force-pushed, so this is now a single commit against the .NET 10 tree. The description above is rewritten to match — the old one described the Hid.Net implementation.

@ColinM9991 thanks, that made it simple: both csproj changes are gone and the files are taken from main as-is. The definition file needs nothing either — `<None Update="Joysticks\**">` already covers it — so this PR now touches no build files at all.

@Koseng the dependency on #3236 turned out to be unnecessary, so this doesn't need to wait for it. `Joystick` on main already exposes the HidSharp `Device`/`Stream` fields, and `HidReportReceiver` isn't involved because only the output direction is used here — the LEDs are written, nothing is read back. The rest of your review is in: the LEDs come from `freejoy.joystick.json` rather than a hardcoded list, so `EnumerateOutputDevices`, `SetOutputDeviceState` and `Stop` are the base class implementations, and `FreeJoyReport` is a pure function with unit tests in the style of `WinCtrlReportTests`.

One thing worth flagging from re-reading my own code: `UpdateOutputDeviceStates()` runs from the poll loop, and the LED write returns early without clearing `RequiresOutputUpdate` when the interface is missing. On a board with no external LED interface that would have logged a warning on every tick, so the warning is now logged once and re-armed only if the interface is lost again.

Verified on hardware after the rebase: FreeJoy v1.7.1, LEDs switching from the Output Config Wizard, full unit test run green.

Taking it out of draft. The two questions from earlier are still open but neither blocks this — they only affect what goes into the definition file.

neilenns:
@chegevarko1982 Catching up on this since it requires a doc update and I didn't get tagged for the PR.

Does this simply need FreeJoy added to our supported game controller table, so we can list it as supporting outputs?

https://docs.mobiflight.com/game-controllers/

DocMoebiuz:
> @chegevarko1982 Catching up on this since it requires a doc update and I didn't get tagged for the PR.
> 
> Does this simply need FreeJoy added to our supported game controller table, so we can list it as supporting outputs?
> 
> https://docs.mobiflight.com/game-controllers/

That was my thought too. I think that is fine.

neilenns:
Doc PR https://github.com/neilenns-projects/mobiflight-docs/pull/556

PR #3236, opened 2026-08-09
Unify all joystick controller to only use HidSharp in a performant way

### LINKED ISSUE #2337: Change HidSharp or Hid.net library
Benefits:
* ~~Hid.net can do everything HidSharp can do~~ Not true, hid.net cannot send feature reports
* Performance benefits
* WinWing and WingFlex are using it already

Koseng:
### Summary
Unify all Joystick controller classes to only use HidSharp. Hid.Net was introduced because of really bad performance of HidSharp in receiving reports with HidDeviceInputReceiver. Now the solution is to use HidSharp in a more low level way with a dedicated blocking read thread per device  and avoid HidDeviceInputReceiver. Performance was measured and is excellent.

### Acceptance criteria

- [x] VKB tested
- [x] WinCtrl tested
- [x] Bodnar tested
- [x] Octavi tested
- [x] Wingflex tested

### DoD Checklist

- [x] Unit tests available
  - [x] .NET backend
  - ~~[ ] Frontend~~
- ~~[ ] Frontend tests~~
- ~~[ ] i18n - All user-facing strings are translated~~
- ~~[ ] documentation~~
     
## Related issues

fixes #2337

### Out-of-scope
Further joystick refactoring and unification.


src/MobiFlightConnector/MobiFlight/Joysticks/HidReportReceiver.cs
```diff
@@ -0,0 +1,199 @@
...
+                var session = ActiveSession;
+                return session != null && session.Active;
+            }
+        }
+
+        /// <summary>
+        /// Returns the payload of a raw report, i.e. everything after the report ID byte
+        /// at index 0. For report parsers whose offsets are payload-relative, following
+        /// the vendor protocol tables.
+        /// </summary>
```
DocMoebiuz: We have to decide whether we really want to strip the report Id. 
We assume that there is always only one report Id that we are going to process.

Controllers could have different report Id and now we would not be able to access it anymore. 

Wingflex controllers for example define more than one. for input processing it was only 1 though.

I would rather leave this report id stripping or ignoring or processing to the UsbReport implementation.

src/MobiFlightConnector/MobiFlight/Joysticks/HidReportReceiver.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// </param>
+        /// <param name="bufferSize">
+        /// The read buffer size, typically <c>HidDevice.GetMaxInputReportLength()</c>.
+        /// Passed in by the caller so hardware that cannot report its input length fails
+        /// on the caller's thread instead of via <paramref name="onError"/>.
+        /// </param>
+        /// <param name="onReport">Called for every received report with a copy of the report bytes.</param>
+        /// <param name="onError">Called once when the loop dies on a read error; may be null.</param>
+        /// <param name="threadName">Optional thread name to identify the device in debugging tools.</param>
+        public void Start(Stream stream, int bufferSize, Action<byte[]> onReport, Action<Exception> onError = null, string threadName = null)
```
DocMoebiuz: I am not a big fan of such lengthy inline documentation.

src/MobiFlightConnector/MobiFlight/Joysticks/WingFlex/Dap500.cs
```diff
@@ -106,15 +96,21 @@ private void RestoreConfig()
             SendData(new DapConfig() { }.ToData);
         }
 
-        private void InputReceiver_Received(object sender, System.EventArgs e)
+        private void OnReportReceived(byte[] inputReport)
+        {
+            ProcessInputReportBuffer(inputReport);
+        }
```
DocMoebiuz: This could totally go into a common base class

src/MobiFlightConnector/MobiFlight/Joysticks/WingFlex/Dap500.cs
```diff
@@ -106,15 +96,21 @@ private void RestoreConfig()
...
 
-            while (inputReceiver.TryRead(inputReportBuffer, 0, out _))
+            // Drop the dead connection so the next Update() can reconnect.
+            lock (this)
             {
-                //
-                ProcessInputReportBuffer(inputReportBuffer);
+                Stream?.Close();
+                Stream = null;
+                Device = null;
```
DocMoebiuz, line 113: This could totally go into a common base class

src/MobiFlightConnector/MobiFlight/Joysticks/WingFlex/Dap500.cs
DocMoebiuz, line 129: This could totally go into a common base class

DocMoebiuz: This could totally go into a common base class

src/MobiFlightConnector/MobiFlight/Joysticks/WingFlex/Dap500.cs
```diff
@@ -216,6 +212,14 @@ protected override void EnumerateDevices()
...
+            Receiver.Stop();
+
+            if (Stream != null)
+            {
+                Stream.Close();
+                Stream = null;
+            }
+
             base.Shutdown();
         }
```
DocMoebiuz, line 225: This could totally go into a common base class

DocMoebiuz (changes requested):
Really cool to consolidate and only use a single library.
Thank you so much!

I feel there is repetitive code that could go into a base class or could be reused across different controller models. maybe some "strategy" pattern? hidinputreport vs directInput

I am also OK with keeping it out of scope and tackle it  in case we want to extend the joystick json to "describe" the input report processing part more universally. And then we would get away with just specifying which byte and bit means what.

DocMoebiuz:
I would really let the USBReportParser receive the Report Id and the process or ignore it on that level - that would be my requested change

src/MobiFlightConnector/MobiFlight/Joysticks/HidReportReceiver.cs
Koseng: The report id is not stripped. The class callback returns it including the report id. That is just a static helper method for the current code, where on some places it is needed without the report id.

src/MobiFlightConnector/MobiFlight/Joysticks/HidReportReceiver.cs
Koseng: Agree, will shorten it.

src/MobiFlightConnector/MobiFlight/Joysticks/HidReportReceiver.cs
Koseng: I'll try to unify the processing and get rid of that helper method. Because some controllers expect a byte array with the leading report id and some without.

src/MobiFlightConnector/MobiFlight/Joysticks/WingFlex/BaseCube.cs
```diff
@@ -48,21 +32,44 @@ public override string Name
...
+            try
+            {
+                return Device?.GetSerialNumber();
+            }
+            catch
+            {
+                return null;
+            }
+        }
+
```
DocMoebiuz, line 76: This entire optimization is theoretical, isn't it?
This is a specific implementation - where it was tested that the device will report a serial.
The Device instance provides a property for it - is this "caching" really rquired?

I prefer KISS over some optimization where we don't have an actual need.

src/MobiFlightConnector/MobiFlight/Joysticks/HidReportReceiver.cs
```diff
@@ -0,0 +1,184 @@
...
+        private Thread ReadThread;
+
+        public bool IsReceiving
+        {
+            get
+            {
+                var session = ActiveSession;
+                return session != null && session.Active;
+            }
+        }
```
DocMoebiuz: When reading through the code, I stumbled over the naming a couple of times.
A more generic: 
- IsReady
- IsInitialized
- IsActive

would be more apropriate - we have only set up everything to be ready for receiving but it still depends on the controller whether they send data or not. So even though we currently are not receiving any data, it would still return true.

src/MobiFlightConnector/MobiFlight/Joysticks/HidReportReceiver.cs
```diff
@@ -0,0 +1,184 @@
...
+        /// Returns the payload of a raw report, i.e. everything after the report ID byte
+        /// at index 0. For report parsers whose offsets are payload-relative, following
+        /// the vendor protocol tables.
+        /// </summary>
+        public static byte[] GetPayload(byte[] rawReport)
+        {
+            byte[] payload = new byte[rawReport.Length - 1];
+            Array.Copy(rawReport, 1, payload, 0, payload.Length);
+            return payload;
+        }
```
DocMoebiuz: HidReportReceiver is receiving HidReports - the HidReport instance should provide access to ReportId and Payload - ideally as properties.

DocMoebiuz (changes requested):
I added some more comments.

If we decide to create a HidController base class outside of this PR - that's fine by me as long as we create the respective issue for it and describe the scope there.

It would be a good exercise for our intern.

DocMoebiuz:
The conflicts are because of the .NET 10 way of defining project resources.

ColinM9991:
@Koseng @DocMoebiuz 

I've reviewed both csproj files. The easiest way to resolve is to take the full csproj files from main and then remove those packages (Device.Net and Hid.Net) again.

Source files are globbed automatically by the dotnet SDK so those parts of the csproj are now irrelevant

src/MobiFlightConnector/MobiFlight/Joysticks/WingFlex/BaseCube.cs
Koseng: ### Why `CachedSerialNumber` was introduced in `BaseCube`

Before the migration to lower-level HidSharp, the serial number came from Device.Net via
`Device?.ConnectedDeviceDefinition?.SerialNumber` — a plain property read on already-enumerated
data, essentially free.

With HidSharp, the serial has to be fetched via `HidDevice.GetSerialNumber()`, which is a native
HID query that performs actual device I/O. On top of that, HidSharp **throws an exception** when
the hardware refuses to report a serial number — which some Cubes do.

The `Serial` property getter is evaluated for **every input event**: `ProcessInputReportBuffer`
(`BaseCube.cs`) calls `UpdateButtons` / `UpdateAxis` in the `Joystick` base class, and each
detected state change constructs the event args with a fresh `Controller` object:

```csharp
OnButtonPressed?.Invoke(this, new InputEventArgs()
{
    Controller = new Controller() { Name = this.Name, Serial = this.Serial },
    ...
});
```

(see `Joystick.cs` — `UpdateButtons`, `UpdateAxis`, `UpdatePOV`)

Without caching, every button press, axis movement, or encoder tick would trigger a native device
query — and potentially a thrown/caught exception — on the HID reader thread. So the serial is
fetched once in `Connect()` and stored in `CachedSerialNumber`. `string.Empty` doubles as an
"already tried, device has no serial" marker so reconnects don't re-query either.

### Why other `Joystick` derivatives are not affected

All other derivatives (`IIDBDevice`, `VKBDevice`, `Octavi`, `Dap500`, the `WinCtrl` family,
`BodnarBoard`) receive a `SharpDX.DirectInput.Joystick` in their constructor and derive their
serial from the DirectInput `InstanceGuid`:

```csharp
get { return SerialPrefix + DIJoystick.Information.InstanceGuid; }
```

That's an in-process COM call reading data DirectInput already holds from device enumeration —
no device I/O, and it cannot throw when the hardware reports no serial.

`BaseCube` is the only class constructed with `base(null, definition)`: there is no DirectInput
device at all, the Cubes run as pure HID devices through HidSharp. It's therefore the only place
where reading the serial has real per-access cost — and the only place that needs the cache.

src/MobiFlightConnector/MobiFlight/Joysticks/HidReportReceiver.cs
Koseng: Agree, it is not very precise. Favorite alternatives:

- IsRunning
- IsListening

What do you like more?

src/MobiFlightConnector/MobiFlight/Joysticks/HidReportReceiver.cs
Koseng: I do not understand. This is a helper class for creating the thread and it returns the original byte array in a callback for each received report. There is no saved state and currently no HidReport object. Later the joysticks create report objects.

What we could do, is introduce a new class and return the byte array in that class and add properties to that new class.   

So do that, or keep it like it is?

src/MobiFlightConnector/MobiFlight/Joysticks/WingFlex/BaseCube.cs
DocMoebiuz: How would claude know this:

With HidSharp, the serial has to be fetched via HidDevice.GetSerialNumber(), which is a native
HID query that performs actual device I/O. On top of that, HidSharp throws an exception when
the hardware refuses to report a serial number — which some Cubes do.

But it is OK. We should then have a backed field, and when it is null - we get the SerialNumber from the controller and store it there - it won't change during the lifetime of that instance.

src/MobiFlightConnector/MobiFlight/Joysticks/HidReportReceiver.cs
DocMoebiuz: IsReady or isRunning would be fine by me

src/MobiFlightConnector/MobiFlight/Joysticks/HidReportReceiver.cs
DocMoebiuz: that's what i had in mind. exactly. We should introduce a lightweight class that holds the byte array and has these two methods. basically, the hidreport. The report parser will then access the payload and report id as needed.

src/MobiFlightConnector/MobiFlight/Joysticks/HidReportReceiver.cs
Koseng: Renamed to IsRunning

src/MobiFlightConnector/MobiFlight/Joysticks/WingFlex/BaseCube.cs
Koseng: ok

src/MobiFlightConnector/MobiFlight/Joysticks/HidReportReceiver.cs
Koseng: ok, I introduce wrapper class.

DocMoebiuz (approved):

PR #3251, opened 2026-08-11
Improve input events log entries

### LINKED ISSUE #3152: only show "executing..." for event types that have an action set
I have OnPress defined for a button:

```
CJ4 audio panel => Com1 MIC => PRESS
CJ4 audio panel => Executing "COM 1 mic select". (PRESS)
CJ4 audio panel => Com1 MIC => RELEASE
CJ4 audio panel => Executing "COM 1 mic select". (RELEASE)
```
However, I have not defined the OnRelease action for this config. So it feels wrong to show that it executed an action that is not defined.

It would be good to only log an action when it actually happens, so that it does not show things that are not happening.

OK434:
### Summary

Only log `Executing...` when an action is actually defined for the triggered event.
This prevents the log from showing that an undefined action was executed.

### Screenshots / recordings

image

### Acceptance criteria

- [x] Only log `Executing...` when an action is defined for the event

### DoD Checklist

- [x] Unit tests available
  - [x] .NET backend

     
## Related issues

fixes #3152

### DIFF
--- a/src/MobiFlightConnector/Base/InputEventExecutor.cs
+++ b/src/MobiFlightConnector/Base/InputEventExecutor.cs
@@ -98,7 +98,15 @@ namespace MobiFlight.Execution
                         continue;
                     }
 
-                    Log.Instance.log($"{e.Controller.Name} => Executing \"{cfg.Name}\". ({e.GetEventActionLabel()})", LogSeverity.Info);
+                    var action = cfg.GetInputAction(e);
+
+                    if (action != null)
+                    {
+                        Log.Instance.log(
+                            $"{e.Controller.Name} => Executing \"{cfg.Name}\". ({e.GetEventActionLabel()})",
+                            LogSeverity.Info
+                        );
+                    }
 
                     if (cfg.button != null)
                         cfg.button.CanExecute = () => cfg.Active && CheckPreconditions(cfg);
--- a/src/MobiFlightConnector/Base/InputEventExecutor.cs
+++ b/src/MobiFlightConnector/Base/InputEventExecutor.cs
@@ -59,7 +59,7 @@ namespace MobiFlight.Execution
         public Dictionary<string, IConfigItem> Execute(InputEventArgs e, bool isStarted)
         {
             var updatedValues = new Dictionary<string, IConfigItem>();
-            string inputKey = CreateInputKey(e);            
+            string inputKey = CreateInputKey(e);
             var msgEventLabel = e.GetMsgEventLabel();
 
             if (!inputCache.ContainsKey(inputKey))
@@ -68,7 +68,7 @@ namespace MobiFlight.Execution
             }
 
             if (inputCache[inputKey].Count == 0)
-            {                
+            {
                 return updatedValues;
             }
 
@@ -100,14 +100,11 @@ namespace MobiFlight.Execution
 
                     var action = cfg.GetInputAction(e);
 
-                    if (action != null)
+                    if (action == null)
                     {
-                        Log.Instance.log(
-                            $"{e.Controller.Name} => Executing \"{cfg.Name}\". ({e.GetEventActionLabel()})",
-                            LogSeverity.Info
-                        );
+                        continue;
                     }
-
+                    Log.Instance.log($"{e.Controller.Name} => Executing \"{cfg.Name}\". ({e.GetEventActionLabel()})", LogSeverity.Info);
                     if (cfg.button != null)
                         cfg.button.CanExecute = () => cfg.Active && CheckPreconditions(cfg);
 
@@ -192,13 +189,13 @@ namespace MobiFlight.Execution
                 return false;
 
             bool deviceNameMatches = cfg.Device.Name == e.Device.Name;
-            
+
             // For backward compatibility we have to make this check
             // because we used to have the label in the config
             // but now we want to store the internal button identifier
             // so that the label can change any time without breaking the config
             bool isJoystickWithLabelMatch = Joystick.IsJoystickSerial(cfg.Controller.Serial) && cfg.Device.Name == e.Device.Label;
-            
+
             return deviceNameMatches || isJoystickWithLabelMatch;
         }
 
--- a/src/MobiFlightConnector/MobiFlight/InputConfigItem.cs
+++ b/src/MobiFlightConnector/MobiFlight/InputConfigItem.cs
@@ -405,18 +405,18 @@ namespace MobiFlight
                             return null;
                     }
                 case DeviceType.Encoder:
-                    switch (Convert.ToInt32(e.Value))
+                    switch ((MobiFlightEncoder.InputEvent)e.Value)
                     {
-                        case 0:
+                        case MobiFlightEncoder.InputEvent.LEFT:
                             return encoder?.onLeft;
 
-                        case 1:
+                        case MobiFlightEncoder.InputEvent.LEFT_FAST:
                             return encoder?.onLeftFast ?? encoder?.onLeft;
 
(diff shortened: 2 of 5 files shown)


src/MobiFlightConnector/Base/InputEventExecutor.cs
```diff
@@ -98,7 +98,15 @@ public Dictionary<string, IConfigItem> Execute(InputEventArgs e, bool isStarted)
...
-                    Log.Instance.log($"{e.Controller.Name} => Executing \"{cfg.Name}\". ({e.GetEventActionLabel()})", LogSeverity.Info);
+                    var action = cfg.GetInputAction(e);
+
+                    if (action != null)
+                    {
+                        Log.Instance.log(
+                            $"{e.Controller.Name} => Executing \"{cfg.Name}\". ({e.GetEventActionLabel()})",
+                            LogSeverity.Info
+                        );
+                    }
```
DocMoebiuz, line 106: Looking at the code, I am wondering if anything should be processed at all in case `action == null` or if we should continue with the next item like we do in case the preconditions are not true?

If we could just continue with the next item, I prefer to use the early return (or in this case "continue").

src/MobiFlightConnector/Base/InputEventExecutor.cs
```diff
@@ -98,7 +98,15 @@ public Dictionary<string, IConfigItem> Execute(InputEventArgs e, bool isStarted)
...
 
                     if (cfg.button != null)
                         cfg.button.CanExecute = () => cfg.Active && CheckPreconditions(cfg);
```
DocMoebiuz, line 109: i am also seeing this unrelated issue:

CheckPrecondition(cfg) which is a repetition. We already checked in line 94 and we know that the CheckPrecondition is true.

src/MobiFlightConnector/MobiFlight/InputConfigItem.cs
```diff
@@ -385,6 +385,50 @@ public override bool Equals(object obj)
...
+                            return encoder?.onLeftFast ?? encoder?.onLeft;
+
+                        case 2:
+                            return encoder?.onRight;
+
+                        case 3:
+                            return encoder?.onRightFast ?? encoder?.onRight;
+
+                        default:
+                            return null;
```
DocMoebiuz, line 423: we should have Encoder constants too - we don't want to work with numbers.
compare with Button

tests/MobiFlightUnitTests/Base/InputEventExecutorTests.cs
```diff
@@ -777,7 +788,7 @@ public void Execute_InputShiftRegisterWithWrongPinButCorrectDeviceType_Skips()
                     Type = DeviceType.InputShiftRegister,
                 },
                 InputType = DeviceType.Button,
-                Value = 0 
+                Value = 0
```
DocMoebiuz: Better use the constants instead

tests/MobiFlightUnitTests/MobiFlight/ExecutionManagerTests.cs
```diff
@@ -1157,7 +1157,7 @@ public void ExecuteConfig_ChangedConfigItemBecomesEffectiveWithoutStopAndRun()
                 Device = new Button() { Name = Device.Name },
                 button = new InputConfig.ButtonInputConfig()
                 {
-                    onPress = new VariableInputAction()
+                    onRelease = new VariableInputAction()
```
DocMoebiuz: Why this change?

tests/MobiFlightUnitTests/MobiFlight/ExecutionManagerTests.cs
DocMoebiuz: Ah because we are using the wrong "number" here. For button press it should have been 0?

That was a "shadowed" bug - for clarity we should also use the constants here.

tests/MobiFlightUnitTests/MobiFlight/ExecutionManagerTests.cs
DocMoebiuz: related to line 1179

tests/MobiFlightUnitTests/MobiFlight/InputConfigItemTests.cs
```diff
@@ -348,6 +348,56 @@ public void GetInputActionsByTypeTest()
             Assert.HasCount(3, result);
         }
 
+
+        [TestMethod()]
+        [DataRow((int)MobiFlightButton.InputEvent.PRESS, "onPress")]
+        [DataRow((int)MobiFlightButton.InputEvent.RELEASE, "onRelease")]
+        [DataRow((int)MobiFlightButton.InputEvent.LONG_RELEASE, "onLongRelease")]
+        public void GetInputAction_Button_ReturnsCorrectAction(
```
DocMoebiuz, line 356: Is there a reason we don't test for encoders and analoginput?

tests/MobiFlightUnitTests/Base/InputEventExecutorTests.cs
```diff
@@ -777,7 +788,7 @@ public void Execute_InputShiftRegisterWithWrongPinButCorrectDeviceType_Skips()
...
             };
```
DocMoebiuz, line 795: We should have an explicit test for the log message showing and not showing.

The test method Execute_MatchingActiveConfigItem_ExecutesSuccessfully is relying on the correctness of the Log message showing, vs not showing. But it has not been verified before.

DocMoebiuz (changes requested):

src/MobiFlightConnector/Base/InputEventExecutor.cs
DocMoebiuz: oh no. disregard my comment.... we are creating a function closure and it will be evaluated at a later point when the action is invoked by the timer. Don't change it... but it deserves a clear comment.

tests/MobiFlightUnitTests/MobiFlight/ExecutionManagerTests.cs
OK434: Because the test triggers a RELEASE event (Value = 1), the config needs an onRelease action for the action to be executed.

OK434: ??

tests/MobiFlightUnitTests/MobiFlight/ExecutionManagerTests.cs
```diff
@@ -1157,7 +1157,8 @@ public void ExecuteConfig_ChangedConfigItemBecomesEffectiveWithoutStopAndRun()
                 Device = new Button() { Name = Device.Name },
                 button = new InputConfig.ButtonInputConfig()
                 {
-                    onPress = new VariableInputAction()
+                    
+                    onRelease = new VariableInputAction()
```
DocMoebiuz: this could have stayed "onPress" (since it is the intuitive event) and the value in 1179 should have been "0" or like you did, should have been (int)MobiFlightButton.InputEvent.PRESS

What about the test that confirms your change? The change that we are not printing "Executing..." in case there is no action

DocMoebiuz (changes requested):

DocMoebiuz (approved):
Thanks!

For the "verify no log" test, I would have also used OnPress. The reason: we have verified correct behavior for OnPress when onPress action was set. Now, we verify that OnPress doesn't log anything if onPress action is not set.

I leave it up to you if you want to change it.

The rest is implemented correctly.

DocMoebiuz (approved):

PR #3255, opened 2026-08-11
Fix controller filter

### LINKED ISSUE #3232: Filtering controllers in input config doesn't work
**Describe the bug**
Attempting to filter the **Select controller...** dropdown in the input config dialog doesn't work.

**Steps To Reproduce**
1. Create a new input config
2. Click in the **Select controller...** dropdown
3. Type a few letters of a known controller

**Expected behavior**
List filters to show the controller

**Actual behavior**
No controller found is displayed

**Additional context**

Here's what's in my controller list:

Image

And here's what shows when I type an `M`:

Image

OK434:
### Summary

The default Command filtering uses fuzzy matching, which can return items that only partially match the search term. The custom filter changes this behavior to require the complete search string to be present. For example, searching for Button 12 should only match Button 12 and not Button 13.
### Screenshots / recordings

image
image

### Acceptance criteria

- [x] Controller search can be opened from the Trigger Panel
- [x] Controller search filters results by the displayed controller label
- [x] Searching for a controller only shows controllers whose label contains the complete search string
- [x] Existing Trigger Panel functionality still works

### DoD Checklist

- [x] Frontend tests

     
## Related issues

fixes #3232


src/MobiFlightConnector/frontend/src/components/wizard/components/ConfigTrigger.tsx
```diff
@@ -159,7 +159,6 @@ const ConfigTrigger = ({ configItem, setConfigItem }: ConfigTriggerProps) => {
               widthClass="w-50"
               getLabel={(controller) => (controller as Controller).Name}
               getValue={(controller) =>
-                (controller as Controller).Serial ??
```
DocMoebiuz, line 164: This change is not 100% correct. You can have multiple Controllers connected with the same name.
The serial number helps to identify them uniquely.

DocMoebiuz: In general, the combobox should filter by label match and not by value match.

DocMoebiuz (changes requested):

src/MobiFlightConnector/frontend/src/components/ComboBox.tsx
```diff
@@ -71,18 +69,25 @@ const ComboBox = <T,>({
...
           <IconChevronDown className="opacity-50" />
         </Button>
       </PopoverTrigger>
       <PopoverContent className={cn(widthClass, "p-0")} align={align}>
-        <Command>
+        <Command
+          filter={(value, search) => {
+            return value.toLowerCase().includes(search.toLowerCase()) ? 1 : 0
+          }}
+        >
```
DocMoebiuz: why are we adding this? is the default search case sensitive?

src/MobiFlightConnector/frontend/src/components/wizard/components/ConfigTrigger.tsx
```diff
@@ -159,7 +159,7 @@ const ConfigTrigger = ({ configItem, setConfigItem }: ConfigTriggerProps) => {
...
+              (controller as Controller).Serial ||
```
DocMoebiuz: let's use the `??` null-coalesce operator to explicitly express that we are handling the `null` case.

Basically, just keep what we had in the beginning.

DocMoebiuz (changes requested):

src/MobiFlightConnector/frontend/src/components/ComboBox.tsx
OK434: We added this to change the default fuzzy filtering behavior. The issue was that searching for multiple characters (e.g. 22) could return items that only contained individual characters from the search term. We need a substring match instead, so the full search string has to be present in the value. The toLowerCase() calls keep the existing case-insensitive behavior.

DocMoebiuz: Ah OK! That makes sense.

This detail should be mentioned in the PRs description, and should also have a comment here.
Instead of an inline function, we should use a proper method with a speaking name which would improve readability.

If we add this address a specific behavior, e.g. Button 12 should only show Button 12, and not Button 13... a playwright test has to be added that verifies this.

src/MobiFlightConnector/frontend/src/components/ComboBox.tsx
```diff
@@ -91,10 +96,11 @@ const ComboBox = <T,>({
             <CommandGroup>
               {items.map((item) => {
                 const itemValue = getValue(item)
+                const itemLabel = getLabel(item)
                 return (
                   <CommandItem
                     key={itemValue}
-                    value={itemValue}
+                    value={itemLabel}
```
DocMoebiuz, line 116: This change, that we are now filtering by label, should also be mentioned explicitly in the PR description, likely as "Acceptance criteria"

DocMoebiuz (changes requested):

src/MobiFlightConnector/frontend/tests/data/connectedControllers.testdata.json
```diff
@@ -143,5 +143,28 @@
...
+        "Name": "12",
+        "Label": "Button 12",
+        "Type": "Button"
+      },
+      {
+        "Name": "2",
+        "Label": "Button 2",
+        "Type": "Button"
+      }
+    ]
```
DocMoebiuz: You could have added these buttons to the Bravo throttle instead of creating an additional controller. I hope it won't break other tests

DocMoebiuz (approved):

PR #3264, opened 2026-08-11
Add awaited task in debug builds to avoid build error

### LINKED ISSUE #3258: Debug builds fail locally due to an unneeded `async` method
**Describe the bug**
A build of the current source fails locally because of this line:

https://github.com/MobiFlight/MobiFlight-Connector/blob/1ce922224de982b5cab544085ee292d52a2ede85/src/MobiFlightConnector/UI/MainForm.cs#L450

The #ifdef is hiding the awaited call in the function, which winds up being empty, so the compiler complains about the async being unneeded.

neilenns:
### Summary
A compile-time #if check was omitting all of the update checking code in debug builds. This is fine, but the function definition includes `async` so the compiler was failing the build with an error about how `async` was unnecessary.

Resolved the issue by adding an `#else` and doing an await of an immediately completing task to make the compiler happy in debug builds.

### Acceptance criteria
- [X] No more compiler warnings in debug builds
- [X] Update checks still happen in release builds

### DoD Checklist
None required
     
## Related issues
Fixes #3258

### DIFF
--- a/src/MobiFlightConnector/UI/MainForm.cs
+++ b/src/MobiFlightConnector/UI/MainForm.cs
@@ -449,8 +449,10 @@ namespace MobiFlight.UI
 
         private async void MainForm_Shown(object sender, EventArgs e)
         {
+#if (DEBUG)
+            await Task.CompletedTask; // Avoid compiler errors about an unneeded "async" modifier on the function definition in debug builds
+#else
             // Check for updates before loading anything else
-#if (!DEBUG)
             try
             {
                 await AutoUpdateChecker.CheckForUpdate(true);
--- a/src/MobiFlightConnector/UI/MainForm.cs
+++ b/src/MobiFlightConnector/UI/MainForm.cs
@@ -449,9 +449,7 @@ namespace MobiFlight.UI
 
         private async void MainForm_Shown(object sender, EventArgs e)
         {
-#if (DEBUG)
-            await Task.CompletedTask; // Avoid compiler errors about an unneeded "async" modifier on the function definition in debug builds
-#else
+#if (!DEBUG)
             // Check for updates before loading anything else
             try
             {
@@ -459,6 +457,8 @@ namespace MobiFlight.UI
             } catch (Exception ex) {
                 Log.Instance.log($"Error checking for updates: {ex.Message}", LogSeverity.Error);
             }
+#else
+            await Task.CompletedTask; // Avoid compiler errors about an unneeded "async" modifier on the function definition in debug builds
 #endif
         }
 


src/MobiFlightConnector/UI/MainForm.cs
```diff
@@ -449,14 +449,16 @@ private void CreateBindingStatusNotification(Project project)
...
             try
             {
                 await AutoUpdateChecker.CheckForUpdate(true);
             } catch (Exception ex) {
                 Log.Instance.log($"Error checking for updates: {ex.Message}", LogSeverity.Error);
             }
+#else
+            await Task.CompletedTask; // Avoid compiler errors about an unneeded "async" modifier on the function definition in debug builds
 #endif
         }
```
DocMoebiuz, line 463: This is a simple fix. no doubt. But also feels like a hack.
What we are trying to do is => Skip checking for update if running DEBUG target.

So why don't we actually add a proper test or do that even inside of the AutoUpdateChecker.

```
var skipAutoUpdateCheck = false;
#if (DEBUG)
 skipAutoUpdateCheck = true;
#endif 

if (skipAutoUpdateCheck) return;
```

DocMoebiuz (changes requested):

src/MobiFlightConnector/UI/MainForm.cs
neilenns: Isn't that still just a hack? It's a variable that only exists to prevent something from happening in debug builds, that the compiler will wind up optimizing away in release builds, using two different ifs: one at compile time and one at runtime.

My change is just a single line that won't even get included in releases.

neilenns: I originally wanted to #if out the entire function, btw, but that broke the mainform.designer.cs file since it has the event handler addition in it.

ColinM9991: Another option

```csharp
#if (!DEBUG)
        private async void MainForm_Shown(object sender, EventArgs e)
        {
            // Check for updates before loading anything else
            try
            {
                await AutoUpdateChecker.CheckForUpdate(true);
            } catch (Exception ex) {
                Log.Instance.log($"Error checking for updates: {ex.Message}", LogSeverity.Error);
            }
        }
#else
        private void MainForm_Shown(object sender, EventArgs e)
        {
            
        }
#endif
```

This may be irrelevant with the newer .NET Core `IHost` and `IEnvironment` as the preprocessor directives can be replaced with actual logic.

For instance, in the future we could have

```csharp
public interface IUpdaterChecker {
    Task CheckUpdateAsync();
}

public sealed class AutoUpdaterChecker : IUpdaterChecker {
    public async Task CheckUpdateAsync() {
        // Do stuff
    }
}

public sealed class FakeUpdaterChecker : IUpdaterChecker {
    public ask CheckUpdateAsync() {
        return Task.CompletedTask;
    }
}

private readonly IUpdaterChecker _checker;
private async void MainForm_Shown(object sender, EventArgs e)
{
    // Check for updates before loading anything else
    try
    {
        await _checker.CheckUpdateAsync(true);
    } catch (Exception ex) {
        Log.Instance.log($"Error checking for updates: {ex.Message}", LogSeverity.Error);
    }
}
```

Where `IUpdaterChecker` is resolved as part of dependency injection based on `IEnvironment.IsProduction() == true`. This would be the more idiomatic way as then you can unit test in isolation while faking dependencies

neilenns: I'm just going to close this since it touches the .NET side of things. Will wait until after the .NET 10 migration is done and will re-open one of your two suggested methods.

PR #3266, opened 2026-08-11
Fix ProSim action localization

### LINKED ISSUE #3235: Localized strings missing in the ProSim action type
**Describe the bug**
There are strings missing for English in the ProSim input action type

**Steps To Reproduce**
1. Create an MSFS project with ProSim enabled
2. Create an input action and select the ProSim action type

**Expected behavior**
All the strings are shown in English

**Actual behavior**
Placeholder text is shown.

Image

neilenns:
### Summary
Fixes a typo in the ProSim localized string path that caused existing strings to not get used.

### Screenshots / recordings

Before

image

After

image

### Acceptance criteria
- [X] Localized strings show

### DoD Checklist

- [X] i18n - All user-facing strings are translated
     
## Related issues
Fixes #3235 

### Notes

Super easy fix, the strings were there, just a capital `S` was required.


DocMoebiuz:
image

In the overview, Github shows you how many tasks are done.

I would simply remove the checkboxes that you don't plan to check.

image

The template currently says strike them out - which is also fine. After using the template for a couple of weeks, I think it looks cluttered when struck through, and instead, i simply remove the checkboxes which don't apply.

DocMoebiuz (changes requested):
Check my comment on the PRs description and its checkboxes

neilenns:
@DocMoebiuz DoD updated

DocMoebiuz (approved):

DocMoebiuz (changes requested):
@neilenns there is apparently no frontend test available to test for correct title and description - otherwise it would have failed and you didn't add one either. So the checkbox for frontend test is not correct. 

If our other panels are also not tested then it is OK without testing headlines. We plan to create an independent test to make sure all placeholders are actually resolving in the translation files.

There is also no unit test involved... so that checkbox is not relevant either.

neilenns:
Fixed the DoD list. I marked it originally because there were existing unit tests for the dialog that still passed after the change.

DocMoebiuz (approved):

PR #3281, opened 2026-08-13
Fix size Dropdowns for controller and device in input config

### LINKED ISSUE #3233: Dropdowns for controller and device in input config dialog should be wider
**Describe the bug**
The dropdown for the device in the input config dialog is narrow and cuts off text, especially with game controllers we ship known joystick configs for.

**Steps To Reproduce**
1. Add a Honeycomb Alpha to your system
2. Create a new input config
3. Select the Alpha as your controller
4. Open the device dropdown

**Expected behavior**
A nice list of all the connected devices where you can read what everything is

**Actual behavior**
The names are chopped off. You can't tell when you are selecting up vs down or press vs release for any of the switches or buttons.

**Additional context**
We have lots of room in the dialog, let's make both the controller and device dropdowns wider

Image

Here's the controller names getting cut off:

Image

OK434:
### Summary

Modify the controller and device dropdown menus in the input settings to be flexible.
### Screenshots / recordings

image
image

### Acceptance criteria

- [x] fix size Dropdowns for controller and device in input config

### DoD Checklist

 - [x] Frontend
- [x] Frontend tests

     
## Related issues

fixes #3233

### DIFF
--- a/src/MobiFlightConnector/frontend/src/components/ComboBox.tsx
+++ b/src/MobiFlightConnector/frontend/src/components/ComboBox.tsx
@@ -71,11 +71,11 @@ const ComboBox = <T,>({
           variant="outline"
           role="combobox"
           aria-expanded={open}
-          className={cn("justify-between h-8", widthClass)}
+          className={cn("justify-between h-8", "w-[250px]")}
           disabled={disabled}
           {...props}
         >
-          <span className={cn(widthClass, "truncate text-sm text-left")} title={label}>
+          <span className={cn(widthClass, "text-sm text-left")} title={label}>
             {label}
           </span>
           <IconChevronDown className="opacity-50" />
--- a/src/MobiFlightConnector/frontend/src/components/ComboBox.tsx
+++ b/src/MobiFlightConnector/frontend/src/components/ComboBox.tsx
@@ -63,6 +63,7 @@ const ComboBox = <T,>({
     ? getLabel(selectedInItems)
     : placeholder
 
+  const popoverContentMinWidth = widthClass.startsWith("w-") ? `min-${widthClass}` : "min-w-50"
   return (
     <Popover open={open} onOpenChange={setOpen} modal={true}>
       <PopoverTrigger asChild>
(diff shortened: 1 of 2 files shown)


neilenns:
@OK434 This fixes the size of the dropdown when it's open, but when it is closed it looks like it will still be cutting off the text. Can we also make the size of the dropdown when it's closed wider?

src/MobiFlightConnector/frontend/src/components/ComboBox.tsx
```diff
@@ -71,11 +71,11 @@ const ComboBox = <T,>({
           variant="outline"
           role="combobox"
           aria-expanded={open}
-          className={cn("justify-between h-8", widthClass)}
+          className={cn("justify-between h-8", "w-[250px]")}
```
DocMoebiuz: this will change the style for all combobox instances. and that is not what we want.

the idea of "widthClass" was that we can pass in a tailwind class for the button and the popover together. so that we can force both widths to be the same. 

w-max is not a tailwind class.
w-full will also not work correctly for the popout because it is an overlay and has a very different root than our button.

src/MobiFlightConnector/frontend/src/components/ComboBox.tsx
```diff
@@ -71,11 +71,11 @@ const ComboBox = <T,>({
...
           disabled={disabled}
           {...props}
         >
-          <span className={cn(widthClass, "truncate text-sm text-left")} title={label}>
+          <span className={cn(widthClass, "text-sm text-left")} title={label}>
```
DocMoebiuz: truncate is important, just in case the controller name is longer than anticipated. but helps to prevent breaking the layout by overflowing content.

DocMoebiuz:
> @OK434 This fixes the size of the dropdown when it's open, but when it is closed it looks like it will still be cutting off the text. Can we also make the size of the dropdown when it's closed wider?

@neilenns, there is a limitation of the combobox the way it is done right now (shadcn) - it is a composition of a button (content which is always visible, the selected item), and a popover (content when dropdown visible and all options).

The button can dynamically resize to content. so you could just not assign a width at all. it is limited by its parent component.
The popover can't because its parent is body, or some other dom element pretty high up the dom tree. 
The popover requires an absolute width - there is no other way.

We truncate long values to preserve layout.
We still have some space left to make the comboboxes in this view wider.
But we will have to pick a value, e.g. w-64 (instead of w-50) and there is still a chance that potentially a controller could have a longer name that won't fit.

99% of the time we will be using Scan for input. 

And it looks like the problem is longer controller device names.
Let's go with w-64. You can always hover.

neilenns:
@DocMoebiuz the combo boxes should at least be wide enough when closed to display the names we ship with MobiFlight. Right now, if you use a Honeycomb Alpha or Bravo you can't see enough of the text to know what you've selected.

This shows the experience when it is open, it's the same experience when it is closed. I shouldn't have to hover to see text we ship with the app. There's *tons* of room in the dialog to make them wider.

image

neilenns:
@DocMoebiuz The current version of this PR looks fantastic.

image

image

neilenns:
The latest push is worse than the previous one:

image

The change in https://github.com/MobiFlight/MobiFlight-Connector/pull/3281/commits/4c7fb63b6c3b246ec33bd07ad816da82275ab322 looked perfectly fine, why can't we use that?

DocMoebiuz:
> The latest push is worse than the previous one:
> 
> image
> The change in [4c7fb63](https://github.com/MobiFlight/MobiFlight-Connector/commit/4c7fb63b6c3b246ec33bd07ad816da82275ab322) looked perfectly fine, why can't we use that?

because it breaks other parts of the app that you haven't checked

PR #3284, opened 2026-08-14
Refactor the update checks in the main form

ColinM9991:
### Summary

This PR provides an example on one approach to refactoring the update checks.

This demonstrates dependency injection, working with `IEnvironment` and cleans up one of the preprocessor directives for update checks.

### DoD Checklist

- [x] Unit tests available
  - [x] .NET backend
~~  - [ ] Frontend~~
~~- [ ] Frontend tests~~
~~- [ ] i18n - All user-facing strings are translated~~
~~  - [ ] core (en,de,es)~~
~~  - [ ] additional langs~~
~~- [ ] documentation~~
~~  - [ ] User docs (docs.mobiflight.com)~~
~~  - [ ] Developer docs (e.g., readme.md)~

### DIFF
--- a/src/MobiFlightConnector/Base/UpdateChecker/StubbedAutoUpdateChecker.cs
+++ b/src/MobiFlightConnector/Base/UpdateChecker/StubbedAutoUpdateChecker.cs
@@ -1,0 +1,11 @@
+﻿using System.Threading.Tasks;
+
+namespace MobiFlight.Base.UpdateChecker;
+
+public sealed class StubbedAutoUpdateChecker : IAutoUpdateChecker
+{
+    public Task CheckForUpdateAsync(bool silent = false)
+    {
+        return Task.CompletedTask;
+    }
+}
--- a/src/MobiFlightConnector/Base/UpdateChecker/AutoUpdateChecker.cs
+++ b/src/MobiFlightConnector/Base/UpdateChecker/AutoUpdateChecker.cs
@@ -1,14 +1,14 @@
-﻿using MobiFlight.UI.Dialogs;
-using System;
+﻿using System;
 using System.Diagnostics;
 using System.IO;
 using System.Reflection;
 using System.Threading.Tasks;
 using System.Windows.Forms;
+using MobiFlight.UI.Dialogs;
 
-namespace MobiFlight.UpdateChecker
+namespace MobiFlight.Base.UpdateChecker
 {
-    static class AutoUpdateChecker
+    public sealed class AutoUpdateChecker : IAutoUpdateChecker
     {
         internal enum UpdatePath
         {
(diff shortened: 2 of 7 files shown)


src/MobiFlightConnector/Base/UpdateChecker/StubbedAutoUpdateChecker.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Threading.Tasks;
+
+namespace MobiFlight.Base.UpdateChecker;
+
+public sealed class StubbedAutoUpdateChecker : IAutoUpdateChecker
```
DocMoebiuz: I would call this "AlwaysSkippingAutoUpdateChecker" or similar to tell what it is doing.

DocMoebiuz:
I really like the option to do to DI - i am wondering if this is the right use case for illustrating it though. It is a lot of boiler plate to get something so simple done. The auto update checker also requires some refactoring.

Is this still considered KISS?

We just wanna say: skip check if in dev mode.

You are also initiating a variable using the compiler options. I am still a fan of the suggestion that i made here https://github.com/MobiFlight/MobiFlight-Connector/pull/3264/changes#r3764599540

Alternatively, we could also move the logic to skip the check inside the AutoUpdateChecker - but the behavior is then moved down one level and harder to discover if somebody wonders why they don't see the update dialog.

Also this prod vs debug behavior for the update? Should this really be guarded by compiler directives? Isn't it a simple runtime  configuration? The purpose of guarding code by ifdef blocks is really that these will never make it into the compiled application. Is that really needed for a dev build?

ColinM9991:
> It is a lot of boiler plate to get something so simple done

This updater refactor is just one example but moving from non-DI to DI is going to incur a lot of boilerplate. There will need to be interfaces and abstractions along with registration of services in a DI container such as `IServiceCollection`. Upon switching to a full DI system, the boilerplate moves from individual classes (from instantiating services and classes) to the application startup (registering services).

> You are also initiating a variable using the compiler options

> Should this really be guarded by compiler directives? Isn't it a simple runtime configuration? The purpose of guarding code by ifdef blocks is really that these will never make it into the compiled application. Is that really needed for a dev build?

`IHostEnvironment` is more commonly used in ASP.NET Core and the value is typically set via an environment variable `ASPNETCORE_ENVIRONMENT`. The preprocessor option is a way of achieving a similar approach without the need for environment options or without the need to specify a preprocessor directive in every service that has conditional flow. Doing this means `IHostEnvironment` can be injected in services which have conditional execution based on where they're running.

It isn't really best practice these days to use compiler preprocessor directives for conditional execution paths because complex rules or many `#ifdef .. #endif` references can result in untestable code and edge cases that are hard to pin down due to code being omitted from certain builds.

Another option is to have a configuration file like
```json
{
    "autoUpdateOnStart": false
}
```

and register an `IOptions<T>` which is injected into `MainForm`. This can then be used like

```csharp
private readonly IOptions<MainFormSettings> _settings;

if (_settings.Value.AutoUpdateOnStart)
{
    await autoUpdaterChecker.CheckUpdateAsync();
}
```

That continues to be testable because the unit tests can mock `IOptions<MainFormSettings>` and `IUpdaterChecker` to ensure it runs when it should, while skipping the update check in correct conditions.

PR #3302, opened 2026-08-19
Ensure installer can update .NET10 based MobiFlight version

ColinM9991:
### Summary
1 Reorders the initialization order of the process matching to ensure `ProcessName` has a value assigned to it before it is used in `FilePathCandidates`
2. Ensures the installer is compiled as a standard binary, whereas `publish` may result in a ClickOnce binary being produced instead.
3. Removes the MobiFlight user.config reading from the installer. The beta flag is now passed directly by MobiFlight.
4. Removes the unused expert mode UI from the installer

To test the installer upgrade path I created a Windows virtual machine and downloaded MobiFlight via the website. I then downloaded Caddy to host a local webserver containing a custom manifest which pointed to a private build of the latest changes in this PR.

Upon the next launch of MobiFlight 11.2.0 it checked my custom manifest for updates, finding the .NET Core beta update. I commenced the update and confirmed that both MobiFlight and the installer updated successfully. There were no issues with MobiFlight getting stuck in an update cycle as we've previously observed during the .NET Core betas. The user configuration file was also migrated successfully thus preserving all settings.

### Acceptance criteria

- [x] user settings are upgraded correctly
- [x] beta version check is still working like before

### DoD Checklist
- [x] Verified that the installer compilation output is as expected
- [x] Tested and verified the installation process manually

### Related issues
#2872 

### Notes
- [x] Fully qualified paths used for MFConnector process
- [x] Removed expert settings option


DocMoebiuz (approved):
weird

DocMoebiuz (approved):
Great!

PR #3308, opened 2026-08-24
Fix HID joystick exclusion handling

### LINKED ISSUE #3239: Ignoring WingFlex devices in Settings is not saved properly
**Describe the bug**
[User report in Discord](https://discord.com/channels/608690978081210392/1532745768023232512)

"In the devices window I keep unchecking the cubes but after every new start of Mobi[Flight] they are checked again"

**Steps To Reproduce**
1. Have WingFlex EFIS, FCU and RMP connected to PC
2. Start MobiFlight, disable the wingflex devices in Settings > Peripherals by unchecking them
3. Close Settings
4. Open Settings again
5. Observe the devices are again checked.

**Expected behavior**
Setting should stick.

**Actual behavior**
Device enabled -setting is not saved. All HID devices can be disabled in Peripherals, but maybe the custom WingFlex support that was added is not expecting that the device could be detected but the user decided to use the WingFlex app instead? 

Image

Might be worth checking if the other "custom supported" HID devices like the Bodnar/Authentikit boards also trigger this bug?

**Details**
 - MobiFlight Version: 11.2.0
 - FlightSim: roger
 - AddOn: wilco

**Additional context**

OK434:
### Summary

Fixed HID joystick exclusion handling by checking the user's ExcludedJoysticks settings before connecting HID controllers. Added unit tests to verify that excluded devices are ignored while non-excluded devices are still connected.
### Acceptance criteria

- [x] Excluded HID devices are not connected.
- [x] Non-excluded HID devices are still connected.

### DoD Checklist

- [x] Unit tests available
  - [x] .NET backend

## Related issues

fixes #3239

### DIFF
--- a/src/MobiFlightConnector/MobiFlight/Joysticks/JoystickManager.cs
+++ b/src/MobiFlightConnector/MobiFlight/Joysticks/JoystickManager.cs
@@ -33,7 +33,7 @@ namespace MobiFlight
         public readonly List<JoystickDefinition> Definitions = new List<JoystickDefinition>();
         public event EventHandler Connected;
         public event ButtonEventHandler OnButtonPressed;
-        private readonly Timer PollTimer = new Timer(); 
+        private readonly Timer PollTimer = new Timer();
         private readonly System.Collections.Concurrent.ConcurrentDictionary<string, Joystick> Joysticks = new System.Collections.Concurrent.ConcurrentDictionary<string, Joystick>();
         private readonly List<Joystick> ExcludedJoysticks = new List<Joystick>();
         private IntPtr Handle;
@@ -79,7 +79,7 @@ namespace MobiFlight
             var schemaFilePath = "Joysticks/mfjoystick.schema.json";
 
             var rawDefinitions = JsonBackedObject.LoadDefinitions<JoystickDefinition>(
-                jsonFiles, 
+                jsonFiles,
                 schemaFilePath,
                 onSuccess: (joystick, definitionFile) => Log.Instance.log($"Loaded joystick definition for {joystick.InstanceName}", LogSeverity.Debug),
                 onError: () => LoadingError = true
@@ -203,7 +203,7 @@ namespace MobiFlight
                 var diJoystick = new SharpDX.DirectInput.Joystick(di, d.InstanceGuid);
                 var productId = diJoystick.Properties.ProductId;
                 var vendorId = diJoystick.Properties.VendorId;
-                
+
                 // Check if this device should be handled later as an HID controller
                 if (HidControllerFactory.CanCreate(d.InstanceName))
                 {
@@ -218,7 +218,7 @@ namespace MobiFlight
 
                 // Use factory to create appropriate controller instance
                 var js = ControllerFactory.Create(d, diJoystick, vendorId, productId, definition, WSServer);
-                
+
                 // If factory returns null, create a standard Joystick
                 if (js == null)
                 {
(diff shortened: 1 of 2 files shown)


src/MobiFlightConnector/MobiFlight/Joysticks/JoystickManager.cs
```diff
@@ -293,16 +297,21 @@ private void ConnectHidController()
...
-                        if (!Joysticks.TryAdd(joystick.Serial, joystick))
+                        // Check against exclusion list
+                        if (ShouldConnectHidJoystick(
+                                               joystick.Name,
+                                            settingsExcludedJoysticks))
                         {
-                            Log.Instance.log($"Error adding HID device: {definition.InstanceName} / {joystick.Serial}. Likely Joystick Serial conflict.", LogSeverity.Error);
-                            return;
+                            joystick.Connect(new IntPtr());
+                            Joysticks.TryAdd(joystick.Serial, joystick);
```
ColinM9991: Are the `OnButtonPressed` and `OnDisconnected` events intentionally removed here?

src/MobiFlightConnector/MobiFlight/Joysticks/JoystickManager.cs
```diff
@@ -293,16 +297,21 @@ private void ConnectHidController()
...
+                        }
+                        else
+                        {
+                            Log.Instance.log(
+                                $"Ignore attached joystick device: {joystick.Name}.",
+                                LogSeverity.Info
+                            );
```
DocMoebiuz: Apply early return/continue pattern

src/MobiFlightConnector/MobiFlight/Joysticks/JoystickManager.cs
```diff
@@ -293,16 +297,21 @@ private void ConnectHidController()
                         var joystick = HidControllerFactory.Create(definition);
 
                         if (joystick == null) return;
-
-                        joystick.Connect(new IntPtr());
-                        joystick.OnButtonPressed += Js_OnButtonPressed;
-                        joystick.OnDisconnected += Js_OnDisconnected;
```
DocMoebiuz, line 299: why did you remove these handler registrations?

src/MobiFlightConnector/MobiFlight/Joysticks/JoystickManager.cs
```diff
@@ -293,16 +297,21 @@ private void ConnectHidController()
...
-                        if (!Joysticks.TryAdd(joystick.Serial, joystick))
+                        // Check against exclusion list
+                        if (ShouldConnectHidJoystick(
+                                               joystick.Name,
+                                            settingsExcludedJoysticks))
                         {
-                            Log.Instance.log($"Error adding HID device: {definition.InstanceName} / {joystick.Serial}. Likely Joystick Serial conflict.", LogSeverity.Error);
-                            return;
```
DocMoebiuz, line 303: Why did you remove the error handling on adding the joystick to the dictionary?

src/MobiFlightConnector/MobiFlight/Joysticks/JoystickManager.cs
```diff
@@ -269,13 +269,17 @@ private JoystickDefinition GetJoystickDefinition(string instanceName, string pro
             // Try to get definition by product name first, then by product ID
             return GetDefinitionByInstanceName(productName) ?? GetDefinitionByProductId(vendorId, productId);
         }
-
+        internal static bool ShouldConnectHidJoystick(string joystickName,List<string> excludedJoysticks)
+        {
+            return !excludedJoysticks.Contains(joystickName);
+        }
```
DocMoebiuz: I like the idea to improve readability.

`IsExcludedJoystick` could be a better way of thinking about this - especially because it would fit the early return / continue pattern.

I would also update the existing code that does the same check but  currently doesn't use this method.

DocMoebiuz (changes requested):
Please review the existing exclusion logic. It looks to me like if we should test for exclusion prior to connecting to the controller. This might actually be a flaw/bug in the existing code base.

DocMoebiuz (approved):

gasher9999:
I tried this build and it is not seeing the Wingflex RMP Cube as a controller. 11.2.0 detects it but exclusion won't hold.

PR #3323, opened 2026-08-30
Use Serilog for logging

ColinM9991:
### Summary

Using Serilog for logging. It was originally thought that logging was saturating the thread pool as that too used Task.Run. Offloading to Serilog allows Serilog to handle log events in a more optimised fashion.
     
## Related issues

Addresses some threadpool bound work as part of #3324 

### Out-of-scope

A full rewrite of logging has not be implemented in this change in order to keep the scope short.


DocMoebiuz (changes requested):
Add dependency to our list in the about dialog.

ColinM9991:
> Add dependency to our list in the about dialog.

Added

DocMoebiuz (approved):
Thanks!

PR #3338, opened 2026-09-02
Fix typos

antonkesy:
### Summary

AFAIK there is no issue for typos, therefore I don't have any to reference.

This PR fixes all typos I could find in: comments, docs, strings and the code.
There should be no functional change (unless someone hard-coded the typos in the UI into some tests/automation)

The PR template does not really fit for this kind of change, so I did not continue filling it out (for now)

If you don't want this kind of PR, feel free to just close it :)

### Screenshots / recordings

N/A

### Acceptance criteria

- [ ] X added
- [ ] Y still working
- [ ] Z changed

### DoD Checklist

- [ ] Unit tests available
  - [ ] .NET backend
  - [ ] Frontend
- [ ] Frontend tests
  - [ ] core (en,de,es)
  - [ ] additional langs
- [ ] documentation
     
## Related issues

fixes #ISSUE_NUMBER


DocMoebiuz:
@antonkesy - hehe, I like people who care about details! Thanks for submitting - the PR has quite a big blast radius in terms of touched files. We are currently in the process of rewriting / changing some central code parts - I will hold off with this PR until we are done.

DocMoebiuz:
I would not patch typos in CmdMessenger library because it is an external lib - reminds me that we should include it as nuget package... which unfortunately is not available yet.

Scope should only be our own code.

antonkesy:
> I would not patch typos in CmdMessenger library because it is an external lib - reminds me that we should include it as nuget package... which unfortunately is not available yet.
> 
> Scope should only be our own code.

🫡 make sense, my bad ^^ somehow misinterpreted this 

I've removed the `CmdMessenger` related fixes (and actually opened a PR in the `CmdMessenger` repo itself)

lib/FSUIPCClient.XML
DocMoebiuz, line 1: Same here, this is an external library that we are using.  I am afraid it is gonna be overwritten next time we update the library dependency.

src/MobiFlightConnector/MobiFlight/MobiFlightLedModule.cs
```diff
@@ -152,7 +152,7 @@ static public byte CalculatePositionInString(string value, byte mask)
             // then we allocate a digit for it correctly
             var lastCharacter = '.';
 
-            // out inital length is 0, because we start on the left
+            // out initial length is 0, because we start on the left
```
DocMoebiuz: it should read as 
"our initial"

DocMoebiuz (changes requested):
two more comments

antonkesy:
Wupsi
Did a rebase pushed inbetween which automatically closed the PR 😞

src/MobiFlightConnector/Base/DeviceReference.cs
```diff
@@ -4,7 +4,7 @@ public class DeviceReference
     {
         public virtual string Name { get; set; }
 
-        // temporay property, will be removed before merging with the main branch
+        // temporary property, will be removed before merging with the main branch
```
antonkesy, line 7: 👀

lib/FSUIPCClient.XML
antonkesy: Now I should have removed all external stuff from my PR 👀 I hope at least 👀

src/MobiFlightConnector/MobiFlight/MobiFlightLedModule.cs
antonkesy: Done 👍🏻

antonkesy:
Sorry for the noise and the fresh changes ^^

Issue #3220, opened 2026-08-05
Arduino Mega 2560 only detected correctly after USB unplug/replug


RemcoteHofstee:
Describe the bug

MobiFlight detects an Arduino Mega 2560 COM port correctly, but on the first
MobiFlight start after boot the board is shown as a Compatible Module.

Firmware version, serial number and device configuration are empty.

After physically unplugging and reconnecting the USB cable, the same board is
detected correctly immediately.

Environment

- MobiFlight Connector 11.2.0
- Firmware 3.1.4
- Windows 11 Pro
- Arduino Mega 2560
- USB hardware ID: VID_2341 / PID_0042
- Driver: Microsoft usbser.inf

Steps to reproduce

1. Install MobiFlight firmware on an Arduino Mega 2560.
2. Leave the Mega connected by USB.
3. Restart Windows.
4. Start MobiFlight.
5. Open the MobiFlight Modules screen.
6. The board is shown as Compatible Module without firmware, serial number or
   device configuration.
7. Physically unplug and reconnect the USB cable.
8. The board is then detected correctly with firmware, serial number and
   configuration.

Expected behavior

The Mega should be detected correctly when MobiFlight starts while the board
is already connected.

Actual behavior

The COM port and board type are detected, but the first GetInfo response does
not contain firmware, serial number or configuration data.

Debug log on failed connection

MobiFlight detects COM10 and connects at 115200 baud with DTR enabled:

Port detected: COM10 Arduino Mega 2560
Connecting to COM10 - Baud rate: 115200, DtrEnable: True
Retrieved board: Arduino Mega 2560, Arduino Mega 2560, /, .
Disconnected Arduino Mega 2560:COM10.

Debug log after unplug/replug

Port detected: COM10 Arduino Mega 2560
Connecting to COM10 - Baud rate: 115200, DtrEnable: True
Retrieved board: MobiFlight Mega, Nano, 3.1.4/3.1.4, SN-G5E-45E.
Registering module Nano:COM10.

Troubleshooting already performed

- Reproduced with two different bare Arduino Mega 2560 boards
- No cockpit hardware connected
- Tested directly on motherboard USB 2.0 port
- Tested through a powered USB hub
- Tested with different USB ports
- Changed COM port number
- Clean Windows startup with non-Microsoft services disabled
- No other serial-port applications running
- Pressing the physical reset button does not help
- pnputil restart-device does not help
- pnputil disable-device / enable-device does not help
[...]

Issue #546, opened 2021-11-01
MobiFlight Board shows garbled Board information


MobiFlight-Admin:
**Describe the bug**
Sometimes with longer device configs, a mobiflight module shows garbled board information where parts of the config information is displayed as type and the name contains wrong information or is empty.

**Steps To Reproduce**
1. Configure many devices
2. Configure 2 LCDs
3. Upload config
4. Restart MF
5. Go to settings and check the modules information

**Expected behavior**
It should show Name, Version and Type correctly

**Actual behavior**
Type shows config parts, name is empty.

**Details**
 - MobiFlight Version: [e.g. 8.2.0]
 - FlightSim: [e.g. MSFS2020]
 - AddOn: [e.g. Default Aircraft]

**Additional context**

DocMoebiuz:
This is not happening with the current firmware (>1.12.2)

Issue #3269, opened 2026-08-12
More robust board detection is needed for supported boards when serial com port is taken by other software


JaimeLeon2:
**Is your feature request related to a problem? Please describe.**
Several apps have been identified that grab serial ports and don't release them.  This, all too often, causes Mobiflight modules (board with firmware loaded) to be reported simply as "Compatible", leaving the user clueless.

**Describe the solution you'd like**
A more robust board detection strategy should be implemented, where Mobiflight can distinguish between boards that do not have the firmware loaded (no response to initial ping), from boards that simply do not respond to anything (serial port taken by another app).  Boards connected to serial ports that do not respond to anything should be flagged maybe as "Port busy" or "Access denied".

**Describe alternatives you've considered**
There is no other alternative inside MF Connector.  The Mobiflight modules with the serial port taken up by another app are reported as "Compatible" even though the firmware is loaded. 

**Additional context**
This generates too many support requests in Discord.

PR #3340, opened 2026-09-03
Retry GetInfo when the first command times out

### LINKED ISSUE #3321: Connected board not identified on first launch after a Windows cold boot — shown as "Compatible" with no firmware or serial
## Summary

A Raspberry Pi Pico that is connected, powered, and running working MobiFlight firmware is not identified on the first Connector launch after a Windows cold boot. It appears in the module list as `Compatible (Raspberry Pico)` with empty Firmware and Serial fields, as if no firmware were installed. Closing the Connector and launching it again — no reboot, no replug — identifies it correctly. This is repeatable.

Separately, with Debug-level logging enabled under the same cold-boot conditions, the Connector never finishes starting at all and has to be killed from Task Manager. At Info level it starts normally and only misidentifies the board.

## Environment

- MobiFlight Connector 11.2.0.5 (commit `388fc3a50e0b336fd6f038fad98b181296d36190`)
- Windows 11
- Four boards connected, all running MobiFlight firmware: three CH340-based Arduino Mega 2560 clones on COM4/COM6/COM8, and one custom Raspberry Pi Pico panel (VID_2E8A/PID_000A) on COM10
- The Pico's hardware ID matches several board definitions — the built-in `raspberrypi_pico.board.json` plus multiple Community ones. This is expected: every RP2040 board built with the stock Pico SDK enumerates with the same USB VID/PID.

## Why I do not think this is my board

Before filing, I checked the board itself, independently of the Connector:

- I sent the `kGetInfo` command to the board directly over its COM port, by hand, repeatedly — including at the exact moment the Connector's module list was showing it as `Compatible`. It answered every time in 4–15 ms with the correct type, name, serial number and firmware version.
- I also went through my own firmware's config handling, its inter-core message path and its I2C timing, and fixed several genuine problems I found there. None of them changed this symptom.

As far as I can tell the board is healthy and responsive the whole time the Connector is failing to identify it.

## Steps to reproduce

1. Connect a board running MobiFlight firmware that identifies correctly on a normal launch.
2. Fully restart Windows. A cold boot matters — relaunching the Connector alone does not reproduce it.
3. Launch the Connector as the first thing after boot, with the board already plugged in.
4. Open the Modules screen.

**Expected:** the board is identified with its type, firmware version and serial number.

**Observed:** the board is listed as `Compatible (<board name>)` with no firmware version and no serial number.

5. Close the Connector and launch it again — still no reboot, no unplug. The board is now identified correctly.

For the startup hang: repeat from step 2 with the log level set to Debug before launching. In my setup this reliably prevents startup from completing. I have four boards attached, which may matter; I have not tried it with a single board.

## Logs

**First launch after a cold boot, Info level — board misidentified:**

```
20:00:32(580) MobiflightModule.connect: Connected to Arduino Mega 2560 at COM8 of type MobiFlight Mega (DTR=>True).
20:00:32(580) MobiflightModule.connect: Connected to Raspberry Pico at COM10 of type MobiFlight RaspiPico (DTR=>True).
20:00:32(581) MobiflightModule.connect: Connected to Arduino Mega 2560 at COM4 of type MobiFlight Mega (DTR=>True).
20:00:32(583) MobiflightModule.connect: Connected to Arduino Mega 2560 at COM6 of type MobiFlight Mega (DTR=>True).
20:00:36(316) Adding attached Midi Device: MiniFuse 2 MIDI In
```

Nothing about module registration appears between those lines.

**First launch after a cold boot, Debug level — never finishes starting:**

```
19:54:17(112) SerialPortMonitor_PortAvailable: Port detected: COM6 / COM8 / COM10 / COM4
19:54:17(116) MobiFlightModule.Connect(): Connecting to COM8 / COM6 / COM4 / COM10
19:54:17(131..135) Connected to ... COM4, COM10, COM8, COM6
19:54:23(378) DeviceMonitor.EventWatcher_EventArrived: \\.\root\CIMV2:__AggregateEvent
```

That is the entire rest of the log. No `GetInfo` line for any of the four ports, and nothing further is ever written.

**A successful second launch, Debug level, for comparison:**

```
19:49:01(846) MobiFlightModule.Connect(): Connecting to COM10 - Baud rate: 115200, DtrEnable: True
19:49:01(864) Connected to Raspberry Pico at COM10 of type MobiFlight RaspiPico (DTR=>True).
19:49:03(128) MobiFlightModule.GetInfo(): Retrieved board: Radio Panel RaspiPico, PICO RADIO, 1.1.2/3.1.4, SN-50436379484F6E1C.
19:49:03(165) MobiFlightCache.RegisterModule(): Registering module PICO RADIO:COM10.
19:49:08(148) CheckIfLookUpFinished: End looking up connected modules. 4 found.
```

I have full copies of both logs and can attach them.

## Related

**#3220** looks like the same misidentification on different hardware: closed, filed against the same 11.2.0 release, a single bare Arduino Mega 2560 with nothing else connected, `Compatible` with empty firmware and serial on the first launch after a boot, cured by unplugging and reconnecting the USB cable. Its log contains the line

```
Retrieved board: Arduino Mega 2560, Arduino Mega 2560, /, .
```

— board type present, version and serial empty. That fingerprint matters for the section below.

**#3269** notes that "Compatible" is an overloaded bucket covering several distinct failures, which matches what I ran into.

---

## What I tried

I built the Connector from your source at tag `11.2.0.5` and added logging around `Connect()` and `GetInfo()`, at Info level only, since Debug logging prevents the startup I need to observe from completing. Every run below is a separate Windows cold boot with the same four boards attached.

**1. Stock build, unmodified.** Reproduces the symptom, so the build is a faithful copy of the released application.

**2. Instrumented build.** The Pico's *first* info command times out while the three Megas succeed in the same run:

```
[COM8]  GetInfo() first command result: Ok=True,  elapsed=1074ms.
[COM4]  GetInfo() first command result: Ok=True,  elapsed=1074ms.
[COM6]  GetInfo() first command result: Ok=True,  elapsed=1075ms.
[COM10] GetInfo() first command result: Ok=False, elapsed=3317ms.
[COM10] Disconnecting (no MobiFlight firmware detected).
```

**3. Same build with the Pico's `ConnectionDelay` raised from 1250 ms to 4000 ms**, changing nothing else. No effect — `Ok=False, elapsed=2971ms`. So the board is not simply being asked too early.

**4. Same build plus a retry when the first info command fails.** The board is identified correctly:

```
[COM10] GetInfo() first command result: Ok=False, elapsed=3135ms.
[COM10] GetInfo() first command failed, retry 1 (timeout=2500ms).
[COM10] GetInfo() retry 1 result: Ok=True, elapsed=1563ms.
```

One retry was enough, on the first attempt, every time I have run it.

## What the numbers show

Two things stand out in the instrumented logs, and both are about the Connector's own startup rather than about any board.

**Everything is far slower than nominal during that first startup.** `Thread.Sleep(Board.Connection.ConnectionDelay)` in `Connect()` overshoots enormously: a 2000 ms sleep logged as starting at 09:36:32 returned at 09:37:00, and the difference between a 2000 ms and a 4000 ms sleep vanished entirely — all four threads woke within 3 ms of each other. The info commands that do succeed take 900–1075 ms for what is a few bytes each way; my board answers the same request in 4–30 ms when I send it by hand.

**The measured waits exceed their own timeout.** `CommandTimeout` is 2500 ms, yet the failing calls report 2971, 3135 and 3317 ms elapsed. The waiting thread is not getting scheduled promptly.

So the picture I end up with is that the first launch after a cold boot is heavily loaded, a fixed 2500 ms timeout with no retry is not robust under that load, and my Pico is the board that happens to fall off the edge. I cannot tell you why it is that board and not one of the Megas — that part I could not work out.

One note if you read the logs: the timestamp on each line is written by the file appender inside its own `Task.Run`, so it is the time the line was flushed, not the time the event happened. Under a loaded startup the lines bunch together. The `elapsed=` values are measured inline and are the trustworthy numbers.

## Two things I noticed in the code

I am a user of this application, not a contributor to it. I read the source only to rule out my own firmware, and pass this along in case it is useful. You know this codebase far better than I do.

**`GetInfo()` has no retry when the command fails.** The `Config` getter a few lines above it does retry, with the comment "Sometimes first attempt times out", and `ExtraConnectionRetry` adds further attempts there for boards that need them. `GetInfo()` gets one attempt. That is the difference my experiment above changes.

**`GetInfo()` reads its fields from a second command whose result is never checked.** Separate from the above, and I did not hit this path myself, but it looks like it would produce the same "Compatible" outcome from a different direction:

```csharp
if (InfoCommand.Ok)
{
    // Workaround
    // the following two lines shall get removed
    // but at the moment something with the timing during startup is wrong.
    command = new SendCommand((int)MobiFlightModule.Command.GetInfo, (int)MobiFlightModule.Command.Info, CommandTimeout);
    InfoCommand = _cmdMessenger.SendCommand(command);

    devInfo.Type = InfoCommand.ReadStringArg();
    devInfo.Name = InfoCommand.ReadStringArg();
    devInfo.Serial = InfoCommand.ReadStringArg();
```

The `if` is evaluated on the first command's result, but the fields are read from the second, whose `Ok` is never checked. If the first succeeds and the second times out, the fields come back empty — which is exactly the fingerprint in #3220's log: `Retrieved board: Arduino Mega 2560, Arduino Mega 2560, /, .`, type present, version and serial empty. In my own runs the second command took up to 2055 ms against the same 2500 ms timeout, so it does not look far from failing either.

For context on why an empty version decides the label: `MobiFlightModule.Type` returns `Compatible` when `HasMfFirmware()` is false — which is just `!String.IsNullOrEmpty(Version)` — and more than one board definition matches the hardware ID.

## The startup hang — I could not explain this one

I have no explanation for the Debug-level hang and would rather not guess. The log stops right after all four ports are connected and before any info command is logged, and the two `__AggregateEvent` lines suggest a USB device change was being processed around then. Since the file appender's timestamps are flush times, it is also possible the application was still running and simply not draining its log queue. I could not attach a debugger to a hung session.

## Expected behaviour

A responsive board should be identified on the first launch after a cold boot without an unplug/replug cycle, and enabling Debug logging should not change whether the application finishes starting.

Happy to attach the full logs from every run above, or to test anything else that would help.

thaildhe172591:
### Summary
On the first launch after a Windows cold boot, a board that is running MobiFlight firmware can show up as `Compatible` with no firmware version and no serial number, and only restarting the Connector or replugging the USB cable fixes it. #3321 traced this to `GetInfo()`: it sent the info command once, and when that single attempt got no answer the board was treated as having no firmware. The `Config` getter a few lines above already retries a timed out first command for the same reason; `GetInfo()` never did.

This change sends the info command in a retry loop of up to three attempts (`GetInfoMaxAttempts`) that stops at the first reply.

It also removes the unconditional second read that `GetInfo()` did after a successful first reply. That read was a workaround for startup timing, which is what the loop now handles, so keeping it only meant asking an answered board once more. Removing it also removes the path that produced type without version — the fingerprint in the log of #3220 — because there is now a single reply the fields are read from.

`CommandTimeout` and `GetInfoMaxAttempts` are no longer `const`, so a test can shorten the wait rather than sit through the real 2.5 s timeout.

### Acceptance criteria
- [x] A board whose first info command times out is identified once the retry is answered
- [x] A board that answers immediately is asked exactly once
- [x] A port that never answers is still listed without firmware
- [x] The tests for those paths run in under a second

### DoD Checklist
- [x] Unit tests available
  - [x] .NET backend
  - ~~Frontend~~
- ~~Frontend tests~~
- ~~i18n - All user-facing strings are translated~~
- ~~documentation~~

## Related issues
fixes #3321
relates to #3220
relates to #3355

### Out-of-scope
The root cause of the lost command. The USB capture in the thread below shows the bytes leaving the host and the board answering nothing at all, so the loss is under the MobiFlight layer. This PR covers the symptom; it does not explain it.

The Debug-level startup hang described in #3321. It looks like the thread pool saturation that #3323 and #3328 addressed after the issue was filed, and nothing here touches it.

### Notes
Technical details: the loop reuses the same `SendCommand` for each attempt the way the `Config` getter does, and logs each timed out attempt at Debug level. `CommandTimeout` and `GetInfoMaxAttempts` change from `const` to instance fields; nothing outside `MobiFlightModule` referenced either of them.

On removing the second read: it arrived in 1f0fd45 for #546, where a first read could come back with parts of the config in it. That issue was closed with the note that it no longer happened on firmware above 1.12.2, and the current firmware is 3.1.4. If you would rather keep the second read anyway, say so and I will put it back behind the same `Ok` check instead.

Cost: a serial device that matches a board VID/PID but does not run the firmware now waits three timeouts (7.5 s) instead of one before it is listed as `Compatible`. Ports are probed in parallel, so startup grows by at most two timeouts in that case. Happy to drop the limit to two if that matters more than the third try.

Tests: `MobiFlightModuleTests` gets a `ScriptedTransport` that answers each written command with the next scripted reply or stays silent, so the real `CmdMessenger` runs into its own timeout. Three tests cover the immediate answer, the retry, and a port that never answers.

Local run: `dotnet build MobiFlightConnector.sln -c Release` and `dotnet test MobiFlightConnector.sln -c Release --no-build` on Windows 11 with the .NET SDK 10.0.302, rebased onto main at 20 commits newer than when this PR opened. MobiFlightUnitTests reports 1297 passed, 31 skipped, 2 failed; the two failures compare decimal formatting under this machine's comma-decimal culture and fail the same way on unmodified main. MobiFlightWwFcuUnitTests reports 492 passed. The three GetInfo tests take 873 ms together.

The cold-boot scenario itself was verified by @chegevarko1982 on the hardware from #3321, not here.


src/MobiFlightConnector/MobiFlight/MobiFlightModule.cs
```diff
@@ -909,13 +909,28 @@ public IModuleInfo GetInfo()
             var command = new SendCommand((int)MobiFlightModule.Command.GetInfo, (int)MobiFlightModule.Command.Info, CommandTimeout);
             var InfoCommand = _cmdMessenger.SendCommand(command);
 
+            // Sometimes first attempt times out, same as for the Config getter.
```
ColinM9991, line 912: A better option might be to wrap the logic in a for loop with a max of 3 attempts, breaking from the loop when it was successful, as it's less code duplication and more aligned with classic retry policy.

chegevarko1982:
I have the hardware and the cold-boot setup this PR could not be tested on, so I built two copies of the Connector and ran them.

Both builds are `main` at `2ea0d3c`. One is unmodified, the other has this PR's diff applied and nothing else. Four boards attached: three Arduino Mega 2560 clones and the Raspberry Pi Pico from #3321. Each run is its own Windows cold boot, log level Debug.

**Without this PR — reproduces:**

```
22:19:40.576 Connected to Raspberry Pico at COM10 of type MobiFlight RaspiPico (DTR=>True).
22:19:44.981 GetInfo(): Retrieved board: Raspberry Pico, Raspberry Pico, /, .
22:19:44.984 Disconnect(): Disconnected Raspberry Pico:COM10.
```

The board is listed as `Compatible` with no firmware and no serial.

**With this PR — identified:**

```
22:26:55.622 Connected to Raspberry Pico at COM10 of type MobiFlight RaspiPico (DTR=>True).
22:27:00.078 GetInfo(): GetInfo timed out on COM10. Retrying...
22:27:00.094 GetInfo(): Retrieved board: MobiFlight RaspiPico, PICO RADIO, 1.1.2/3.1.4, SN-50436379484F6E1C.
22:27:00.109 RegisterModule(): Registering module PICO RADIO:COM10.
```

One retry, first attempt, board identified. This exercises the retry path; the guard on the follow-up read was not hit in either run.

Two things from the numbers that may be worth knowing.

**The first command is lost, not slow.** It goes out about 1.25 s after connect and gets nothing for roughly 3.2 s. The retry is answered 16 ms later — same board, same port, same command. That also explains why raising `ConnectionDelay` from 1250 to 4000 ms changed nothing when I tried it: waiting longer does not help if the first exchange is being lost rather than being early. I have no explanation for the mechanism.

**The Debug-level hang from #3321 is gone.** Both runs above are at Debug, the level that used to stop startup from completing. The Connector started, listed the modules and shut down cleanly both times. Your out-of-scope note looks right — #3323 and #3328 fixed it.

One correction to my own issue text while I am here. I wrote there that the first launch after a cold boot is heavily loaded and that a fixed 2.5 s timeout is not robust under that load. On current `main` that load is gone — the four ports open in 20 ms and the three Megas are identified 4.0 s after connect, right on their nominal delay — and the Pico still misses its first command. So the load was not the cause, or not all of it. The retry is still what fixes it.

Full logs from both runs available if useful.

elral:
Very weird what happens, especially that it happens with a Pico. In contrast to AVR's the Pico does **not** get resetted on connect via the DTR line and is not busy with config loading or other.
I am more and more confident that we have different problems. A retry of reading the config seems at least for this problem here a solution, although we don't know the reason.

Monitoring the COM port with a serial sniffer app could help.

src/MobiFlightConnector/MobiFlight/MobiFlightModule.cs
thaildhe172591: Done in 5b824ca: GetInfo now loops up to GetInfoMaxAttempts (3) and breaks on the first reply. One side effect to be aware of: a port that matches a board by VID/PID but has no firmware now waits three timeouts (7.5 s) before it shows as Compatible, where it was one timeout before this PR. Happy to drop it to two if you would rather keep startup shorter for that case.

thaildhe172591:
Thanks for running both builds on the actual boards, that is the confirmation I could not produce here. After the review the retry is now a loop of up to three attempts, so the Debug line reads "GetInfo timed out on COM10, attempt 1 of 3." instead of "Retrying..." if you compare logs again.

chegevarko1982:
@elral suggested a sniffer, so I captured the USB traffic with USBPcap during a cold boot that reproduces the failure. Same build as before, `main` at `2ea0d3c` without this PR. The capture covers the root hub that carries the Pico and two of the three Megas, so a board that works and the board that fails are in the same file.

The result contradicts what I wrote here earlier, so let me correct that first.

**The command reaches the board. The board does not answer.**

Pico, device address 10, `VID_2E8A&PID_000A`:

```
10:18:07.501126  ep 0x02 OUT    2 bytes: 39 3b  ("9;")
10:18:07.501177  ep 0x02        completed, status success
10:18:10.214398  ep 0x82 IN     0 bytes, status C0010000 (USBD_STATUS_CANCELED)
```

The bulk IN read was posted at 10:18:06.221 and was still empty when the Connector gave up and cancelled it 2.713 s later. Nothing came back on the data endpoint in between.

An Arduino Mega on the same root hub, in the same capture:

```
10:18:08.241169  ep 0x02 OUT    2 bytes: 39 3b  ("9;")
10:18:08.244384  ep 0x82 IN    32 bytes: "10,Gagagu FCU/EFIS Mega,Mobiflig..."
```

3.2 ms.

**The correction.** I wrote earlier in this thread that "the first command is lost, not slow". That is wrong. The command is not lost: it is delivered, the OUT transfer completes successfully, and my board stays silent for the whole timeout. @elral's read of this was right — the host side timing is not what is broken here.

So the root cause is in my own firmware and my issue #3321 pointed at the wrong component. I am carrying on from the firmware side and will report back.

One thing from that report still stands without any board of mine involved: `GetInfo()` reads its fields from a second command whose result is never checked, so a timed out follow-up read produces an entry with a board type but no version and no serial. That is the line in #3220's log, on a bare Arduino Mega. The `secondInfoCommand.Ok` guard in this PR covers it.

Whether the retry itself is worth keeping is your call, and it should be judged on its own merits rather than on my earlier claim that my board was innocent. What it does do is let a board that misses one reply be recognised instead of being listed as having no firmware.

The capture is 25 KB, happy to attach it.

chegevarko1982:
Follow-up to the capture above, and then I am stopping here — the retry in this PR solves my problem, and the cause is below the layer I can usefully dig into. Posting what I measured in case it saves someone time.

**The lost command never reaches CmdMessenger at all.**

`OnUnknownCommand` is attached as the default callback and it answers:

```cpp
void OnUnknownCommand() {
    lastCommand = millis();
    cmdMessenger.sendCmd(kStatus, F("n/a"));
}
```

So any message that reaches `handleMessage()` produces a reply — the info reply, or `n/a`. The capture shows zero bytes coming back. Nothing arrives at the dispatcher, so the bytes are lost below the MobiFlight layer, between the USB hardware acknowledging the OUT transfer and `tud_cdc_available()` reporting them.

That rules out three things I had suspected:

- **Leftover bytes in the parse buffer.** `commandBuffer` in `processLine()` survives a port close, so a partial command with no terminator would merge with the next `9;` into one unknown command. Plausible, but it would answer `n/a`. It does not.
- **`pauseProcessing`.** It gates `feedinSerialData()` entirely, so the second command would fail too. It is answered in 11–15 ms.
- **`boardReady`.** It only gates input polling (`Button`, `Encoder`, `Analog`, `InputShifter`, `DigInMux`), not command reception.

`SerialUSB::available()` is a thin wrapper (`tud_task(); return tud_cdc_available();`), and `_running` is set once in `setup()` and never cleared, since `SerialUSB::end()` is an empty TODO. So the gap is inside the CDC receive path.

**How it reproduces, for whoever picks this up:**

- Only on the *first* contact after the board has gone a long time completely untouched. Both of my reproductions were that, and all four cold boots were that.
- Rapid open / send `9;` / close cycles never fail: **0 losses in 120 trials**.
- The settle time before the first command makes no difference: **0/40 each at 250 ms, 1250 ms and 3000 ms**. That is why raising `ConnectionDelay` from 1250 to 4000 never helped.
- The command is lost for good, not answered late. I kept listening for 10 s past the timeout: nothing. The next command on the same open is answered in 11–15 ms.
- Gaps of 30 and 45 min after a successful exchange did not reproduce it, so it is not a simple timeout. Two failures is too small a sample to call it a threshold or a probability, and getting more costs an hour per data point, which is where I stopped.

One unrelated thing I noticed while reading, in case it is useful: `CustomDevice::PowerSave()` calls `customDevice[i].set()` directly, without the `USE_2ND_CORE` guard that `OnSet()` a few lines below has. On a Pico with `USE_2ND_CORE` that runs the custom device's `set()` on core 0 while core 1 may be inside its `update()`.

Happy to run anything else on this hardware if it helps — it is the one part I can still contribute.

thaildhe172591:
Summarising the capture, since @chegevarko1982 has stepped back and it is the kind of detail that gets lost in a long thread: the GetInfo bytes leave the host and nothing at all comes back from the board. That rules out the dispatcher, because `OnUnknownCommand` is the default callback and answers every message that reaches `handleMessage()`, so an unrecognised command would still produce `n/a`. Silence means the bytes never arrived. The loss is below the MobiFlight layer.

It does not identify the cause, but it does say the retry here is covering a real symptom rather than a protocol bug in the Connector, and it matches @elral's point that the Pico is not busy resetting on connect.

@ColinM9991 the loop is in as you asked. The one question still open from my last comment: three attempts means a port that matches a board by VID/PID but has no firmware waits 7.5 s before it shows as Compatible, against 2.5 s before this PR. I am happy to drop it to two if that startup cost matters more than the third try.

ColinM9991:
> @ColinM9991 the loop is in as you asked. The one question still open from my last comment: three attempts means a port that matches a board by VID/PID but has no firmware waits 7.5 s before it shows as Compatible, against 2.5 s before this PR. I am happy to drop it to two if that startup cost matters more than the third try.

@thaildhe172591 does the delay come from the retry logic when there's no firmware? Is that because `.Ok` returns false?
Since the retry is implemented as a foreach loop, lines 925-936 no longer seem relevant to me because the comments describe a timeout scenario that has been addressed by the loop. This means that we'll actually retry N+1 times here.

For example, if our retry limit is 3 and the device connects on the 3rd retry. There will be a 4th retry on line 929

DocMoebiuz:
@thaildhe172591 and @chegevarko1982 - thank you so much for the detailed investigation. It is great that we found a way to make it work - yet, it feels like we're just dealing with the symptoms but not really with the root cause.

This week, we will have the MobiFlight Meetup. This will be an opportunity for Ralf and me, to sit down and look into this together. We will report back with our findings.

DocMoebiuz:
@thaildhe172591 i would definitely make the retry timeout adjustable so that we can use smaller values when testing. Maybe we could even solve it differently so that tests don't have to wait for multiple seconds to complete.

src/MobiFlightConnector/MobiFlight/MobiFlightModule.cs
```diff
@@ -907,15 +908,32 @@ public IModuleInfo GetInfo()
...
+                Log.Instance.log($"GetInfo timed out on {_comPort}, attempt {attempt} of {GetInfoMaxAttempts}.", LogSeverity.Debug);
+            }
 
             if (InfoCommand.Ok)
             {
                 // Workaround
-                // the following two lines shall get removed
+                // the following read shall get removed
                 // but at the moment something with the timing during startup is wrong.
                 command = new SendCommand((int)MobiFlightModule.Command.GetInfo, (int)MobiFlightModule.Command.Info, CommandTimeout);
```
DocMoebiuz, line 928: Why are we retrying again a second time (fourth time)?

DocMoebiuz (changes requested):

DocMoebiuz:
I reviewed the PR and I am not sure what our findings are. The old code already had a 2 times retry logic because we noticed that this might be necessary in some cases.

This PR suggests to try the Info Command up to three times, and then continues with the existing logic of trying it a second time.

chegevarko1982:
@DocMoebiuz one clarification, since it changes what this PR is fixing.

`GetInfo()` had no retry before this PR. The second send sat inside `if (InfoCommand.Ok)`, so it only ran when the first command had already succeeded — an unconditional re-read after success, not a retry after failure. When the first command timed out, that whole block was skipped and the board was dropped as having no firmware. That is the case I hit.

The retry with the "Sometimes first attempt times out" comment does exist, but in the `Config` getter for `GetConfig`, alongside `ExtraConnectionRetry` for boards that need further attempts. `GetInfo()` never had it.

I ran into this the hard way: my first attempt at a fix changed only that second command, and it did nothing, because the instrumented log showed the *first* command failing.

On the open question of two attempts versus three: in every run of mine one retry was enough — `retry 1 result: Ok=True`, every time. So two attempts covers everything I observed, if the extra 2.5 s on a port without firmware is worth avoiding.

Hope this helps with the investigation.

DocMoebiuz:
@thaildhe172591 @chegevarko1982 - https://github.com/MobiFlight/MobiFlight-Connector/pull/3355 please try this build. We added more logging to better understand what's going on and hopefully where it is failing.

Please report your findings as comments at #3355

DocMoebiuz:
@thaildhe172591 are you on discord? Please join us and we can have a more interactive conversation there.

src/MobiFlightConnector/MobiFlight/MobiFlightModule.cs
thaildhe172591: You are right, and it is gone in f1f95d5.

To answer the question directly: that second send was not a retry. It sat inside `if (InfoCommand.Ok)`, so it only ran once the command had already been answered, and the fields were then read from its result. With the loop in front of it, an answered board was asked once more for no reason, and a board that answered on attempt three was asked a fourth time, which is what you and @ColinM9991 both saw.

I have removed it rather than guarded it. Its comment says it is working around startup timing, and that is what the loop now handles. It came in with 1f0fd45 for #546, where a first read could return parts of the config; that issue was closed with your note that it no longer happened above firmware 1.12.2, and current firmware is 3.1.4. Removing it also closes the other way a board could end up as `Compatible`: when only that second read timed out, the fields came back empty, which is the `Arduino Mega 2560, Arduino Mega 2560, /, .` line in #3220.

If you would rather keep it, I will put it back behind the same `Ok` check.

src/MobiFlightConnector/MobiFlight/MobiFlightModule.cs
thaildhe172591: Both confirmed, and the N+1 is fixed in f1f95d5 by removing those lines rather than reworking them, since as you say the timeout scenario their comment describes is now the loop's job.

On the 7.5 s: yes, it comes from the loop, and yes, because `.Ok` is false. `ReceivedCommand.Ok` is `CmdId >= 0`, and a command that times out returns a default `ReceivedCommand` with `CmdId` of -1, so a port with no firmware looks exactly like a board that did not answer in time. There is no way to tell those apart at this level, which is also what #3269 is about. The loop therefore spends its full three attempts on any such port.

The limit is a field now, so it is one line to make it two attempts. Say which you prefer and I will set it.

thaildhe172591:
@DocMoebiuz I will join the Discord, thanks for the invitation. I expect to still do most of the detailed back and forth here, mainly so that the reasoning stays attached to the diff for whoever reads this later, but I will be reachable there.

Pushed f1f95d5, rebased onto current main. Two changes, both from your review and @ColinM9991's:

The unconditional second read is gone, which removes the N+1 you both asked about. Details are in the two review threads.

`CommandTimeout` and `GetInfoMaxAttempts` are no longer `const`, so a test can set a short timeout on the module instead of waiting for the real one. The three GetInfo tests went from 13 s to 873 ms. Nothing outside `MobiFlightModule` referenced either constant. If you would rather have this as a constructor argument or on the board definition, that is easy to change; a plain field was the smallest thing that made the tests fast.

On the root cause, I agree this stays a symptom fix and I would not argue otherwise. What the capture in this thread does rule out is the MobiFlight protocol layer: `OnUnknownCommand` answers `n/a` to anything that reaches `handleMessage()`, and the board sends nothing at all, so the bytes are lost below it. Whatever you and Ralf find at the meetup, a `GetInfo()` that gives up after one unanswered command looks worth fixing on its own, and this is small enough to sit alongside a proper fix rather than block it.

Happy to close this in favour of #3355 or a follow-up of yours if you would rather carry it there.

DocMoebiuz (approved):

DocMoebiuz:
Let's give this a try... i am wondering why we had the second attempt in the code even if InfoCommand.OK was true. But we will see what the users' feedback is going to be.

chegevarko1982:
Tried build 0.0.3340.3672 (`f1f95d5`) on the hardware from #3321. Cold boot, Debug level, same four boards. The Pico is identified on the first launch:

```
18:58:27.707 Connected to Raspberry Pico at COM10 of type MobiFlight RaspiPico (DTR=>True).
18:58:31.498 GetInfo(): GetInfo timed out on COM10, attempt 1 of 3.
18:58:31.515 GetInfo(): Retrieved board: MobiFlight RaspiPico, PICO RADIO, 1.1.2/3.1.4, SN-50436379484F6E1C.
18:58:31.525 RegisterModule(): Registering module PICO RADIO:COM10.
```

The first attempt still goes unanswered, as expected — the underlying cause is untouched. The second attempt is answered 17 ms after the first one times out.

On why the second read was there: with it removed, every field comes from a single reply and all of them are complete — type, name, firmware and core version, full serial. The three Megas behave the same way: identified about 15 ms after their connection delay, one command each, no retries logged. On this hardware nothing is lost without that second read.

DocMoebiuz:
@chegevarko1982 - this PR has been merged already and is been released with Beta 11.2.0.7

Issue #3271, opened 2026-08-13
add unit test for properly resolving a missing feed.json from non-core translation


tigert:
**Is your feature request related to a problem? Please describe.**
When removing feed.json from Finnish (non-core) just realized we should check that there is a unit test that makes sure the fallback works properly.

**Describe the solution you'd like**
Add test if it doesnt exist yet

DocMoebiuz:
included in #3283

PR #3341, opened 2026-09-03
use english news feed for non-core languages, and use english fallback if online feed cannot be loaded.

### LINKED ISSUE #3283: use english news feed for non-core languages, and use english fallback if online feed cannot be loaded.
Core languages have the fallback json translated, but this is not done for non-core languages, because in most cases the online feed should be loaded and the static file is just a fallback if the feed cannot be downloaded for some reason.

However the non-core languages always fall back to the static feed and do not even try to load the online feed. While it would be best to show a localised feed, users of non-core languages are missing out on the news and announcements.

Try in this order:
- try loading localised online feed (might exist, but likely not for non-core languages)
- if that does not work, try loading english online feed (should always exist unless it cannot be accessed)
- if that does not work, try loading localised static feed (might exist but not for non-core languages)
- if that does not work (non core language in use) load the english static feed that always exists.

Use [Azure Translator](https://azure.microsoft.com/en-gb/pricing/details/translator/) on server side and cache the result as an option for non-core languages, label it as "automatically translated" and have link to "show original" so the user can switch to english if azure does a bad job. The service seems free for moderate amounts, so maybe it could be used on server side, so not all users will request a translation of the same news post.

**Additional context**

JunMyungH:
### Summary

It was necessary to improve the community feed fallback behavior for non-core languages, because previously it relied on the bundled static fallback feed.

The app now tries to load the localized online feed first and if the feed is not available, it tries the English online feed next.

This helps users of non-core languages still receive current MobiFlight news and announcements, even when no localized online feed exists.

### Acceptance criteria

- [x] Localized online feed is tried first
- [x] English online feed is used if the localized online feed cannot be loaded
- [x] Existing static feed fallback behavior still works
- [x] Non-core languages without their own `feed.json` still show community feed content

### DoD Checklist

- [x] Frontend tests
     
## Related issues

fixes #3283 
relates #3271

### Out-of-scope

Server-side automatic translation of feed posts is not included in this PR.

This PR also does not add new static `feed.json` files for non-core languages.

### Notes

The intended fallback order:

1. localized online feed
2. English online feed
3. localized static feed
4. English static feed

### DIFF
--- a/src/MobiFlightConnector/frontend/src/components/community/CommunityMainCard.tsx
+++ b/src/MobiFlightConnector/frontend/src/components/community/CommunityMainCard.tsx
@@ -11,7 +11,6 @@ import {
 } from "@/components/ui/card"
 import { ScrollArea } from "@/components/ui/scroll-area"
 import { fetchRemoteCommunityFeedWithFallback } from "@/lib/feed"
-// import { fetchRemoteCommunityFeed } from "@/lib/feed"
 import { useErrorFallbackTest } from "@/lib/hooks/useErrorFallbackTest"
 import { CommunityPost } from "@/types/feed"
 import { useQuery } from "@tanstack/react-query"
--- a/src/MobiFlightConnector/frontend/src/components/community/CommunityMainCard.tsx
+++ b/src/MobiFlightConnector/frontend/src/components/community/CommunityMainCard.tsx
@@ -39,8 +39,7 @@ const CommunityMainCard = () => {
     import.meta.env.VITE_FEED_REMOTE_BASE_URL ?? remoteFeedDefaultBaseUrl
   ).trim()
 
-  const language =
-    i18n.languages?.[0] || i18n.language || i18n.resolvedLanguage || "en"
+  const language = i18n.language || i18n.resolvedLanguage || "en"
 
   const remoteFeedQuery = useQuery({
     queryKey: ["community-feed", language, remoteFeedBaseUrl],
--- a/src/MobiFlightConnector/frontend/src/components/community/CommunityMainCard.tsx
+++ b/src/MobiFlightConnector/frontend/src/components/community/CommunityMainCard.tsx
@@ -10,7 +10,8 @@ import {
   CardTitle,
 } from "@/components/ui/card"
 import { ScrollArea } from "@/components/ui/scroll-area"
-import { fetchRemoteCommunityFeed } from "@/lib/feed"
+import { fetchRemoteCommunityFeedWithFallback } from "@/lib/feed"
+// import { fetchRemoteCommunityFeed } from "@/lib/feed"
 import { useErrorFallbackTest } from "@/lib/hooks/useErrorFallbackTest"
 import { CommunityPost } from "@/types/feed"
 import { useQuery } from "@tanstack/react-query"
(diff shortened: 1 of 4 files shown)


src/MobiFlightConnector/frontend/src/components/community/CommunityMainCard.tsx
```diff
@@ -31,25 +32,27 @@ const CommunityMainCard = () => {
...
   // If not set, the frontend will default to "https://mobiflight.com/feed" as the base url for the feed
   const remoteFeedDefaultBaseUrl = "https://mobiflight.com/feed"
-  const remoteFeedBaseUrl = (import.meta.env.VITE_FEED_REMOTE_BASE_URL ?? remoteFeedDefaultBaseUrl).trim()
+  const remoteFeedBaseUrl = (
+    import.meta.env.VITE_FEED_REMOTE_BASE_URL ?? remoteFeedDefaultBaseUrl
+  ).trim()
 
-  const language = i18n.resolvedLanguage || i18n.language || "en"
+  const language =
+    i18n.languages?.[0] || i18n.language || i18n.resolvedLanguage || "en"
```
DocMoebiuz: This requires a comment. Why has the order changed and why did we add the additional ` i18n.languages?.[0]`

src/MobiFlightConnector/frontend/src/components/community/CommunityMainCard.tsx
```diff
@@ -10,7 +10,8 @@ import {
   CardTitle,
 } from "@/components/ui/card"
 import { ScrollArea } from "@/components/ui/scroll-area"
-import { fetchRemoteCommunityFeed } from "@/lib/feed"
+import { fetchRemoteCommunityFeedWithFallback } from "@/lib/feed"
+// import { fetchRemoteCommunityFeed } from "@/lib/feed"
```
DocMoebiuz: remove commented imports

src/MobiFlightConnector/frontend/src/lib/feed.ts
```diff
@@ -73,6 +73,37 @@ const makeUrlAbsolute = (
...
+  baseUrl,
+  language,
+  timeoutMs = 3000,
+}: FetchRemoteFeedOptions): Promise<CommunityPost[]> => {
+  const primaryLanguage = resolveLanguage(language)
+  const fallbackLanguage = "en"
+
+  const languagesToTry = Array.from(
+    new Set([primaryLanguage, fallbackLanguage]),
+  )
```
DocMoebiuz: As pattern we have used the following syntax in other places (e.g. filter lists)
```
const languagesToTry = [...new Set([
  primaryLanguage,
  fallbackLanguage,
])]
```

the `spread`-operator makes it look a little nicer to ready - at least in my opinion.

DocMoebiuz (changes requested):

src/MobiFlightConnector/frontend/src/lib/feed.ts
DocMoebiuz: @JunMyungH do you agree?

JunMyungH: yup, I will change it as you suggested

DocMoebiuz (approved):

PR #3342, opened 2026-09-03
Migrate Settings dialog from WinForms to React UI

### LINKED ISSUE #3326: Migrate Settings Dialog to new UI framework
We want to migrate the old settings dialog over to the new UI framework.

* As a user I can access the settings dialog via the main menu "Extras" > "Settings"
* As a user I can access the controller settings dialog (legacy) via the main menu "Extras" > "Manage controllers"
* As a user I can manage the following settings through the new dialog:
  * All settings that are available on "General" tab
  * All settings from ProSim tab
* Use shadcnUI components
* Main groups (maybe as cards): 
  * General - All MobiFlight settings
  * Simulator - ProSim and other simulator specific settings
* Two step editing, like other dialogs
  * All settings are changed temporary
  * Applied only, if user applies changes
  * If user cancels all changes will be reverted 
* All settings are persisted by the .NET app, using the existing user.config mechanism

**Out-of-scope**
* Saving settings via cloud api

Old layout:
Image

OK434:
### Summary
Migrated the Settings dialog from the legacy WinForms UI to the modern React UI framework using `shadcn/ui` components. Settings are organized into "General" and "Simulator" tabs and feature two-step editing where changes are held in a draft state and applied only upon clicking Save (or reverted on Cancel with a discard confirmation dialog). All settings continue to be persisted by the .NET backend using the existing `user.config` mechanism. Legacy controller settings remain accessible through a dedicated menu item.

### Screenshots / recordings

https://github.com/user-attachments/assets/5731b1b1-ca74-402c-9493-902691020d99

### Acceptance criteria
- [x] As a user I can access the settings dialog via the main menu "Extras" > "Settings"
- [x] As a user I can access the controller settings dialog (legacy) via the main menu "Extras" > "Manage controllers"
- [x] All settings from the legacy "General" tab migrated to the new React UI
- [x] All settings from the legacy "ProSim" tab migrated to the new "Simulator" tab
- [x] Settings UI built with modern `shadcn/ui` components (Cards, Tabs, Inputs, Checkboxes, ComboBoxes, Sliders)
- [x] Two-step editing implemented (temporary draft state, applied on Save, reverted on Cancel)
- [x] Confirmation dialog appears when attempting to exit with unsaved changes
- [x] All settings persisted by .NET backend via `CommandUpdateSettingsHandler` and `Properties.Settings.Default.Save()`
- [x] Legacy General and ProSim tabs removed from the legacy WinForms `SettingsDialog`

### DoD Checklist
- [x] Frontend tests
  - [x] Modal open & close tests
  - [x] Save & publish `CommandUpdateSettings` test
  - [x] Cancel without command test
  - [x] Tab switching test
  - [x] Discard changes confirmation test
  - [x] core (en, de, es)
  - [x] additional langs (fi, pt, ru)

## Related issues
fixes #3326 

### Out-of-scope
- Saving settings via cloud API


DocMoebiuz:
@OK434 - Good job! It already looks quite good but I would like to make some adjustments.

image

General - is OK.
Simulator - should be `Simulators`

image
I don't think repeating "General" is necessary. If we want to group the settings into sections based on a category, then we might find other labels. I would leave it out of scope for now though.

For some reason i am not convinced of the top-level card. Have you tried it without? How would that look?

On the simulators-tab, ProSim / HomeSim should be the label instead of "Simulator".

DocMoebiuz:
```
### Application
Language                                             [Drop Down]
Beta Updates                                            [switch]
Community Feedback Program                              [switch]

### Logging
Show Log Panel                                          [switch]
Log Level                                            [Drop Down]
Log Joystick Axis                                       [switch]

### Startup and Run options
Check HubHop presets                                    [switch]
Auto Retrigger                                          [switch]
Minimize on AutoRun                                     [switch]
Config Execution Speed                                  [slider]
Test Mode Speed                                         [slider]

File: src/MobiFlightConnector/frontend/src/components/settings/components/GeneralSettingsCard.tsx [hunk 1]
```diff
@@ -0,0 +1,288 @@
...
+    { value: "de-DE", label: "Deutsch" },
+    { value: "es-ES", label: "Español" },
+    { value: "fi-FI", label: "Suomi" },
+    { value: "pt-PT", label: "Português" },
+    { value: "ru-RU", label: "Русский" },
+  ]
+
+  return (
+    <Card className="w-full">
+      <CardContent className="space-y-6 p-6">
```
DocMoebiuz, line 45: So far in the code base, we are not using space-* - we always use flex-row/col and gap-*.
I really think it is better to have one approach across the app.

src/MobiFlightConnector/frontend/src/components/settings/components/GeneralSettingsCard.tsx
```diff
@@ -0,0 +1,288 @@
...
+        {/* Application */}
+        <div className="space-y-3">
+          <h3 className="text-base font-bold">
+            {t("Settings.General.Application.Title")}
+          </h3>
+
+          <div className="space-y-1">
+            <div className="flex items-center justify-between gap-4 -mx-2 rounded-md p-2 transition-colors hover:bg-muted/70">
```
DocMoebiuz, line 53: This is very repetitive - it is going to be painful if you want to make a change.
Consider creating a small reusable (wrapper-)component for a settings item. I think that will help and is DRY. 

If you look at the similarities of each settings row - I think you can come up with a nice reusable component.

src/MobiFlightConnector/MobiFlight/BrowserMessages/Incoming/CommandUpdateSettings.cs
```diff
@@ -0,0 +1,58 @@
...
+    public class CommandUpdateSettings
+    {
+        // General Settings
+        [JsonProperty]
+        public int RecentFilesMaxCount { get; set; }
+
+        [JsonProperty]
+        public bool LogEnabled { get; set; }
+
+        [JsonProperty]
```
DocMoebiuz, line 14: Are you re-creating a Base/Settings-class? We should not need this.

src/MobiFlightConnector/UI/Dialogs/SettingsDialog.cs
```diff
@@ -76,9 +79,6 @@ private void Init()
         /// </summary>
         private void loadSettings ()
         {            
-            // TAB General            
-            generalPanel.loadSettings();
-            
             // TAB Arcaze
```
DocMoebiuz, line 82: The general panel can now be deleted from the project.
Same for the prosim panel

DocMoebiuz (changes requested):
Found some things to improve!

src/MobiFlightConnector/frontend/src/types/commands.d.ts
```diff
@@ -235,3 +239,8 @@ export interface CommandRefreshPresets extends CommandMessageBase {
     type: "prosim" | "vjoy"
   }
 }
+
+export interface CommandUpdateSettings extends CommandMessageBase {
+  key: "CommandUpdateSettings"
+  payload: Partial<Settings>
+}
```
DocMoebiuz, line 248:
```suggestion
export interface CommandUpdateSettings extends CommandMessageBase {
  key: "CommandUpdateSettings"
  payload: {
    Settings: Partial<Settings>
  }
}
```

src/MobiFlightConnector/MobiFlight/BrowserMessages/Incoming/Handler/CommandUpdateSettingsHandler.cs
```diff
@@ -0,0 +1,59 @@
...
+                {
+                    Log.Instance.Severity = LogSeverity.Info;
+                }
+            }
+
+            // Save to user.config
+            Properties.Settings.Default.Save();
+
+            Log.Instance.log("Settings updated and saved successfully.", LogSeverity.Info);
+        }
```
DocMoebiuz, line 70: Please take a look at our Base/Settings.cs class - it is a good candidate to do the reverse update.
This would also allow for unit testing the logic. Right now this class doesn't have a unit test class.

src/MobiFlightConnector/MobiFlight/BrowserMessages/Incoming/CommandUpdateSettings.cs
```diff
@@ -0,0 +1,58 @@
...
+        public int ProSimPort { get; set; }
+
+        [JsonProperty]
+        public bool ProSimAutoConnectEnabled { get; set; }
+
+        [JsonProperty]
+        public int ProSimMaxRetryAttempts { get; set; }
+    }
+}
```
DocMoebiuz, line 11: All these properties exist in our Settings class... you don't have to redefine them here.
Use `Payload` or `Settings` as property with type of `Settings` and then it will work just out-of-the box in terms of properties.

src/MobiFlightConnector/UI/Dialogs/SettingsDialog.cs
```diff
@@ -111,15 +107,6 @@ private void saveSettings()
...
 
-            // ProSim Tab
-            proSimPanel.SaveSettings();
-            
-            // Reset ProSim connection state when settings change
-            if (execManager != null)
-            {
-                execManager.ResetProSimConnectionState();
-            }
-
```
DocMoebiuz, line 122: I did not check what is happening inside this method, but you should definitely double-check and maybe do this in the `CommandUpdateSettingsHandler.CommandUpdateSettingsHandler` method

DocMoebiuz (changes requested):

src/MobiFlightConnector/frontend/src/components/settings/components/GeneralSettingsCard.tsx
```diff
@@ -0,0 +1,257 @@
...
+
+  const logEnabled = values.LogEnabled ?? true
+  const logLevel = values.LogLevel ?? "info"
+  const language = values.Language ?? ""
+
+  const logOptions: { value: LogLevel; label: string }[] = [
+    { value: "debug", label: "Debug" },
+    { value: "info", label: "Info" },
+    { value: "warn", label: "Warn" },
+    { value: "error", label: "Error" },
```
DocMoebiuz, line 57: You should use the type LogLevel and it defined values as keys.

DocMoebiuz:
image

Config execution speed and test speed values are both not saved / restored correctly.

DocMoebiuz (changes requested):
The last change would be to move the settings dialog specific tests into its own spec-file

DocMoebiuz (approved):
Woohoo! It took some iterations but now it is a great first new version!

PR #3343, opened 2026-09-03
Log shows only one space as data even if multiple are send

### LINKED ISSUE #3195: Log shows only one space as data even if multiple are send
**Describe the bug**
The log entries for 7 segment displays shows only one space character as data even if multiple spaces are send. Most likely this happens also for other devices like LCD displays.

**Steps To Reproduce**
1. activate log
2. send a value to 7 segement display
3. clear the display
4. the log shows only one space for the data send

**Expected behavior**
For the 7 segment displays the log should show 8 spaces for clearing the display

**Actual behavior**
see above

**Details**
 - MobiFlight Version: 11.x.y
 - FlightSim: all
 - AddOn: all

**Additional context**
n/a

JunMyungH:
### Summary

Previously, multiple spaces in a log message were collapsed by normal HTML whitespace rendering. This made log entries for cleared displays appear as if only one space had been sent, even when the actual log message contained multiple spaces.
The log panel now preserves consecutive spaces when rendering log messages, while keeping the existing truncation behavior for long log lines.

### Screenshots / recordings

Preseved_whitespace

### Acceptance criteria

- [x] Consecutive spaces in log messages are preserved visually

### DoD Checklist

- [x] Frontend tests
     
## Related issues

fixes #3195 

### Out-of-scope

This PR does not change how log messages are generated by the backend.
It also does not change the actual log data. The fix only affects how existing log messages are rendered in the frontend log panel.

### Notes

The fix uses CSS whitespace handling instead of modifying the log message text itself. This avoids replacing spaces with special characters and keeps the original log message unchanged.

### DIFF
--- a/src/MobiFlightConnector/frontend/src/components/LogPanel.tsx
+++ b/src/MobiFlightConnector/frontend/src/components/LogPanel.tsx
@@ -211,7 +211,7 @@ const LogPanel = () => {
                 </div>
                 <div
                   data-testid="log-entry-message"
-                  className="min-w-0 overflow-hidden text-ellipsis whitespace-pre"
+                  className="truncate whitespace-pre"
                 >
                   {entry.Message}
                 </div>
--- a/src/MobiFlightConnector/frontend/src/components/LogPanel.tsx
+++ b/src/MobiFlightConnector/frontend/src/components/LogPanel.tsx
@@ -209,7 +209,12 @@ const LogPanel = () => {
                 >
                   {entry.Severity}
                 </div>
-                <div className="truncate">{entry.Message}</div>
+                <div
+                  data-testid="log-entry-message"
+                  className="min-w-0 overflow-hidden text-ellipsis whitespace-pre"
+                >
+                  {entry.Message}
+                </div>
               </div>
             ))
           )}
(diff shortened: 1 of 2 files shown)


src/MobiFlightConnector/frontend/src/components/LogPanel.tsx
```diff
@@ -209,7 +209,12 @@ const LogPanel = () => {
                 >
                   {entry.Severity}
                 </div>
-                <div className="truncate">{entry.Message}</div>
+                <div
+                  data-testid="log-entry-message"
+                  className="min-w-0 overflow-hidden text-ellipsis whitespace-pre"
```
DocMoebiuz: `truncate` is a tailwind utility that does `overflow-hidden text-ellipsis` already.
You can just add `whitespace-pre`

why do we need `min-w-0`?

DocMoebiuz (changes requested):

src/MobiFlightConnector/frontend/src/components/LogPanel.tsx
JunMyungH: I just worried if the texts would broken, so I wanted to let it shrink. Then I will use only truncate and whitespace-pre

DocMoebiuz (approved):
Thanks!

PR #3344, opened 2026-09-03
Add WinCtrl CDU support for X-Plane UNS-1.

maboehme:
The UNS-1 is used in multiple aircraft. I've added two to ScriptMappings.json that I have and am able to test with: The FlyJSim 737-200 and FlyJSim Q4XP.

The UNS-1 supports various "canned" graphics that can be displayed on the holding page, controlled by the `hold_graphics` dataref. As the WinCtrl CDU only supports text, I use ASCII art to draw the holding graphics.

Here's an example of what the holding page looks like on the in-sim UNS-1:

uns1_holding

And here's what that looks like on the WinCtrl CDU:

uns1_holding_winctrl

It's not exactly beautiful, but it's workable.

### DIFF
--- a/src/MobiFlightConnector/Scripts/ScriptMappings.json
+++ b/src/MobiFlightConnector/Scripts/ScriptMappings.json
@@ -185,6 +185,18 @@
       "ProductIds": [ "WinwingCDUs" ],
       "AircraftMatchPattern": "microsoft_atr",
       "ScriptName": "ms_atr_42_72_winwing_cdu.py"
+    },
+    {
+      "VendorId": "0x4098",
+      "ProductIds": [ "WinwingCDUs" ],
+      "AircraftMatchPattern": "fjs 732",
+      "ScriptName": "xplane_uns_1.py"
+    },
+    {
+      "VendorId": "0x4098",
+      "ProductIds": [ "WinwingCDUs" ],
+      "AircraftMatchPattern": "flyjsim q4xp",
+      "ScriptName": "xplane_uns_1.py"
     }
   ]
 }
--- a/src/MobiFlightConnector/Scripts/Winwing/xplane_uns_1.py
+++ b/src/MobiFlightConnector/Scripts/Winwing/xplane_uns_1.py
@@ -1,0 +1,506 @@
+"""
+Adds support for the X-Plane UNS-1.
+
+Many X-Plane aircraft have similar formats for datarefs and the means of retrieving, translating and sending updates is mostly the same.
+
+In order to support multiple CDU devices seamlessly, a dynamic approach is taken whereby an enum class is defined that contains the supported devices.
+A device is considered "supported" if it exists in the aircraft. Some aircraft have 3 CDUs while others have 2.
+Each enum member is assigned a value that is used to construct the X-Plane dataref identifier. Example: "cdu1" in "uns1/cdu1/text_line_0".
+
+Upon script start, MobiFlight is probed (get_available_devices()) to detect the devices connected to the PC. Any device that returns a successful response is then tracked.
+
+Two tasks are started independently for each available CDU device.
+1. handle_dataref_updates -> Listens to X-Plane's WebSocket server for dataref updates for that specific CDU and pushes an event to a queue
+2. handle_device_update   -> Listens to the queue and dispatches updates to MobiFlight to update that CDU
+
+Tasks are started independently for each CDU device to ensure each device can update quickly, particularly when players might be performing shared cockpit flights.
+
+Upon a failed connection while dispatching updates to MobiFlight, the handle_device_update function uses `async for` with the websockets client. The failed message is put back in the queue, the loop continues to the next iteration which then reconnects again.
+The failed message is picked back up and dispatched to MobiFlight. This ensures a user's device eventually receives the updated display contents and doesn't hang which would require the user to cycle the page again.
+"""
+
+import asyncio
+import base64
+import json
+import logging
+import os
+import urllib.request
+from dataclasses import dataclass, KW_ONLY
+from enum import StrEnum
+
+import websockets
+
+CDU_COLUMNS = 24
+CDU_ROWS = 11
+CDU_CELLS = CDU_COLUMNS * CDU_ROWS
+
+WEBSOCKET_HOST = "localhost"
+WEBSOCKET_PORT = 8320
+
+BASE_REST_URL = "http://localhost:8086/api/v2/datarefs"
+BASE_WEBSOCKET_URI = f"ws://{WEBSOCKET_HOST}:8086/api/v2"
+
+WS_CAPTAIN = f"ws://{WEBSOCKET_HOST}:{WEBSOCKET_PORT}/winwing/cdu-captain"
+WS_CO_PILOT = f"ws://{WEBSOCKET_HOST}:{WEBSOCKET_PORT}/winwing/cdu-co-pilot"
+
+# Q4XP
+COLOR_MAP_0 = {
+    0: "w",
+    1: "w",
+    2: "g",
+    3: "y",
+    4: "c",
+    5: "m",
+    6: "g",
+    7: "o",
+    8: "r",
+    9: "e",
+    10: "w",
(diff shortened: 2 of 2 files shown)


Koseng:
@maboehme please check out Claude Review:

Solid, well-structured work that follows the established `xplane_default_fmc.py` pattern closely. CI is green (pylint `F,E` on 3.11–3.13), `ScriptMappings.json` is valid, no `.csproj` change is needed (the project is SDK-style now and globs `Scripts\**`), and no existing `AircraftMatchPattern` collides with `fjs 732` / `flyjsim q4xp`. The `Overlay` abstraction is clean — every overlay row is exactly 24 chars and every `row_offset + len(text)` stays inside `CDU_ROWS`. Two things are outright improvements over the base template: `decode(errors='replace')` and keeping reverse-video spaces.

Findings below, most important first.

---

## 1. Any missing dataref kills the entire script, both CDUs

`generate_display_json` indexes `values[...]` directly for 25 keys (11 text + 11 style + `color_set`, `hold_dir`, `hold_graphics`). A `KeyError` there is raised inside the `try` in `handle_device_update` — which only catches `ConnectionClosed` — so it propagates out of the `async for`, fails the task, and `asyncio.gather` (no `return_exceptions`) re-raises it out of `main()`. The process exits and *both* CDUs freeze on their last frame.

Two ways in, both reproduced against the script:

```
A: aircraft doesn't publish hold_graphics  -> KeyError: 'uns1/cdu1/hold_graphics'
B: first X-Plane update carries a subset   -> KeyError: 'uns1/cdu1/text_line_1'
```

(B) is inherited from the template, but this script raises the stakes: it now requires three *scalar* datarefs that must exist under both prefixes (`uns1/…` on the 732, `FJS/Q4XP/…` on the Q4XP), and `last_known_values` starts empty while `queue.put` fires on the very first `dataref_update_values` message.

`rotate_md11.py:109` already uses the defensive form:

```python
color_set     = values.get(device.get_color_set_dataref(), 0)
hold_dir      = values.get(device.get_hold_dir_dataref(), 0)
hold_graphics = values.get(device.get_hold_graphics_dataref(), 0)
```

…and skipping the frame until `len(new_values) == len(dataref_map)` would close (B) for the text/style rows.

## 2. `text[row][col]` can raise `IndexError` on short lines

`text_line_*` is a byte array, but the UNS-1 writes UTF-8 box-drawing glyphs into it (that is what `CHAR_MAP` exists for) — and those are 3 bytes each, so a fixed-size array decodes to *fewer* characters than bytes. If a decoded line is under 24 chars, `text[row][col]` raises `IndexError` (reproduced), the same fatal path as above. `zip()` in `draw_onto` compounds it by silently truncating the row to the shorter operand.

`laminar_747_400.py:101` guards this with a one-liner:

```python
text = [values[device.get_text_dataref(row)].ljust(CDU_COLUMNS) for row in range(CDU_ROWS)]
```

Your testing suggests the arrays are generously sized, so this is insurance rather than an observed break — but it is cheap insurance on exactly the pages this PR targets.

## 3. Question: is `hold_dir` really 0 off the holding page?

The racetrack is gated *only* on `hold_dir == ±1` — `hold_graphics` merely selects which arrows go on top, and there is no `HOLD_GRAPHICS_NONE`. If `hold_dir` retains its last value while another page is displayed, the racetrack gets superimposed on the flight plan. The overlay is applied over arbitrary underlying text with no page check:

```
 3|R03    ----------       |
 4|R04   /   -   →  \      |
 5|R05   ↑          |      |
```

If the dataref is not self-clearing, this needs a second gate (page id, or `hold_graphics`).

## 4. `is_q4xp()` is called once per device, inside the loop

`get_available_devices` calls it per candidate device, but the answer cannot differ between them — and each call downloads X-Plane's complete dataref catalogue. With `fetch_dataref_mapping` doing the same fetch per device, that is up to 4 full catalogue downloads on the startup path where 1 would do. Hoisting it above the loop fixes it.

(Both are blocking `urllib` calls in async context — inherited from the template, so not something to fix in this PR.)

## 5. Nits

- `text: [str]` / `style: [bytes]` are list *literals*, not annotations — should be `list[str]` / `list[int]` (the style row yields ints, not bytes).
- `get_dataref_prefixes(self) -> str` returns a list — should be `-> list[str]`.
- `COLOR_MAP_0`…`COLOR_MAP_3` + `COLOR_SETS` could just be a list indexed by color set, with `COLOR_SETS[3]` as the fallback.

Koseng:
@neilenns could you please add documentation for FlyJSim 737-200 and FlyJSim Q4XP? Merci

neilenns:
> @neilenns could you please add documentation for FlyJSim 737-200 and FlyJSim Q4XP? Merci

Can you confirm what aircraft? This PR title is for the UNS-1.

maboehme:
> > @neilenns could you please add documentation for FlyJSim 737-200 and FlyJSim Q4XP? Merci
> 
> Can you confirm what aircraft? This PR title is for the UNS-1.

I can confirm that the aircraft to add are the FlyJSim 737-200 and FlyJSim Q4XP.

The UNS-1 is an FMS that is used in multiple aircraft addons; the two I've mentioned are ones that I have been able to test with and that I've added to ScriptMappings.json.

I'll respond to the Claude comments in the coming days.

maboehme:
Here are my responses to the Claude comments.

> Findings below, most important first.
> 
> ## 1. Any missing dataref kills the entire script, both CDUs

I've replaced the `[]` dataref lookups with `.get()`. I've never seen any issues with missing datarefs, but I get it can't hurt to be defensive. 

Just for completeness, some notes on where Claude doesn't understand the details of the code (though this doesn't impact the main point here).

> [...]
> Two ways in, both reproduced against the script:

I have my doubts whether Claude has actually reproduced anything here -- surely it can't run the entire MobiFlight application?

> (B) is inherited from the template, but this script raises the stakes: it now requires three _scalar_ datarefs that must exist under both prefixes (`uns1/…` on the 732, `FJS/Q4XP/…` on the Q4XP), and `last_known_values` starts empty while `queue.put` fires on the very first `dataref_update_values` message.

I'm not sure why Claude is concerned about `last_known_values` starting empty -- it's not as if we ever provide these values to the rest of the code, they're just used to decide whether we need to update the display.

> [...]
> …and skipping the frame until `len(new_values) == len(dataref_map)` would close (B) for the text/style rows.

I've decided instead to do this too simply by using `.get()` instead of `[]`, which seemed simpler.

> ## 2. `text[row][col]` can raise `IndexError` on short lines
> [...]

I've added the `.ljust(CDU_COLUMNS)` as Claude suggests.

> ## 3. Question: is `hold_dir` really 0 off the holding page?
> The racetrack is gated _only_ on `hold_dir == ±1` — `hold_graphics` merely selects which arrows go on top, and there is no `HOLD_GRAPHICS_NONE`.

Yes, this is by design. I contacted Amy Parent, the author of the UNS-1, and she provided me with documentation on the datarefs that confirms that `hold_dir` is zero when the holding graphics should not be displayed.

> If `hold_dir` retains its last value while another page is displayed, the racetrack gets superimposed on the flight plan.

This doesn't happen.

> ## 4. `is_q4xp()` is called once per device, inside the loop

I've hoisted this above the loop, as suggested. (Good call!)

> ## 5. Nits
> * `text: [str]` / `style: [bytes]` are list _literals_, not annotations — should be `list[str]` / `list[int]` (the style row yields ints, not bytes).

Thanks, changed!

> * `get_dataref_prefixes(self) -> str` returns a list — should be `-> list[str]`.

Good catch, fixed.

> * `COLOR_MAP_0`…`COLOR_MAP_3` + `COLOR_SETS` could just be a list indexed by color set, with `COLOR_SETS[3]` as the fallback.

Yes, but then I wouldn't be able to use the `.get()` that I currently use to apply a default if the index is out of bounds and would instead need to explicitly check against the bounds of the array. Not a big deal, but the `.get()` seemed simpler, and I don't think performance is a concern here.

neilenns:
Docs added to https://github.com/neilenns-projects/mobiflight-docs/pull/558

Issue #3188, opened 2026-07-25
Review, improve and migrate existing confirmation dialogs to new UI


DocMoebiuz:

DocMoebiuz:
related to original #1701

DocMoebiuz:
> [!NOTE]
> This list is already outdated and needs review
> some dialogs have been migrated already or are obsolete

### Confirmation dialogs (YesNo / OKCancel)
Base\AutoUpdateChecker.cs — AutoUpdateChecker.CheckForUpdate
- Message: String.Format(i18n._tr("uiMessageNewUpdateAvailablePleaseUpdate"), newVersion)
- Title: i18n._tr("uiMessageNewUpdateAvailable")
- Buttons: MessageBoxButtons.YesNo
- Icon: (none)
- Action: on Yes → start installer (Process.Start(mobiFlightInstaller, "/install " + newVersion)) and exit

UI\MainForm.cs — UpdateAircraft
- Message: i18n._tr("uiMessageConfirmDiscardUnsaved")
- Title: i18n._tr("uiMessageConfirmDiscardUnsavedTitle")
- Buttons: MessageBoxButtons.YesNo
- Usage: confirm before discarding unsaved changes

UI\MainForm.cs — _autoloadConfig (cmdline file not found)
- Message: i18n._tr("uiMessageCmdParamConfigFileDoesNotExist") + "\r" + cmdLineParams.ConfigFile
- Title: i18n._tr("Hint")
- Buttons: MessageBoxButtons.OK
- Icon: MessageBoxIcon.Exclamation (explicitly passed)

UI\MainForm.cs — LoadConfig
- Cases:
- i18n._tr("uiMessageMigrateConfigFileYesNo") — MessageBoxButtons.OKCancel
- i18n._tr("uiMessageMigrateConfigFileV60YesNo") — MessageBoxButtons.YesNo
- Titles: i18n._tr("Hint")
- Follow-ups: user may be asked for SaveFileDialog or migration steps

UI\MainForm.cs — newFileToolStripMenuItem_Click
- Message: i18n._tr("uiMessageConfirmNewConfig")
- Title: i18n._tr("uiMessageConfirmNewConfigTitle")
- Buttons: MessageBoxButtons.OKCancel

UI\MainForm.cs — MainForm_FormClosing
- Message: i18n._tr("uiMessageConfirmDiscardUnsaved")
[...]

PR #3347, opened 2026-09-04
Ask "Are you sure you want to quit?" with unsaved changes

### LINKED ISSUE #3165: Ask "Are you sure you want to quit?" with unsaved changes
If you try to quit MobiFlight with unsaved changes, MobiFlight asks if you want to save your profile. This question has different context that what the user is trying to do. 

If you choose "no" you will quit without saving. This can catch you off guard, as your mind is focused on closing the app, realize you didn't save, and accidentally choosing "no" (to not quit) which is exactly not what you want.

I would like to play safe and make the question explicitly about quit without changing context to saving the profile, so "no" means "no, do not continue", and backing off is the safe choice.

"You have unsaved changes, are you sure you want to quit?"

Eventually we might want to change to a model where changes are saved instantly, and changes can be reverted with undo. But this is is a separate discussion altogether and out of scope here.

JunMyungH:
### Summary

Replacement of the old WinForms shutdown confirmation for unsaved changes with a new frontend confirmation dialog was required.
When the user tries to quit MobiFlight with unsaved project changes, MobiFlight now shows a clearer confirmation dialog focused on the shutdown action. The user can either discard the changes and close MobiFlight right away, or cancel and keep the application open and continue to edit the configuration.

### Screenshots / recordings

image

### Acceptance criteria

- [x] The old WinForms MessageBox is no longer shown for this shutdown flow
- [x] Clicking `Keep editing` keeps MobiFlight open
- [x] Clicking `Discard changes` closes MobiFlight without saving
- [x] Window close button uses the new confirmation flow
- [x] File -> Exit  uses the new confirmation flow
- [x] Pending dialog strings are translated
- [x] The project view and app itself show same dialog, if they have unsaved changes

### DoD Checklist

- [x] Frontend tests
  - [x] core (en,de,es)
  - [x] additional langs
- [x] Backend tests 
     
## Related issues

fixes #3165
relates #3188 

### Out-of-scope

This PR does not add autosave or an undo mechanism. These topics are separate discussions and outside the scope of this issue.

### Notes

The shutdown flow is now handled through a backend/frontend message exchange. The backend cancels the first shutdown request when there are unsaved changes and asks the frontend to show the confirmation dialog. If the user confirms discarding the changes, the frontend sends a shutdown command back to the backend and the application closes.


DocMoebiuz:
@JunMyungH 
image

Please check the translation file for "dialog" we already had (most of) the strings defined there.

JunMyungH:
Well, the previous version of dialog used also the lines from "UnsavedChanges". Then maybe I have to move it to be under the category of dialog and use the lines from Dialog.

DocMoebiuz:
> Well, the previous version of dialog used also the lines from "UnsavedChanges". Then maybe I have to move it to be under the category of dialog and use the lines from Dialog.

yes, that makes sense.

DocMoebiuz:
I just saw that we have a dialog for the settings changes (#3342) that looks like this:

image

I think we could almost use the same one.

src/MobiFlightConnector/frontend/public/locales/en/translation.json
```diff
@@ -817,6 +812,11 @@
           "SelectPresetPlaceholder": "Select preset..."
         }
       }
+    },
+    "UnsavedChanges": {
+      "Title": "Discard changes?",
+      "Description": "You have unsaved changes. Are you sure you want to discard them?",
+      "Cancel": "Keep editing"
```
DocMoebiuz: I would not call this "Cancel" if we use "KeepEditing" - it will make it harder for translating to other languages.

DocMoebiuz:
This PR does not change the general save or discard dialog used when switching projects or navigating away from project settings.

This comment is not entirely true.

JunMyungH:
Deleted that command part, because I changed the saving and discard part of the original dialog after last commit.

src/MobiFlightConnector/UI/MainForm.cs
```diff
@@ -108,6 +111,8 @@ public bool ProjectHasUnsavedChanges
             }
         }
 
+        private bool shutdownConfirmed = false;
+
```
DocMoebiuz: I don't think this is needed because you have two explicit handlers.

src/MobiFlightConnector/frontend/src/components/ConfirmationDialogShutdownHandler.tsx
```diff
@@ -0,0 +1,39 @@
...
+      key: "CommandShutdown",
+      payload: {
+        action: "discardChanges",
+      },
+    })
+  }
+
+  const handleCancel = () => {
+    setOpen(false)
+  }
```
DocMoebiuz, line 27: This logic is a good candidate for a useShutdownConfirmation-hook.
You can bundle the subscription to the app message there and also expose the callbacks.

src/MobiFlightConnector/frontend/src/types/messages.d.ts
```diff
@@ -152,6 +158,8 @@ export interface AuthenticationStatus {
   Authenticated: boolean
 }
 
+export type ShutdownConfirmationRequested = Record<string, never>
```
DocMoebiuz, line 161: Why did you chose this data type of Record<string, never>?

src/MobiFlightConnector/frontend/src/components/ConfirmationDialogShutdownHandler.tsx
DocMoebiuz: Ok. Disregard. You are trying to make this one component that you can use in App.tsx - Ok. That works.

src/MobiFlightConnector/UI/MainForm.cs
DocMoebiuz: And if you ever needed state - then you would want to put that into our Handler class.

DocMoebiuz: Ok... so you are using it as a flag for the first Close request.
In that case, it is better to move that state into your handler class and out of MainForm

DocMoebiuz (changes requested):

src/MobiFlightConnector/UI/MainForm.cs
JunMyungH: Yes, then I will move it to the handler class, CommandShutdownHandler

src/MobiFlightConnector/frontend/src/components/ConfirmationDialogShutdownHandler.tsx
JunMyungH: Yes, as I showed yesterday the jsx part of App.tsx

src/MobiFlightConnector/frontend/src/types/messages.d.ts
JunMyungH: Well, it is because ShutdownConfirmationRequested has no payload fields, but the message API still expected a payload type. I can change it if you prefer another way for empty payloads.

src/MobiFlightConnector/frontend/public/locales/en/translation.json
JunMyungH: Fair enough. Then I will change it as "KeepEditing": "Keep editing"

DocMoebiuz (approved):

PR #3350, opened 2026-09-05
Improved UI Performance via Native websocket messaging

DocMoebiuz:
### Summary
Replaces the WebView2 postMessage bridge between the .NET app and the embedded React frontend with a native, in-process WebSocket server implementation.

Motivation: The postMessage approach was happening on the UI-thread, blocking it and slowing down the overall performance of the app significantly. We first did a spike to verify that the assumption is right: a websocket publisher that runs on its own thread will relieve the stress on the main thread without blocking the app. Resulting in higher throughput at reduced CPU usage.

### Acceptance criteria
- [x] Frontend and backend exchange messages continues to work (drop-in replacement)
  - [x] .NET -> Frontend (App Messages)
  - [x] Frontend -> Backend (Commands)
- [x] Auth WebView flow (login/cancel) continues to work (still uses postMessage)
- [x] All resources are freed on app shutdown 
  - [x] MessageServer/WebSocketServerPublisher disposes cleanly on app shutdown (no leaked sockets/threads)

### DoD Checklist

- [x] Unit tests available
  - [x] .NET backend 
  - [x] Frontend
- [x] Frontend tests
  - ~~core (en,de,es)~~ 
  - ~~additional langs~~
- [x] documentation
  - ~~User docs (docs.mobiflight.com)~~
  - [x] Developer docs (docs/architecture/frontend-backend-messaging.md)

### Out-of-scope
n/a

### Notes
What changed:
- New MessageServer (BCL TcpListener + WebSocket.CreateFromStream, ~40 lines of handshake code) replaces Kestrel/Fleck as options — avoids pulling in Microsoft.AspNetCore.App (~15–25MB) or an unmaintained library.
- WebSocketServerPublisher replaces CompositePublisher/PostMessagePublisher for outbound messages; broadcasts to all connections.
- Origin allow-list on the server as a defense-in-depth measure (loopback binding is the real boundary); no-Origin connections are allowed for future non-browser clients.
- The auth WebView (UserAuthenticationWebView) deliberately stays on postMessage - it spends most of its life on a foreign IdP origin, which the socket's origin check would (correctly) reject. WebViewMessageReceiver gives it a receive-only path into MessageExchange.
- MessageExchange's threading default inverted: Subscribe<T> now runs inline on the delivering thread (no longer assumes WebView2's UI-thread delivery); SubscribeOnUiThread<T> added for handlers that need marshaling, via a captured SynchronizationContext.
- Frontend gains a transport-indirection layer (messageClient / MessageTransport, websocketTransport.ts + postMessageTransport.ts) so components only ever see {key, payload} - this also keeps the Playwright/Vite dev-server E2E path working with no .NET backend.
- New architecture doc: docs/architecture/frontend-backend-messaging.md.
- Follow-up commit disposes MessageServer resources properly.


src/MobiFlightConnector/UI/Panels/FrontendPanel.cs
```diff
@@ -48,16 +48,16 @@ async void InitializeAsync()
...
+            // MainForm's constructor already set the WebSocket publisher on MessageExchange.
+            await InitializeWebView(FrontendWebView, "/start", injectWsUrl: true);
 
-            compositePublisher.AddPublisher("frontend", new PostMessagePublisher(FrontendWebView));
-            compositePublisher.AddPublisher("auth", new PostMessagePublisher(UserAuthenticationWebView));
-
-            MessageExchange.Instance.SetPublisher(compositePublisher);
+            // Auth WebView stays on postMessage - see docs/architecture/frontend-backend-messaging.md.
+            await InitializeWebView(UserAuthenticationWebView, "/", injectWsUrl: false);
+            new WebViewMessageReceiver(UserAuthenticationWebView).MessageReceived
```
ColinM9991, line 55: We should store the instantiated object as a field on this class. `IDisposable` should be included to handle resource cleanup during the shutdown lifecycle to minimize false-positive exceptions

src/MobiFlightConnector/UI/MainForm.cs
```diff
@@ -216,17 +222,63 @@ public MainForm()
...
+            MessageExchange.Instance.SetSynchronizationContext(System.Threading.SynchronizationContext.Current);
+
+            InitializeMessageServer();
+        }
+
+        private void InitializeMessageServer()
+        {
+            var port = Properties.Settings.Default.FrontendWebSocketPort;
+
+#if DEBUG
```
ColinM9991, line 239: Worth mentioning that my recent example of setting IEnvironment is where scenarios such as this can be replaced instead with dependency injection and IOptions<T>. It saves on reusing many preprocessor directives throughout the codebase.

src/MobiFlightConnector/UI/MainForm.cs
```diff
@@ -216,17 +222,63 @@ public MainForm()
             InitializeTracking();
         }
 
+        /// <summary>Captures the UI SynchronizationContext and starts the message server. Overridden as a no-op in tests - see MainFormTests.TestableMainForm.</summary>
+        protected virtual void InitializeMessaging()
+        {
+            if (System.Threading.SynchronizationContext.Current == null)
```
ColinM9991, line 228: Could this be injected into MainForm.cs instead? Is there a scenario where the `Current` context would be null?

Windows Forms should assign the context in all cases. It may even be beneficial to have a service wrapper around the context which exposes a `Post` function that's used to dispatch events to the UI thread.

src/MobiFlightConnector/UI/MainForm.cs
```diff
@@ -216,17 +222,63 @@ public MainForm()
...
+            messageServer = new MessageServer(port, allowedOrigin);
+
+            try
+            {
+                messageServer.Start();
+            }
+            catch (Exception ex)
+            {
+                Log.Instance.log($"Failed to start frontend WebSocket server on port {port}: {ex.Message}", LogSeverity.Error);
+                return;
```
ColinM9991, line 256: Should this re-throw instead? If the WS server can't be started the presumably MobiFlight would be in an unrecoverable state

src/MobiFlightConnector/MobiFlight/WebView/ThreadSafeWebView2.cs
```diff
@@ -7,18 +7,6 @@ namespace MobiFlight.WebView
 {
     public class ThreadSafeWebView2 : WebView2, IWebView2Adapter
```
ColinM9991: With the removal of `PostWeb<essageAsJsonThreadSafe` I believe thread safety is no longer a requirement for WebView2 as it's handled elsewhere. Perhaps `ThreadSafeWebView2` could be removed entirely in favour if using `WebView2` directly?

src/MobiFlightConnector/MobiFlight/BrowserMessages/Transport/WebViewMessageReceiver.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace MobiFlight.BrowserMessages.Transport
+{
+    /// <summary>Inbound-only bridge for a WebView2 that stays on postMessage (the auth WebView). Wire MessageReceived to MessageExchange.PublishReceivedMessage.</summary>
+    public class WebViewMessageReceiver
+    {
+        public event Action<string> MessageReceived;
+
+        public WebViewMessageReceiver(ThreadSafeWebView2 webView)
+        {
+            webView.WebMessageReceived += (sender, args) => MessageReceived?.Invoke(args.WebMessageAsJson);
```
ColinM9991, line 13: We should unsubscribe from the event during the shutdown lifecycle to ensure messages aren't dispatched when WebView2 may be diposed, disposing or closed.

src/MobiFlightConnector/MobiFlight/BrowserMessages/MessageExchange.cs
```diff
@@ -102,30 +110,31 @@ private void PublishReceivedMessage(string jsonMessage)
...
                     {
-                        // If no synchronization context is available, invoke the subscriber directly.
                         invokeSubscriber();
                         continue;
                     }
 
-                    // if synchronization context is available
-                    // post the deserialized payload to the subscriber on the synchronization context thread
-                    synchronizationContext.Post((_) =>
+                    var synchronizationContext = _syncContextProvider != null
```
ColinM9991: This could be simplified by updating tests to use `SetSynchronizationContext` and removing the `_syncContextProvider`

src/MobiFlightConnector/MobiFlight/BrowserMessages/MessageExchange.cs
```diff
@@ -102,30 +110,31 @@ private void PublishReceivedMessage(string jsonMessage)
...
+                        ? _syncContextProvider.Invoke()
+                        : (_uiSynchronizationContext ?? System.Threading.SynchronizationContext.Current);
+
+                    if (synchronizationContext == null)
```
ColinM9991: This seems like an edge case where the invocation occurs on a background thread when it should occur on the UI thread. IMO, this should raise an exception as `synchronizationContext` definitely shouldn't be null here.

ColinM9991 (commented):
This makes my WebView2 refactoring redundant as many of the messaging concerns are addressed with this approach. I'll close my PR.

I've reviewed the .NET side of the code and have left a few comments and observations. Not all comments are requests for change, several are just observations.

src/MobiFlightConnector/UI/MainForm.cs
DocMoebiuz: @ColinM9991 you mean the one in the PR that you had shared. Correct?

src/MobiFlightConnector/UI/MainForm.cs
DocMoebiuz: True. I gotta think about that.

src/MobiFlightConnector/MobiFlight/WebView/ThreadSafeWebView2.cs
DocMoebiuz: I think it might be possible and i haven't tried yet - ExecuteScriptAsync is probably only invoked by the UI thread. I was too lazy to double check - now you don't let me another choice :D

src/MobiFlightConnector/UI/MainForm.cs
ColinM9991: Yeah exactly. It was just an example at the time but the whole `IOptions<T>` pattern with an `IEnvironment` set achieves this same thing but in a more testable manner as mocking will be possible

src/MobiFlightConnector/MobiFlight/WebView/ThreadSafeWebView2.cs
ColinM9991: Good point. It might be that we still require this abstraction for the UI dispatching. Feel free to resolve any comments that aren't possible at this point

DocMoebiuz: The IWebView interface is required for mocking and our tests. We will need a small wrapper to make it work. Only that the name is misleading.

DocMoebiuz: I simplified the class and renamed it to WebView2Adapter.

src/MobiFlightConnector/MobiFlight/BrowserMessages/Transport/WebViewMessageReceiver.cs
DocMoebiuz: I don't think that is an issue because the messages can only be generated from a WebView2 instance's content.
I would also argue that MessageExchange.Instance.PublishReceivedMessage would potentially be not defined anymore and that is handled during delegate invocation.

src/MobiFlightConnector/UI/MainForm.cs
DocMoebiuz: Showing an error page in this case would be better. Meh. Did I mention I am lazy :D

ColinM9991: I know the feeling, believe me 😂

DocMoebiuz: Ok. I added a simple dialog for now. I think we can collect feedback during beta and then decide how to improve this.

src/MobiFlightConnector/UI/Panels/FrontendPanel.cs
DocMoebiuz: I don't think we have to store the WebViewMessageReceiver. There is nothing really that needs to be `Dispose`d.
if the WebView is disposed, WebViewMessageReceiver callbacks won't be called anymore.
If the UserAuthenticationWebView is disposed, the delegate won't be valid anymore.

ColinM9991 (approved):

PR #3355, opened 2026-09-08
GetInfo() and Config - Add additional debug messages

DocMoebiuz:
### Summary
Adding detailed log statements to better understand timing and flow of board discovery during MobiFlight startup.
This shall provide more insights why users are experiencing the issue as described in #3321 

Log statements log at loglevel=DEBUG.

### Acceptance criteria

- [x] Log statements added
  - [x] GetInfo()
  - [x] Config
  - [x] Config parsing 

### DoD Checklist
n/a

## Related issues

related to #3321
related to #3340

### DIFF
--- a/src/MobiFlightConnector/MobiFlight/Firmware/Config.cs
+++ b/src/MobiFlightConnector/MobiFlight/Firmware/Config.cs
@@ -195,6 +195,8 @@ namespace MobiFlight.Firmware
                 }
                 catch (ArgumentException ex)
                 {
+                    Log.Instance.log($"Config not valid. Type not valid: {ex.Message}", LogSeverity.Debug);
+                    Log.Instance.log(value, LogSeverity.Debug);
                     if (throwException)
                         throw new FormatException("Config not valid. Type not valid", ex);
                     else
@@ -202,6 +204,8 @@ namespace MobiFlight.Firmware
                 }
                 catch (FormatException ex)
                 {
+                    Log.Instance.log($"Config not valid. Type not valid: {ex.Message}", LogSeverity.Debug);
+                    Log.Instance.log(value, LogSeverity.Debug);
                     if (throwException)
                         throw new FormatException("Config not valid. Type not valid", ex);
                     else
@@ -209,6 +213,8 @@ namespace MobiFlight.Firmware
                 }
                 catch (Exception ex)
                 {
+                    Log.Instance.log($"Config not valid.: {ex.Message}", LogSeverity.Debug);
+                    Log.Instance.log(value, LogSeverity.Debug);
                     if (throwException)
                         throw new Exception("Config not valid.", ex);
                     else
--- a/src/MobiFlightConnector/MobiFlight/MobiFlightModule.cs
+++ b/src/MobiFlightConnector/MobiFlight/MobiFlightModule.cs
@@ -5,6 +5,7 @@ using MobiFlight.Firmware;
 using MobiFlight.UI.Panels.Settings.Device;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO.Ports;
 using System.Linq;
 using System.Text.RegularExpressions;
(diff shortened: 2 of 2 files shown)


chegevarko1982:
Ran the build from this PR on the hardware from #3321. Cold boot, Debug level, four boards: three Arduino Mega 2560 clones and the Raspberry Pi Pico. It reproduces, and the new logging catches it:

```
20:39:08.297 Connected to Raspberry Pico at COM10 of type MobiFlight RaspiPico (DTR=>True).
20:39:12.069 GetInfo(): InfoCommand: 1st call - Unable to retrieve info from Raspberry Pico at COM10.
20:39:12.069 GetInfo(): Retrieved board: Raspberry Pico, Raspberry Pico, /, .
```

The board is listed as `Compatible` with no firmware and no serial.

The three Megas in the same run:

```
1st call - COM8 Elapsed=6ms   COM4 Elapsed=8ms   COM6 Elapsed=8ms
2nd call - COM8 Elapsed=13ms  COM4 Elapsed=16ms  COM6 Elapsed=22ms
```

Their `Config` reads went through too: attempt 1 Ok=True at 10 ms, 22 ms and 82 ms, parsed 1, 8 and 53 items. No parse failures anywhere in the log.

Two things from the numbers.

**The Pico's timeout is clean now.** It connected at 08.297, `ConnectionDelay` for the built-in Raspberry Pico definition is 1250 ms, so the command went out around 09.55 and the failure was logged at 12.069 — 2.52 s, essentially exactly `CommandTimeout`. Back in August on 11.2.0.5 the same wait overshot to 2971–3317 ms and the Megas needed 900–1075 ms per command instead of the 6–8 ms above. The startup load that #3323 and #3328 removed is gone, and the Pico still misses. Load was not the factor.

**The 1st call failure line has no `Elapsed=`.** Both success branches log it, the failure branch does not, and the failure is the interesting one. Worth adding.

For context on the layer below this, the USB capture in #3340: the `9;` leaves the host and the bulk OUT transfer completes successfully, and the board sends nothing back at all — for the whole timeout, and for 10 s beyond it when I kept listening.

Happy to run anything else.

chegevarko1982:
One more data point, since the official Pico builds use a current arduino-pico and mine was pinned to an old one.

I rebuilt my firmware with nothing changed except the framework: arduino-pico 3.8.0 → 6.1.0, which moves TinyUSB from 0.15.0 to 0.18.0 and the Pico SDK from 1.x to 2.x. Same firmware source, same build of this PR's Connector. Cold boot, Debug level:

```
21:02:40.333 Connected to Raspberry Pico at COM7 of type MobiFlight RaspiPico (DTR=>True).
21:02:44.120 GetInfo(): InfoCommand: 1st call - Unable to retrieve info from Raspberry Pico at COM7.
21:02:44.120 GetInfo(): Retrieved board: Raspberry Pico, Raspberry Pico, /, .
```

Identical to the 3.8.0 runs: the first command goes unanswered for the full timeout, while the three Megas answer their first call in 6–8 ms. So the framework version is not the factor, and an old TinyUSB is not the explanation.

In case it matters for reproducing this elsewhere: the firmware is a custom device built with `USE_2ND_CORE`, with `update()` running on core 1. COM7 instead of COM10 is only Windows renumbering the port after the reflash.

chegevarko1982:
Follow-up on the `USE_2ND_CORE` note in my last comment: that is not it either.

I rebuilt the same firmware on the original arduino-pico 3.8.0 pin with `-DUSE_2ND_CORE` removed and nothing else changed, so core 1 is never started and the custom device runs entirely on core 0. Cold boot, Debug level, same build of this PR's Connector:

```
21:27:52.897 Connected to Raspberry Pico at COM10 of type MobiFlight RaspiPico (DTR=>True).
21:27:56.684 GetInfo(): InfoCommand: 1st call - Unable to retrieve info from Raspberry Pico at COM10.
21:27:56.685 GetInfo(): Retrieved board: Raspberry Pico, Raspberry Pico, /, .
```

The three Megas answered their first call in 6–15 ms, as before.

So across everything I have tried, the first command is lost on every cold boot regardless of framework version or the second core:

| firmware build | cold boots | first command lost |
|---|---|---|
| arduino-pico 3.8.0, `USE_2ND_CORE` | 11 | 11 |
| arduino-pico 6.1.0, `USE_2ND_CORE` | 1 | 1 |
| arduino-pico 3.8.0, no `USE_2ND_CORE` | 1 | 1 |

The board is back on its released firmware.

PR #3358, opened 2026-09-09
Async method in AutoUpdateChecker won't fail build

### LINKED ISSUE #3258: Debug builds fail locally due to an unneeded `async` method

JunMyungH:
### Summary

Fixes the debug build issue caused by `MainForm_Shown` being declared as `async` while its awaited update-check code was excluded in debug builds.

The debug-specific update-check logic is now handled inside `AutoUpdateChecker`. In debug builds, `CheckForUpdate` returns early without checking for updates. `MainForm` can therefore use the same update-check call in both debug and non-debug builds.

### Acceptance criteria

- [x] Auto update checks are skipped in debug builds
- [x] `MainForm_Shown` contains the awaited update-check call in debug builds 
- [x] Auto update checks still run normally in non-debug builds

### DoD Checklist

None is required
     
## Related issues

fixes #3258


DocMoebiuz (changes requested):
Hi @JunMyungH - thanks for tackling this. The approach you took was suggested by me at some point. However, in our planning session, we said we wanted to move this `#if DEBUG`-wrapping to the `AutoUpdateChecker` class instead of having the "logic" inside the MainForm.

src/MobiFlightConnector/Base/AutoUpdateChecker.cs
```diff
@@ -68,6 +68,14 @@ internal static UpdatePath GetUpdatePath(bool currentIsBeta, string targetChanne
...
         public static async Task CheckForUpdate(bool silent = false)
         {
+            var skipAutoUpdateCheck = false;
+#if DEBUG
+            skipAutoUpdateCheck = true;
+#endif
+
+            if (skipAutoUpdateCheck)
+                return;
+
```
DocMoebiuz: there is an even simpler way to achieve the same result:

```suggestion
#if DEBUG
            return;
#endif
```

DocMoebiuz (changes requested):
One tiny improvement.

DocMoebiuz (approved):

PR #3365, opened 2026-09-10
Screen data not sent anymore to disabled WinCtrl controller

### LINKED ISSUE #3345: Mobiflight still trying to send winctrl screen data to device which has been set to ignore in settings
**Describe the bug**
Discord user "Admin-Erik" reported they were having issues where the screen on their winctrl PFP3N was not displaying the screen after the running script message. after reviewing the logs they provided the also had a winctrl MCDU which they had disabled as not wanting to use it in mobiflight but mobiflight was still trying to send the data to the MCDU instead of the PFP3N. Unplugging the MCDU allowed the screen on the PFP3N to work. they tested this behaviour in both 11.2.0 stable and 11.2.0.6 beta and the issue was present in both. user was testing with PMDG 777 with the sdk enabled in the options file.

**Steps To Reproduce**
1. Connect both MCDU and PFP3N devices to the pc
2. Disable the MCDU in settings
3. Run a project for a compatible aircraft
4. PFP3N display does not display the data from the sim.

**Expected behavior**
Mobiflight send display data to the PFP3N as the MCDU is disabled.

**Actual behavior**
PFP3N display remains on the running script message. no data from the sim shown

**Details**
 - MobiFlight Version: 11.2.0 & 11.2.0.6
 - FlightSim: MSFS2024
 - AddOn: PMDG 777

**Additional context**
Discord thread link - https://discord.com/channels/608690978081210392/1545377324822954054

[Winctrl display issue logs.txt](https://github.com/user-attachments/files/31832100/Winctrl.display.issue.logs.txt)

JunMyungH:
### Summary

Fixes an issue where an ignored WinCtrl CDU could still reserve its WebSocket endpoint and prevent another enabled WinCtrl CDU device from receiving display data.

Previously, WinCtrl CDU WebSocket services were registered when `WinCtrlDisplayControl` was constructed. Since ignored joystick devices are still constructed before the exclusion check, an ignored MCDU could reserve `/winwing/cdu-captain` before an enabled PFP3N was connected.

### Acceptance criteria

- [x] Ignored WinCtrl CDU devices do not register a CDU WebSocket service
- [x] An enabled PFP3N can register the Captain CDU WebSocket endpoint when an ignored MCDU is also present
- [x] CDU WebSocket registration still uses the existing `/winwing/cdu-*` endpoints

### DoD Checklist

- [x] Unit tests available
  - [x] .NET backend
     
## Related issues

fixes #3345

### Out-of-scope

Changes to the existing WinCtrl CDU WebSocket endpoint names, changes to the WinCtrl display protocol and Handling multiple enabled CDU devices that use the same Captain, Co-Pilot, or Observer endpoint are not included.

### Notes

The fix moves CDU WebSocket registration from controller initialization to the connection phase.

This matches the existing joystick exclusion flow: ignored devices may still be constructed so they can be shown in settings, but they are not connected and therefore should not register external resources such as WebSocket services.


src/MobiFlightWwFcu/WinCtrlDisplayControl.cs
```diff
@@ -70,19 +73,38 @@ private void ErrorMessageHandler(string message)
...
+            AddToCoupledControllers(controller);
+        }
+
+        internal void RegisterCduWebSocketService() 
+        {
+            if (string.IsNullOrEmpty(CduWebSocketPath) ||
+                CduWebSocketController == null)
+            {
+                return;
+            }
```
DocMoebiuz: i like to use the pattern
```
var variableNameThatExplainsTheCondition = (string.IsNullOrEmpty(CduWebSocketPath) ||
                CduWebSocketController == null)
```

When reading the code right now, I don't really understand the meaning of `string.IsNullOrEmpty(CduWebSocketPath) || CduWebSocketController == null)`

Under what circumstances will path or controller be null? It looks like it gets initialized in `AddCduController`.

`path` already exists, and `CduWebSocketPath` was added. We seem to be using the additional property as an indicator if the controller was "Added" - but why does `CduWebSocketController` not suffice?

src/MobiFlightWwFcu/WinCtrlDisplayControl.cs
```diff
@@ -214,6 +236,9 @@ public void Connect()
             {
                 controller.Connect();
             }
+
+            RegisterCduWebSocketService();
+
```
DocMoebiuz, line 222: Yes. This is a good idea. I am actually thinking that we should remove the `Init()` from the constructor and call `Init()` in connect.

I am not seeing a real reason to init the controller if we never call connect() and connect() should only be called on Controllers that are not Excluded.

DocMoebiuz (changes requested):
We should check whether we can simply move Init() to Connect() method.
All the stuff that happens in Init() seems to be required only, if you really want to use the controller. If that is the case, then moving Init() to Connect() would be the best option.

tests/MobiFlightWwFcuUnitTests/WinCtrlDisplayControlTests.cs
```diff
@@ -0,0 +1,130 @@
...
+                Path.Combine(
+                    baseDirectory,
+                    "Scripts",
+                    "Winwing",
+                    "Fonts",
+                    "Default",
+                    "PFP"
+                )
+            );
+        }
```
DocMoebiuz: This is probably the last thing to mention: when creating temporary folders for the unit tests - ideally you clean them up afterwards. Best practice is to keep TearDown() symmetrical to Setup() for these kind of things.

DocMoebiuz: it is not so much of a problem when running on a container in CI - but we also are running the tests locally on our computers - there it makes sense to clean up.

DocMoebiuz: Are we needing these directories? It looks like we are not providing any content.

DocMoebiuz (changes requested):

DocMoebiuz (approved):

DocMoebiuz (approved):

PR #3385, opened 2026-09-17
Add log level filter to the log panel

### LINKED ISSUE #3379: Add possibility to filter log items by log level in Log Panel
We have the Log Panel and it allows for filtering. However, you can only filter by message text and not by log levels.

Provide a possibility to filter by log level. This could be either by using the full text search, or by providing a dedicated filter drop down.

8rulerstar:
### Summary
The log panel could only be filtered by message text. This adds a log level filter next to it, so you can pick which severities to show. It only affects what is displayed; the log level in the settings is untouched.

The control is the same multi-select menu the config list filters use. It now lives in a standalone `FacetedFilterOptions` component that both share:
- Pick one or more levels; with nothing picked, every level is shown.
- Only levels that actually arrived are offered, each with the number of entries.
- "Clear filters" in the menu resets the levels; the existing clear button next to the text filter resets both filters.
- The level filter and the text filter apply together.

Two small things differ from the config list filter, and I'm happy to drop either if you'd rather have them identical:
- **Level names are shown in their severity colour.** I thought reusing the colours of the log entries would make the list easier to scan.
- **The "Clear filters" row stays in place, disabled, while nothing is selected.** In the config list it only appears once something is selected, and I felt that shifting the rows under the pointer was a little distracting here.

`FacetedFilterOptions` only takes `options`, `values` and `onValuesChange` (plus the optional `facets`, `title`, `keepClearVisible` and `className`) and knows nothing about tables. `DataTableFacetedFilter` is now a thin wrapper that maps a table column onto it, and the only place that keeps `"use no memo"`; its call sites in the config list toolbar are unchanged.

One visible change applies to both: the trigger has a fixed width and shows the picked values in place of the title (the first two by name, then `+N`), so selecting values no longer shifts the layout. The full selection is in a tooltip, and the title stays in the accessible name.

### Screenshots / recordings

**Before**
final_1_before

**After**
after-1-menu

after-2-filtered

### Acceptance criteria
- [x] Log entries can be filtered by one or more log levels
- [x] The level filter works together with the text filter
- [x] The clear button resets both filters
- [x] Only levels that arrived are offered, with their entry counts
- [x] The log level setting is not changed
- [x] Existing log panel behaviour (text filter, pause, copy, close) still works
- [x] The config list filters work as before (their trigger now shows the selection in a fixed width)

### DoD Checklist
- [ ] Unit tests available
  - [ ] ~~.NET backend~~
  - [ ] ~~Frontend~~
- [x] Frontend tests
  - [x] core (en,de,es)
  - [ ] ~~additional langs~~
- [ ] documentation
  - [ ] ~~Developer docs (e.g., readme.md)~~

## Related issues
fixes #3379

### Out-of-scope
- The selected levels are not remembered when the panel is closed or the app restarts.

### Notes
- Frontend only, no message or backend changes. No new strings: the level names come from the existing Settings translations, and the strings the menu shares moved from `ConfigList.Toolbar.Filter` to `General.FacetedFilter` in every language.
- Four Playwright tests in `LogPanel.spec.ts` cover this; they fail without the change and pass with it. I ran them repeatedly in parallel to check they are stable, and the config list filter tests still pass.
- In the full Playwright run, 7 tests in `ConfigListView.spec.ts` and `InputConfigWizard.spec.ts` fail on my machine (macOS), but they fail the same way on `main` without this change.
- I tested in the browser against the dev server, not in the Windows app with the real backend.

This is my first contribution to MobiFlight, so please let me know if anything should be done differently. Thank you!


tigert:
A few questions:
- should we mirror the Settings log level selector here, so it is just a handy shortcut and lets you change the log level without temporarily?
- should it also change the log level setting if you adjust it here?
- is there a real use case to pick for example ERROR and INFO but not WARN? Doing this is of course possible, but it will make the user experience more complicated.
- if we think of this as an extension to filtering of course this could work as a temporary filter, but doing it as a multiselect dropdown is harder to use, and if you just expect to set the log level like it works in preferences, this would work different and maybe unexpected?

8rulerstar:
Thanks for taking a look and for the thoughtful questions! I think this is important from a user's point of view.

I added multi-select for cases like needing to check info messages while a flood of warnings is coming in. But as you said, picking ERROR and INFO without WARN is probably not common, and it makes the UI more complicated, so a selector like the one in Settings sounds better.

It might also be more useful if it changed the actual log level setting. It would work the way people already expect from Settings, and it would let you switch to Debug right from the log panel, which a view-only filter can't do, since those entries never arrive.

What do you think? Thanks again!

DocMoebiuz:
i disagree. i dont want to change any global settings.
I want to filter what is there because we have noise.  I needed it for troubleshooting. The issue also explains that it is about filtering.

I never thought of picking multiple levels. But I really like the idea. It could be similar to the faceted search that we got for the config list view.

DocMoebiuz:
> A few questions:
> 
> * should we mirror the Settings log level selector here, so it is just a handy shortcut and lets you change the log level without temporarily?

No. Filtering doesn't effect what's logged, just what's been displayed. That is exactly what i want. 

> * should it also change the log level setting if you adjust it here?

No. See above.

> * is there a real use case to pick for example ERROR and INFO but not WARN? Doing this is of course possible, but it will make the user experience more complicated.

Not sure. But we have a similar control in the filter bar for the config list view items. So we could simply reuse it here.

> * if we think of this as an extension to filtering of course this could work as a temporary filter, but doing it as a multiselect dropdown is harder to use, and if you just expect to set the log level like it works in preferences, this would work different and maybe unexpected?

I don't understand this worry.


To clarify: the log panel shows all items based on your log level settings. As a user I am not able to see an ERROR easily when DEBUG log level is set. But I need DEBUG log level, because I am troubleshooting sth. I am interested in the ERROR message, as well as detailed DEBUG information. I cannot and don't want to change the log level temporarily nor do I want to change it permanently. I just want to filter.

8rulerstar:
Thanks a lot for taking the time to explain it so clearly — that really helped me understand what you were after. I've reworked it as a display-only filter that reuses the faceted filter from the config list view, and updated the description and screenshots above.

Two details differ from the config list filter: I used the log colours for the level names, and I kept the "Clear filters" row in place (disabled) instead of having it appear once something is selected, since I felt the rows shifting under the pointer was a little distracting. Happy to make it identical if you prefer.

I'm really enjoying working on this, so any further feedback or ideas are very welcome!

DocMoebiuz:
@8rulerstar - thanks for making the changes.

If we want to reuse the `FacetedFilterOptions` - and if you are enjoying this - then let's create a `FacetedFilterOptions` component that is independent from a Table. So instead of extending the current interface with additional options to make it work, create the component without knowing from where the data comes. For the DataTableFacetedFilter, we wrap `FacetedFilterOptions` with a small layer that maps the Table specific calls and passes these values on to the component. We can then also get rid of "use memo" inside the `FacetedFilterOptions` and only keep it for the 

Tbh, I would like to find a better way of multi-selecting items in the near future. The fact that selecting multiple items affects the layout flow is not nice. Also, i just took a look at https://ui.shadcn.com/ and i saw that they have removed the faceted search example entirely. And even if we do so, having a generic `FacetedFilterOptions`-component will help. 

I really would like to have a reusable component that we could use for the preset category filters, too (aircraft, vendor, system).

8rulerstar:
@DocMoebiuz thanks a lot for the detailed feedback — that made the direction very clear.

I pushed a commit that follows your suggestion:

- **`FacetedFilterOptions`** is now a standalone component (`src/components/FacetedFilterOptions.tsx`, next to `ComboBox`). It only takes `options`, `values`, `onValuesChange` and optionally `facets` / `title` — it has no idea where the data comes from and doesn't import anything from TanStack Table.
- **`DataTableFacetedFilter`** is a thin wrapper that maps the table column (`getFacetedUniqueValues` / `getFilterValue` / `setFilterValue`) onto it. `"use no memo"` lives only there now; `FacetedFilterOptions` doesn't need it anymore. The four call sites in the toolbar are unchanged.
- The **log panel** uses `FacetedFilterOptions` directly.
- The three strings the menu shares moved from `ConfigList.Toolbar.Filter.*` to `General.FacetedFilter.*`, same as `General.ComboBox.*`.

About the layout flow when selecting multiple items — I agree it isn't nice, so I tried a small change for that as well: the trigger now has a fixed width, and the selected values take the place of the title (the first two by name, then `+N`). The full selection is in a tooltip, and the title stays in the accessible name. With that, selecting values no longer moves the neighbouring controls. This is what the commit contains (**Option A**):

option-a

I wasn't sure what you would prefer here, so I built a few alternatives in a local playground to compare them with the same props. Two that seemed worth showing:

**Option B — title stays, values follow** (`Log Level: Warn, Error`) — clearer when several filters sit next to each other, but it truncates sooner:
option-b


**Option C — the trigger from A, plus the selection as removable chips and "Select all" inside the popover** — nicer for long lists like vendors, but it needs a new translation key and felt like too much for this PR:


option-c

I also tried the component on the preset category filters (vendor / aircraft / system) to check that it really is generic. It plugs in without changes to the component, but it turns those filters from single- into multi-select, which is a behaviour change, so I left it out of this PR.

All of this is only meant as input for the discussion — please feel free to tell me if you'd rather have B or C, a different trigger altogether, want to keep the previous look, or see it going in another direction entirely. I'm happy to adjust.

